PR #7, opened 2023-09-23
Escape OUTPUT_DIR in makefile

omer54463:
My output path has a space in it :/
This worked for me.


wmmc88 (approved):
LGTM!

As indicated in the FIXME at the top of this file, I do intend to convert these `cargo-make` script blocks into more re-usable `cargo-make` commands, which should fix all these escaping-type issues.

Anyways, thanks for the contribution @omer54463

hamzamust (approved):

PR #35, opened 2023-11-08
Arm64 Build and Packaging Improvements

wmmc88:
This PR consists of various fixes and features that resolve outstanding issues when compiling for arm64. Enhancements to the build and packaging, while made with the purpose of improving the arm64 experience, it does also add new features and improve the experience for non-arm64 builds (see README)

Outstanding Issues for this PR:
- [x] add arm64 pipeline changes
- [x] convert cargo-make scripts to commands that use the arch specific tools
- [x] handle multiple drivers crates in the same workspace
~- [ ] embed signature in sys file before signing cat, so that default flow works for boot-start drivers~ moved to separate PR #79 

fixes #14

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -25,6 +25,10 @@ jobs:
           - dev
           - release
 
+        target_triple:
+          - x86_64-pc-windows-msvc
+          # - aarch64-pc-windows-msvc FIXME: Add support for executing ARM64 tests. Maybe use target specific test runner in emulator: https://doc.rust-lang.org/cargo/reference/config.html#target
+
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v4
@@ -44,6 +48,7 @@ jobs:
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.rust_toolchain }}
+          targets: ${{ matrix.target_triple }}
 
       - name: Install Cargo Expand
         uses: taiki-e/install-action@v2
@@ -52,10 +57,8 @@ jobs:
 
       # FIXME: wdk-sys layout tests fail, but only on github hosted runner
       - name: Run Cargo Test
-        # Final driver crates must be excluded since theres no way to prevent linker args from being passed to their unit tests: https://github.com/rust-lang/cargo/issues/12663
-        run: cargo test --locked --profile ${{ matrix.cargo_profile }}  --workspace --exclude sample-* --exclude wdk-sys
+        run: cargo +${{ matrix.rust_toolchain }} test --locked --profile ${{ matrix.cargo_profile }} --target ${{ matrix.target_triple }} --workspace --exclude wdk-sys
 
       - name: Run Cargo Test (--features nightly)
         if: matrix.rust_toolchain == 'nightly'
-        # Final driver crates must be excluded since theres no way to prevent linker args from being passed to their unit tests: https://github.com/rust-lang/cargo/issues/12663
-        run: cargo test --locked --profile ${{ matrix.cargo_profile }} --features nightly --workspace --exclude sample-* --exclude wdk-sys
+        run: cargo +${{ matrix.rust_toolchain }} test --locked --profile ${{ matrix.cargo_profile }} --target ${{ matrix.target_triple }} --workspace --exclude wdk-sys --features nightly
--- a/crates/sample-kmdf-driver/Makefile.toml
+++ b/crates/sample-kmdf-driver/Makefile.toml
@@ -1,1 +1,4 @@
 extend = "../../rust-driver-makefile.toml"
+
+[env]
+WDK_BUILD_ADDITIONAL_INFVERIF_FLAGS = "/msft"
--- a/crates/wdk-build/src/cargo_make.rs
+++ b/crates/wdk-build/src/cargo_make.rs
@@ -1,0 +1,416 @@
+// Copyright (c) Microsoft Corporation
+// License: MIT OR Apache-2.0
+
+//! This module provides argument parsing functionality used by
+//! `rust-driver-makefile.toml` to validate and forward arguments common to
+//! cargo commands. It uses a combination of `clap` and `clap_cargo` to provide
+//! a CLI very close to cargo's own, but only exposes the arguments supported by
+//! `rust-driver-makefile.toml`. Help text and other `clap::Arg`
+
+use clap::{Args, Parser};
+
+/// The name of the environment variable that cargo-make uses during `cargo
+/// build` and `cargo test` commands
+const CARGO_MAKE_CARGO_BUILD_TEST_FLAGS_ENV_VAR: &str = "CARGO_MAKE_CARGO_BUILD_TEST_FLAGS";
+
+const CARGO_MAKE_PROFILE_ENV_VAR: &str = "CARGO_MAKE_PROFILE";
+const CARGO_MAKE_CARGO_PROFILE_ENV_VAR: &str = "CARGO_MAKE_CARGO_PROFILE";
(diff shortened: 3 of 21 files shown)


.github/workflows/test.yaml
```diff
@@ -24,6 +24,10 @@ jobs:
           - dev
           - release
 
+        target_triple:
+          - x86_64-pc-windows-msvc
+          # - aarch64-pc-windows-msvc FIXME: Add support for executing ARM64 tests. Maybe use target specific test runner in emulator: https://doc.rust-lang.org/cargo/reference/config.html#target
```
wmmc88, line 30: Support for ARM64 tests is possibly possible in the future via target-specific test runner: https://doc.rust-lang.org/cargo/reference/config.html#target

crates/wdk-build/src/cargo_make.rs
```diff
@@ -0,0 +1,629 @@
...
+    #[arg(
+        short,
+        long,
+        value_name = "N",
+        allow_negative_numbers = true,
+        help = "Number of parallel jobs, defaults to # of CPUs."
+    )]
+    jobs: Option<String>,
+
+    // TODO: support building multiple targets at once
```
wmmc88, line 97: Support for building multiple targets in a single cargo make invocation should be added in the future

rust-driver-makefile.toml
```diff
@@ -1,66 +1,393 @@
 # This file can be leveraged to build downstream drivers. See examples at https://github.com/microsoft/Windows-rust-drivers-samples
 
-# FIXME: replace all script blocks with cargo-make commands: "Favor commands over scripts, as commands support more features such as automatic dependencies installation, argument functions, and more..."
 # FIXME: this flow is based on the signing process of a KMDF PNP driver. There should be different flows availabe for different types of drivers as outlined in https://learn.microsoft.com/en-us/windows-hardware/drivers/install/test-signing-driver-packages
```
wmmc88, line 3: Support for other types of drivers (ex. non-PNP) should be added in the future

rust-driver-makefile.toml
```diff
@@ -1,66 +1,393 @@
...
 '''
 
 [tasks.inf2cat]
-dependencies = ["stampinf"]
+private = true
+dependencies = ["copy-sys-to-package", "copy-inf-to-package"]
+command = "inf2cat"
+args = [
+  "/driver:${WDK_BUILD_OUTPUT_DIRECTORY}/${CARGO_MAKE_CRATE_FS_NAME}_package",
+  "/os:10_NI_X64,10_VB_X64",                                                   # TODO: this should be a parameter
```
wmmc88, line 289: Future improvement: add support for specifying specific OS'es to sign for in inf2cat task

rust-driver-makefile.toml
```diff
@@ -1,66 +1,393 @@
...
+command = "makecert"
+args = [
+  "-r",
+  "-pe",
+  "-a",
+  "SHA256",
+  "-eku",
+  "1.3.6.1.5.5.7.3.3",
+  "-ss",
+  "WDRTestCertStore",                                   # TODO: this should be a parameter
```
wmmc88, line 316: Future improvement: add support for customizing the name of the Cert Store in `generate-certificate` and `signtool` task

rust-driver-makefile.toml
```diff
@@ -1,66 +1,424 @@
...
+  "copy-inf-to-package",
+  "copy-map-to-package",
+  "copy-certificate-to-package",
+  "signtool",
+  "infverif",
+]
+
+[tasks.package-driver-flow]
+# Only run flow if the current package is marked as a driver
+condition_script = [
```
wmmc88: With https://github.com/sagiegurari/cargo-make/issues/1025 being resolved in cargo make v0.37.8, these condition scripts could be refactored to be rust-based for better maintainability, but since its fully functional right now, I'll leave that as a future improvement.

rust-driver-makefile.toml
```diff
@@ -1,66 +1,424 @@
...
+command = "inf2cat"
+args = [
+  "/driver:${WDK_BUILD_OUTPUT_DIRECTORY}/${CARGO_MAKE_CRATE_FS_NAME}_package",
+  "/os:10_NI_X64,10_VB_X64",                                                   # TODO: this should be a parameter
+  "/uselocaltime",
+]
+
+[tasks.generate-certificate]
+private = true
+condition_script = [
```
wmmc88: With https://github.com/sagiegurari/cargo-make/issues/1025 being resolved in cargo make v0.37.8, these condition scripts could be refactored to be rust-based for better maintainability, but since its fully functional right now, I'll leave that as a future improvement.

crates/sample-kmdf-driver/Makefile.toml
```diff
@@ -1 +1,4 @@
 extend = "../../rust-driver-makefile.toml"
+
+[env]
+WDK_BUILD_ADDITIONAL_INFVERIF_FLAGS = "/msft"
```
hamzamust, line 4: Probably shouldn't list this flag here

crates/wdk/src/lib.rs
```diff
@@ -42,7 +44,7 @@ pub fn dbg_break() {
     unsafe {
         #[cfg(target_arch = "aarch64")]
         {
-            core::arch::asm!("int 3");
+            core::arch::asm!("brk #0xF000");
```
hamzamust, line 47: Is this instruction documented somewhere? I couldn't find the asm instruction for breakpoint on Aarch64

crates/sample-kmdf-driver/Makefile.toml
wmmc88: This is actually required since we use the `Sample` class guid in the `sample-kmdf-driver`, and that guid is reserved only for Microsoft. If we don't have this flag here, we would get the following infverif errors:
```
ERROR(1284) in D:\git-repos\github\windows-drivers-rs.git\worktrees-folder\arm64-improvements\target\debug\sample_kmdf_driver.inf, line 8: Class "Sample" is reserved for use by Microsoft.
ERROR(1285) in D:\git-repos\github\windows-drivers-rs.git\worktrees-folder\arm64-improvements\target\debug\sample_kmdf_driver.inf, line 26: Cannot specify [ClassInstall32] section for Microsoft-defined class.
```

We also [use this flag](https://github.com/microsoft/Windows-driver-samples/blob/64c1afad486ce84bb7d7128afab0aba0fd00f9ef/Build-Sample.ps1#L150) in the [`microsoft/Windows-driver-samples` repo](https://github.com/microsoft/Windows-driver-samples)

crates/wdk/src/lib.rs
wmmc88: The documentation is linked in the docstring for this function

crates/sample-kmdf-driver/Makefile.toml
wmmc88: Also, this is in `Makefile.toml` within the `sample-kmdf-driver` crate, so it only applies to `sample-kmdf-driver`.

This set of commits added functionality for `WDK_BUILD_ADDITIONAL_INFVERIF_FLAGS` which allows downstream users to add additional infverif flags akin to the `InfVerif_AdditionalOptions` property

hamzamust: Interesting. I thought it was a private flag since it is undocumented.

hamzamust (approved):

crates/sample-kmdf-driver/Makefile.toml
wmmc88: It is private, and typically shouldn't be used by those outside Microsoft. In this case, we are using it as Microsoft for our driver samples.

hamzamust (approved):

rajeshbg (approved):

PR #43, opened 2023-11-16
fix: sign inf and sys with SHA256 certificate by default

olljanat:
Use SHA256 certificate by default which have been requirement for a while already.
Also sign .sys file.

### DIFF
--- a/rust-driver-makefile.toml
+++ b/rust-driver-makefile.toml
@@ -55,11 +55,12 @@ dependencies = ["rename-dll-to-sys", "inf2cat", "infverif"]
 script = '''
 call "%VC_BUILD_DIR%"
 if not exist DriverCertificate.cer (
-  makecert -r -pe -ss PrivateCertStore -n CN=DriverCertificate DriverCertificate.cer
+  makecert -a sha256 -r -pe -ss PrivateCertStore -n CN=DriverCertificate DriverCertificate.cer
 ) else (
   echo Certificate already exists.
 )
-signtool sign /a /v /s PrivateCertStore /n DriverCertificate /fd certHash /t http://timestamp.digicert.com "%OUTPUT_DIR%\package\%CARGO_MAKE_CRATE_FS_NAME%.cat"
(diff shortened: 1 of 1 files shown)


rust-driver-makefile.toml
```diff
@@ -55,11 +55,12 @@ dependencies = ["rename-dll-to-sys", "inf2cat", "infverif"]
 script = '''
 call "%VC_BUILD_DIR%"
 if not exist DriverCertificate.cer (
-  makecert -r -pe -ss PrivateCertStore -n CN=DriverCertificate DriverCertificate.cer
+  makecert -a sha256 -r -pe -ss PrivateCertStore -n CN=DriverCertificate DriverCertificate.cer
 ) else (
   echo Certificate already exists.
 )
-signtool sign /a /v /s PrivateCertStore /n DriverCertificate /fd certHash /t http://timestamp.digicert.com "%OUTPUT_DIR%\package\%CARGO_MAKE_CRATE_FS_NAME%.cat"
+signtool sign /a /v /s PrivateCertStore /n DriverCertificate /fd sha256 /t http://timestamp.digicert.com "%OUTPUT_DIR%\package\%CARGO_MAKE_CRATE_FS_NAME%.cat"
```
wmmc88, line 62: Per https://learn.microsoft.com/en-us/windows-hardware/drivers/install/test-signing-driver-packages#-pnp-kernel-mode-boot-start-driver, if you want to sign both the sys file and the catalog file, the sys file should be signed before the catalog file.

wmmc88 (commented):
Hi @olljanat! Thanks for submitting a PR.

I'm making some large improvements to the way driver packaging works in WDR over at https://github.com/microsoft/windows-drivers-rs/pull/35, and signing with sha256 is something I have added already (although it looks like i haven't pushed my latest iteration yet). 

For embedding a signature into the sys file, we're you encountering issues with installing the hello-world driver when following https://github.com/microsoft/windows-drivers-rs/tree/main/crates/sample-kmdf-driver#sample-kmdf-rust-driver? The driver is not a boot-start driver so this shouldn't be causing any issues there. Signing the sys file seems reasonable to me, and I'll add it to https://github.com/microsoft/windows-drivers-rs/pull/35, but I want to understand if there's another issue at play here.

olljanat:
> Signing the sys file seems reasonable to me, and I'll add it to #35, but I want to understand if there's another issue at play here.

Ok, thanks. Will close this one then.

PR #44, opened 2023-11-16
docs: clarify sample-kmdf-driver install steps

olljanat:
Minor corrections to  example KMDF driver install steps

### DIFF
--- a/crates/sample-kmdf-driver/README.md
+++ b/crates/sample-kmdf-driver/README.md
@@ -15,17 +15,18 @@
    1. [`.\target\x86_64-pc-windows-msvc\debug\package`](.\target\x86_64-pc-windows-msvc\debug\package)
    2. [`.\DriverCertificate.cer`](.\DriverCertificate.cer)
    3. The version of `devgen.exe` from the WDK Developer Tools that matches the archtecture of your DUT
-     * Ex. `C:\Program Files\Windows Kits\10\Tools\10.0.22621.0\x64\devgen.exe`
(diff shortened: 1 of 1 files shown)


crates/sample-kmdf-driver/README.md
```diff
@@ -15,17 +15,18 @@
    1. [`.\target\x86_64-pc-windows-msvc\debug\package`](.\target\x86_64-pc-windows-msvc\debug\package)
    2. [`.\DriverCertificate.cer`](.\DriverCertificate.cer)
    3. The version of `devgen.exe` from the WDK Developer Tools that matches the archtecture of your DUT
-     * Ex. `C:\Program Files\Windows Kits\10\Tools\10.0.22621.0\x64\devgen.exe`
+     * Ex. `C:\Program Files (x86)\Windows Kits\10\Tools\10.0.22621.0\x64\devgen.exe`
```
wmmc88: This path is going to differ based off of whether you're using the ewdk or using installed wdk, so I'd rather make that clear instead of modifying the example path for a specific configuration. 
```suggestion
     * Ex. `C:\Program Files\Windows Kits\10\Tools\10.0.22621.0\x64\devgen.exe`. Note: This path will vary based off your WDK environment
```

wmmc88 (changes requested):
Thanks for the PR! Just one small change

crates/sample-kmdf-driver/README.md
olljanat: Ah, Ok. Included

wmmc88 (approved):
Thanks again for the contribution!

rajeshbg (approved):

PR #50, opened 2023-12-16
Refactor `call_unsafe_wdf_function_binding` so invocations look like function calls

DrChat:
This PR just changes the syntax for `call_unsafe_wdf_function_binding` so that the contents actually look like a function call.

This is important both for ergonomics (`rust-analyzer` will highlight the function like a function), as well as because I assume eventually you want to get rid of this macro altogether and this will make it easier for users to refactor their code when the time comes.

### DIFF
--- a/crates/sample-kmdf-driver/src/lib.rs
+++ b/crates/sample-kmdf-driver/src/lib.rs
@@ -82,14 +82,14 @@ pub unsafe extern "system" fn driver_entry(
     let driver_handle_output = WDF_NO_HANDLE.cast::<*mut wdk_sys::WDFDRIVER__>();
 
     let wdf_driver_create_ntstatus = unsafe {
-        call_unsafe_wdf_function_binding!(
-            WdfDriverCreate,
-            driver as wdk_sys::PDRIVER_OBJECT,
-            registry_path,
-            driver_attributes,
-            &mut driver_config,
-            driver_handle_output,
-        )
+        call_unsafe_wdf_function_binding! {
+            WdfDriverCreate(
+                driver as wdk_sys::PDRIVER_OBJECT,
+                registry_path,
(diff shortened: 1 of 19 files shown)


wmmc88:
> This is important both for ergonomics (e.g. `rust-analyzer` will highlight the function like a function

Could you elaborate on what ergonomics you're referring to here? Does allow rust-analyzer to provide any additional feature beyond just highlighting as a function?

> , as well as because I assume eventually you want to get rid of this macro altogether and this will make it easier for users to refactor their code when the time comes.

Our rough plan right now does not include the removal of the macro. Our thinking is that as we work to write idiomatic rust wrappers for the individual apis, those wrappers would use the `call_unsafe_wdf_function_binding`. Those who want to use the raw bindings in `wdk-sys` can continue to use them via the macro, but those who want to use the idiomatic api in `wdk` will need some refactor work anyways. these idiomatic apis are still something we are experimenting with so it will take some time before we get to the point of merging those in.

DrChat:
Ah sorry - I forgot to follow up on this!

> Could you elaborate on what ergonomics you're referring to here? Does allow rust-analyzer to provide any additional feature beyond just highlighting as a function?

Really the only benefits would be that this makes the function call _look_ like a function call both to developers and rust-analyzer.
It's important IMO for allowing developers to quickly understand any code that makes these calls, whether internal or external.

> Our rough plan right now does not include the removal of the macro. Our thinking is that as we work to write idiomatic rust wrappers for the individual apis, those wrappers would use the call_unsafe_wdf_function_binding.

Hopefully I can change this plan with my other PR :)
I do think that one is more valuable because it has the above benefits _plus_ it allows `rust-analyzer` to display rich information about both input parameters as well as the return value for each `Wdf` function.

PR #108, opened 2024-02-12
refactor: add lint exceptions for clippy::manual_c_str_literals and clippy::ref_as_ptr

wmmc88:
These lint errors have been causing nightly clippy to fail since 2024-02-10

### DIFF
--- a/crates/wdk-sys/src/types.rs
+++ b/crates/wdk-sys/src/types.rs
@@ -38,6 +38,13 @@
     ),
     allow(clippy::pub_underscore_fields)
 )]
+#[rustversion::attr(
+    any(
+        all(not(nightly), since(1.79)),
+        all(nightly, since(2024-02-09)),
+    ),
+    allow(clippy::ref_as_ptr)
+)]
 #[allow(clippy::semicolon_if_nothing_returned)]
 #[allow(clippy::too_many_arguments)]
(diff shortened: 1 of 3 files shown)


crates/wdk-sys/src/types.rs
```diff
@@ -38,6 +38,13 @@
     ),
     allow(clippy::pub_underscore_fields)
```
rajavenms, line 39: Could we remove the old blocks?

wmmc88: I'm not sure what you're referring to when you say old blocks. The change here is that the nightly version of the rust compiler added some new lints that are triggering on generated code. This block allows an exception for the new lint only in the compiler versions that are newer than the nightly version that introduced the lint. This is required so that older compiler versions won't fail build because of an unknown lint (since older compilers didn't have the lint). Our CI runs builds for stable, beta, and nightly builds so that any possible issues in upcoming rust versions are caught and addressed early.

Since these are triggering in generated code, I could actually just add a blanket exception for all the generated code, but I've chosen to keep it one by one since sometimes the lints can be avoided by tweaking a configuration option in bindgen to improve the quality of the generated code. I've found these nightly breakages to be fairly low effort to address and stay on top of, so I would prefer to continue doing it this way for now

rajavenms (approved):

PR #115, opened 2024-02-23
docs: fix `winget` llvm install command option

PrototypeNM1:
`-i` is an argument to `install`.


wmmc88 (approved):
Thanks for the docs fix @PrototypeNM1

hamzamust (approved):

PR #122, opened 2024-03-07
fix: only emit must_use hint when wdf function has return type

wmmc88:
Previously, the macro would emit `must_use` regardless of the return type WDF function being called, including ones that returned `()`. Now, the macro will parse the generated bindings to determine if there is a return type, before emitting `must_use`. this removes the need for ugly code blocks to avoid compiler warnings like:
```
let [()] = [macros::call_unsafe_wdf_function_binding!(
                WdfSpinLockAcquire,
                self.wdf_spin_lock
            )];
```

After this change:
```
macros::call_unsafe_wdf_function_binding!(WdfSpinLockAcquire, self.wdf_spin_lock);
```

This also stabilizes the `must_use` functionality so that its available without the `nightly` feature enabled. `cargo bloat` was used to validate that the new implementation with the nested function does not increase code size in release mode.

To make the macro more unit-testable and maintainable, its logic was broken into 4 distinct parts that are individually unit-testable:
1. Parse macro inputs into `Inputs`
2. Based off of `Inputs`, generate new AST fragments in `DerivedASTFragments`
3. Use the `DerivedASTFragments` to create the major blocks of the macro generates in `IntermediateOutputASTFragments`
4. Piece all the `IntermediateOutputASTFragments` into the tokens output by the macro

**Currently blocked on:**
- [x] fs4 crate release with fix from https://github.com/al8n/fs4-rs/pull/19#issuecomment-2029100702 (currently causing Github Dependency Review pipeline failure)
- [x] cargo-make release with fix from https://github.com/sagiegurari/cargo-make/pull/1060 (currently causing Local Development Makefile pipeline failure)
- [x] #127 merging


Makefile.toml
```diff
@@ -61,6 +66,7 @@ alias = "pre-test"
 
 [tasks.nightly-test]
 extend = "test"
+install_crate = { crate_name = "cargo-expand", version = "1.0.80" }
 env = { CARGO_MAKE_CARGO_BUILD_TEST_FLAGS = "--features nightly" }
```
wmmc88:
```suggestion
extend = "test"
env = { CARGO_MAKE_CARGO_BUILD_TEST_FLAGS = "--features nightly" }
```

Makefile.toml
```diff
@@ -88,9 +93,45 @@ env = { CARGO_MAKE_CLIPPY_ARGS = "--features nightly ${CARGO_MAKE_CLIPPY_ARGS}",
...
+[tasks.beta-test-flow]
+extend = "test-flow"
+dependencies = ["pre-beta-test", "beta-test", "post-beta-test"]
+
+[tasks.pre-beta-test]
+alias = "pre-test"
+
+[tasks.beta-test]
+extend = "test"
+install_crate = { crate_name = "cargo-expand", version = "1.0.80" }
```
wmmc88:
```suggestion
```

hamzamust (approved):

crates/wdk-macros/tests/inputs/trybuild/wdf_device_create_unused_return_type.rs
```diff
@@ -11,12 +11,15 @@ extern "C" fn evt_driver_device_add(
...
         wdk_macros::call_unsafe_wdf_function_binding!(
             WdfDeviceCreate,
             &mut device_init,
             WDF_NO_OBJECT_ATTRIBUTES,
             &mut device_handle_output,
         )
-    }
+    };
+
+    0
```
ayodejiige, line 24: Do we have an actual NTSTATUS type we can use here?

ayodejiige (approved):

PR #134, opened 2024-04-03
refactor: use a standardized workspace lint table

wmmc88:
Since the lints table feature has been released for the last 2 stable rust versions, I've migrated all of our linting configurations to use it so we can keep the configuration mostly in one place. 

Unfortunately, due to https://github.com/rust-lang/cargo/issues/13157, we cannot override workspace level lints yet, so several of the packages have locally defined lints in their Cargo manifest. This is still an improvement over before as now it applies at a package level instead of at the module level (and has caught several small issues)

### DIFF
--- a/crates/wdk-build/Cargo.toml
+++ b/crates/wdk-build/Cargo.toml
@@ -27,3 +27,38 @@ rustversion = "1.0.14"
 
 [dev-dependencies]
 windows = { version = "0.52.0", features = ["Win32_UI_Shell"] }
+
+# Cannot inherit workspace lints since overriding them is not supported yet: https://github.com/rust-lang/cargo/issues/13157
+# [lints]
+# workspace = true
+# 
+# Differences from the workspace lints have comments explaining why they are different
+
+[lints.rust]
+missing_docs = "warn"
+unsafe_op_in_unsafe_fn = "forbid"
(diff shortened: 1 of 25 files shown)


crates/wdk-build/Cargo.toml
```diff
@@ -27,3 +27,38 @@ rustversion = "1.0.14"
...
+unsafe_op_in_unsafe_fn = "forbid"
+
+[lints.clippy]
+# Lint Groups
+all = "deny"
+pedantic = "warn"
+nursery = "warn"
+cargo = "warn"
+# Individual Lints
+# multiple_unsafe_ops_per_block = "forbid"
```
ayodejiige, line 48: are these comments necessary?

ayodejiige: I see! These are modifications from the configuration at the workspace level.

ayodejiige (approved):

crates/wdk-build/Cargo.toml
wmmc88: Yup. I'm just documenting what things were changed and why. Seems a little messy right now, but once upstream cargo figures out inheritance in the next while, we can move to that for this to be cleaner

PR #143, opened 2024-04-26
fix: skip infverif task for sample drivers built with certain GE WDK versions

NateD-MSFT:
We expect that drivers built for Windows pass InfVerif.  However, the driver samples provided by Microsoft generally have settings that do not pass the default InfVerif ruleset.

In the current (E)WDK, this can be alleviated by setting /msft when calling InfVerif.exe, but in the upcoming release's WDK this syntax is no longer available.  While we wait for the InfVerif team to enable a new "/samples" flag for us, this change updates the cargo make logic to:

- Gather the WDK version
- Check the version against the build where InfVerif changed behavior
- If using the old WDK, use "/msft"; otherwise, indicate that we would like to use "/samples" but in practice skip the InfVerif task for the moment.

Note that with this change, samples should extend the makefile with
```
[tasks.infverif]
dependencies = ["wdk-samples-setup"]
condition = { env_not_set = ["WDK_INF_USING_NEW_VERSION"] }
```

When testing on one machine, I did see an issue where a ; was incorrectly appended to the /msft flag.  I haven't been able to repro this on other machines yet.


crates/sample-kmdf-driver/Makefile.toml
```diff
@@ -1,4 +1,4 @@
 extend = "../wdk-build/rust-driver-makefile.toml"
 
 [env]
-WDK_BUILD_ADDITIONAL_INFVERIF_FLAGS = "/msft"
+WDK_IS_SAMPLE = true
```
wmmc88: I don't like adding a `WDK_IS_SAMPLE` variable that all rust drivers will have in their build environment. I think we should limit sample-specific logic polluting this global space

crates/wdk-build/rust-driver-makefile.toml
```diff
@@ -177,13 +180,16 @@ args = [
   "1.33",
 ]
 
+
 [tasks.infverif]
 private = true
 dependencies = ["stampinf"]
+condition = { env_not_set = ["WDK_INF_NEW_VERSION"] }
```
wmmc88: this should be set only in samples (ie. WindowsRustDriverSamples repo + sample_kmdf_driver makefile). you could do something like add the following snippets to their local makefiles:

```
[tasks.infverif]
condition = { env_not_set = ["WDK_INF_NEW_VERSION"] }
```

This will only override the `condition` for `tasks.infverif` locally for flows run in those dirs

crates/wdk-build/rust-driver-makefile.toml
```diff
@@ -177,13 +180,16 @@ args = [
...
 command = "infverif"
 args = [
   "/v",
   "/w",
+  "@@decode(WDK_IS_SAMPLE,true,${WDK_INFVERIF_SAMPLE_FLAG})",
```
wmmc88: I think we should not be adding extra vars here specific to samples logic, when its evaluated for all drivers. sample drivers should locally override `WDK_BUILD_ADDITIONAL_INFVERIF_FLAGS`

crates/wdk-build/src/cargo_make.rs
```diff
@@ -20,6 +20,13 @@ use crate::{
 };
 
 const PATH_ENV_VAR: &str = "Path";
+const WDK_VERSION_ENV_VAR: &str = "WDK_VER";
```
wmmc88: Is `WDK_VER` an environment variable thats typically already exposed in the eWDK prompt, but we don't expose in WDR cargo make? If its not a "standard" WDK env var name, I think we should name this as `WDK_BUILD_DETECTED_WDK_VERISON` or similar.

crates/wdk-build/src/cargo_make.rs
```diff
@@ -492,6 +499,57 @@ pub fn setup_path() -> Result<(), ConfigError> {
...
+}
+
+/// Sets the `WDK_INFVERIF_SAMPLE_FLAG` environment variable to contain the
+/// appropriate flag for building samples.
+///
+/// # Errors
+///
+/// This function returns a [`ConfigError::WDKContentRootDetectionError`] if the
+/// WDK version is ill-formed (that is, it does not match the form
+/// 10.x.yyyyy.z).
```
wmmc88: I think ill-formed detection should happen in the function that exposes the version variable to the build environment (before it exposes it)

wmmc88 (changes requested):
Hey @NateD-MSFT! Thanks for the PR.

I like the change overall, but was wondering if we could isolate the sample-specific behavior outside of the core wdk-crates. This way we can avoid polluting `windows-drivers-rs` with extra environment that is not relevant to the wdk or to building rust drivers.

My suggestion:
* keep the logic to detect, validate, and expose the WDK version inside of `wdk-build` and `rust-driver-makefile.toml`
* move the sample-specific infverif version mitigation logic to a script block in `sample-kmdf-driver/Makefile.toml` (and the same makefiles in the [samples repo](https://github.com/microsoft/windows-rust-driver-samples)). It can be refactored to update `WDK_BUILD_ADDITIONAL_INFVERIF_FLAGS` by printing to stdout and using the `rust-env-update` plugin I wrote.

crates/wdk-build/rust-driver-makefile.toml
NateD-MSFT: Thanks for the pointer, changing it.

crates/wdk-build/rust-driver-makefile.toml
```diff
@@ -177,6 +194,7 @@ args = [
   "1.33",
 ]
 
+
 [tasks.infverif]
```
wmmc88:
```suggestion
]

[tasks.infverif]
```

crates/wdk-build/src/cargo_make.rs
```diff
@@ -20,6 +20,11 @@ use crate::{
 };
 
 const PATH_ENV_VAR: &str = "Path";
+const WDK_VERSION_ENV_VAR: &str = "WDK_BUILD_DETECTED_WDK_VERISON";
+/// The first WDK version with the new `InfVerif` behavior.
+const WDK_INF_NEW_VERSION: i32 = 25798;
```
wmmc88: nit: can you rename this variable to be something more descriptive. maybe `MINIMUM_SAMPLES_FLAG_WDK_VERSION` or similar

crates/sample-kmdf-driver/Makefile.toml
```diff
@@ -1,4 +1,5 @@
 extend = "../wdk-build/rust-driver-makefile.toml"
 
-[env]
-WDK_BUILD_ADDITIONAL_INFVERIF_FLAGS = "/msft"
+[tasks.infverif]
+dependencies = ["wdk-samples-setup"]
```
wmmc88:
```suggestion
dependencies = ["wdk-samples-setup"]
```

The way that cargo-make extension works, this will override the original `dependencies = ["stampinf"]` line. you probably dont want to remove the `stampinf` dependency. Something like:
```suggestion
dependencies = ["wdk-samples-setup", "stampinf"]
```

crates/wdk-build/src/utils.rs
```diff
@@ -281,6 +281,46 @@ pub fn detect_cpu_architecture_in_build_script() -> CPUArchitecture {
...
 
+/// Validates that a given string matches the WDK version format (10.xxx.yyy.zzz
+/// where xxx, yyy, and zzz are numeric and not necessarily 3 digits long)
+/// and returns the yyy portion (the build number) if so.
+///
+/// # Errors
+///
+/// This function returns a [`ConfigError::WDKContentRootDetectionError`] if the
+/// version string provided is ill-formed.
+pub fn validate_wdk_version_format<S: AsRef<str>>(version_string: &S) -> Result<&str, ConfigError> {
```
wmmc88:
```suggestion
pub fn validate_wdk_version_format<S: AsRef<str>>(version_string: S) -> Result<&str, ConfigError> {
```

There's an accidental extra layer of indirection here

crates/wdk-build/rust-driver-makefile.toml
```diff
@@ -82,6 +82,23 @@ wdk_build::cargo_make::validate_and_forward_args();
...
+//! ```cargo
+//! [dependencies]
+//! wdk-build = { path = ".", version = "0.2.0" }
+//! ```
+#![allow(unused_doc_comments)]
+
+wdk_build::cargo_make::validate_and_forward_args();
+let wdk_version = wdk_build::cargo_make::setup_wdk_version()?;
+wdk_build::cargo_make::set_sample_infverif(wdk_version)?;
+'''
```
wmmc88:
```suggestion

wdk_build::cargo_make::set_sample_infverif(wdk_version)?;
'''
```

remove `validate_and_forward_args` since the init task will already run this for all flows. I also think `setup_wdk_version` should be called in the init task (instead of here) since detected_wdk_version is useful to have in the build env and is not specific to samples. Only thing that should be here is the `set_sample_infverif` task

crates/wdk-build/src/cargo_make.rs
```diff
@@ -492,6 +497,60 @@ pub fn setup_path() -> Result<(), ConfigError> {
...
+///
+/// This function returns a [`ConfigError::WDKContentRootDetectionError`] if
+/// an invalid WDK version is provided.
+///
+/// # Panics
+///
+/// This function will panic if the function for validating a WDK version string
+/// is ever changed to no longer validate that each part of the version string
+/// is an i32.
+pub fn set_sample_infverif<S: AsRef<str>>(version: S) -> Result<(), ConfigError> {
```
wmmc88:
```suggestion
pub fn setup_infverif_for_samples<S: AsRef<str>>(version: S) -> Result<(), ConfigError> {
```

crates/wdk-build/src/cargo_make.rs
```diff
@@ -492,6 +497,60 @@ pub fn setup_path() -> Result<(), ConfigError> {
...
+    let validated_version_string = crate::utils::validate_wdk_version_format(&version)?;
```
wmmc88: I think `validate_wdk_version_format` should be split up into `is_valid_wdk_version_format` returning a `bool` and `get_wdk_build_number` that returns a `Result<&str, ConfigError>`. validation can be done only in `setup_wdk_version` and `get_wdk_build_number` is only needed in this function`.

crates/wdk-build/src/utils.rs
```diff
@@ -281,6 +281,46 @@ pub fn detect_cpu_architecture_in_build_script() -> CPUArchitecture {
...
+    let version = version_string.as_ref();
+    let mut version_parts = version.split('.');
+
+    // To make the code more readable we recreate the iterator
+    // for each validity check we do.
+
+    // First, check if we have "10" as our first value
+    if !version_parts.next().is_some_and(|first| first == "10") {
+        return Err(ConfigError::WDKContentRootDetectionError);
```
wmmc88: Add a new error type: `WDKVersionDetectionError`

crates/sample-kmdf-driver/Makefile.toml
```diff
@@ -1,4 +1,5 @@
...
+condition = { env_not_set = ["WDK_INF_USING_NEW_VERSION"] }
```
wmmc88: Have you validated this works? My understanding is that condition here actually gets evaluated before the dependencies get run. So as is currently written, `wdk-samples-setup` is never run and `WDK_INF_USING_NEW_VERSION` is never set. This would be fixed if `setup_wdk_version` is run in the init task, as suggested in another comment

wmmc88: still feels a little janky to introduce this env variable just to skip infverif for the bugged wdks. I spent some time playing around with this and came up with this:
```suggestion
condition_script = '''
#!@rust
//! ```cargo
//! [dependencies]
//! wdk-build = { path = ".", version = "0.2.0" }
//! ```
#![allow(unused_doc_comments)]

// This range is inclusive of 25798. todo can be shrunk to end at the patched version
const MISSING_SAMPLE_FLAG_WDK_BUILD_NUMBER_RANGE: u32 = 25798..;

std::panic::catch_unwind(|| {
    let wdk_version= std::env::var("WDK_BUILD_DETECTED_VERSION").expect("WDK_BUILD_DETECTED_VERSION should always be set by wdk-build-init cargo make task");
    let wdk_build_number = wdk_build::cargo_make::get_wdk_build_number(wdk_version);

    if MISSING_SAMPLE_FLAG_WDK_BUILD_NUMBER_RANGE.contains(&wdk_build_number) {
        // cargo_make will interpret returning an error from the rust-script condition_script as skipping the task
        return Err(format!("Skipping InfVerif. InfVerif in WDK Build {version} is bugged and does not contain the /samples flag.").into());
    }
    Ok(())
}).unwrap_or_else(|_|{
    // panic contents would have already been printed to console
    eprintln!("condition_script for infverif task panicked while executing. Defaulting to running inverif task.");
    Ok(())
})?
'''
```

NateD-MSFT:
I did one more quick refactor to clean things up, go ahead and review when you have time.

wmmc88 (changes requested):
Could you also spin up an accompanying pr for the samples repo? It won't merge until we do another crates.io release for WDR, but you can test the changes using git or path deps in cargo

crates/sample-kmdf-driver/Makefile.toml
NateD-MSFT: Yes, I tested that this works with newer versions of the EWDK.  WDK_INF_USING_NEW_VERSION gets correctly set and the task is skipped.

crates/wdk-build/rust-driver-makefile.toml
NateD-MSFT: Without `validate_and_forward_args` there, today, the task fails:

`[cargo-make] ERROR - Error while running plugin: Source: Unknown Line: 44 - [tasks.wdk-samples-setup]'s script output did not begin with "FORWARDING ARGS TO CARGO-MAKE:". Was `--help` passed as one of the arguments?
stdout:
WDK_BUILD_DETECTED_WDK_VERISON=10.0.27578.0;
WDK_INF_USING_NEW_VERSION=true
WDK_BUILD_ADDITIONAL_INFVERIF_FLAGS=/samples`

That aside, I'll move the call to `setup_wdk_version` to init and use the environment variable in the call to `set_sample_infverif`.

crates/sample-kmdf-driver/Makefile.toml
NateD-MSFT: I guess it's a tradeoff of jank.  From my perspective having this script in each sample TOML feels less clean than just having an environment variable they can consume.  I like your more explicit messages here though and you're correct that the fewer environment variables we have floating around the better.

Maybe we can split this into a "samples.toml" that then gets extended/imported by samples instead, and also use that for any future sample-specific logic we wind up needing? Then we don't have a big script block in every sample TOML.

wmmc88: I actually love the idea of of a `samples-makefile.toml` that all samples can extend. It also looks like cargo make supports extending multiple files and will honour the loading order you specify. We could put the whole infverif override block in the `samples-makefile.toml` and just extend `samples-makefile.toml` after `rust-driver-makefile.toml`, so that the samples `infverif` tasks are all overidden:
```toml
extend = [ { path = "extend = "target/rust-driver-makefile.toml" }, { path = "extend = "target/samples-makefile.toml" } ]
```
Also love that we can put any further samples-specific extension points there too. 

Assuming we want to put `samples-makefile.toml` in this repo (which i think is reasonable since it requires explicit inclusion on the consumer side) and have it extended in both the samples in this repo and the samples in the dedicated samples repo, we would need to package and distribute it the same way we do for `rust-driver-makefile.toml`. 

This would mean creating a new function in `wdk_build::cargo_make`, called something like `load_samples_makefile`. This new function could reuse most of the logic in `load_rust_driver_makefile` (this could get refactored out into a common function). The sample repo's load_script block would now look like:
```toml
load_script = '''
#!@rust
//! ```cargo
//! [dependencies]
//! wdk-build = "0.2.0"
//! ```
#![allow(unused_doc_comments)]

wdk_build::cargo_make::load_rust_driver_makefile()?;
wdk_build::cargo_make::load_samples_makefile()? // This line is new!
'''
```

crates/wdk-build/rust-driver-makefile.toml
wmmc88: ah! `rust-env-update` [looks for `FORWARDING ARGS TO CARGO-MAKE:`](https://github.com/microsoft/windows-drivers-rs/blob/88d5ee25638ea3656a619f5df1dac43840c887c2/crates/wdk-build/rust-driver-makefile.toml#L58) as a delimiter to start parsing rustscript output to forward env vars to the executing cargo make process. 

This [part of `validate_and_forward_args` explains it](https://github.com/microsoft/windows-drivers-rs/blob/88d5ee25638ea3656a619f5df1dac43840c887c2/crates/wdk-build/src/cargo_make.rs#L404). adding the print in `set_sample_infverif` right before you start printing the vars you want to update will fix the issue

crates/wdk-build/rust-driver-sample-makefile.toml
```diff
@@ -0,0 +1,53 @@
+# This file is used to extend the standarrd rust-driver-makefile to build official sample drivers. See examples at https://github.com/microsoft/Windows-rust-drivers-samples
```
wmmc88:
```suggestion
# This file is used to extend the standard rust-driver-makefile to build official sample drivers. See examples at https://github.com/microsoft/Windows-rust-drivers-samples
```

crates/wdk-build/src/lib.rs
```diff
@@ -119,6 +119,11 @@ pub enum ConfigError {
     )]
     WDKContentRootDetectionError,
 
+    /// Error returned when the WDK version string does not match the expected
+    /// format
+    #[error("The WDK version string provided was not in a valid format.")]
```
wmmc88: Especially since the version is parsed from things in the environment, it would be pretty useful to include the version in the error message. Similar to how `ConfigError::DirectoryNotFound` will print the directory it was trying to find

crates/wdk-build/src/cargo_make.rs
```diff
@@ -566,13 +639,61 @@ pub fn copy_to_driver_package_folder<P: AsRef<Path>>(path_to_copy: P) -> Result<
...
 ///
 /// # Panics
 ///
 /// This function will panic if the `CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY`
 /// environment variable is not set
 pub fn load_rust_driver_makefile() -> Result<(), ConfigError> {
+    load_rust_driver_makefile_internal(RUST_DRIVER_MAKEFILE_NAME)
+}
+
+/// Symlinks `rust-driversample-makefiletoml` to the `target` folder where it
```
wmmc88:
```suggestion
/// Symlinks `rust-driver-sample-makefile.toml` to the `target` folder where it
```

crates/wdk-build/src/cargo_make.rs
```diff
@@ -566,13 +639,61 @@ pub fn copy_to_driver_package_folder<P: AsRef<Path>>(path_to_copy: P) -> Result<
...
+/// - [`ConfigError::MultipleWDKBuildCratesDetected`] if there are multiple
+///   versions of the WDK build crate detected
+/// - [`ConfigError::IoError`] if there is an error creating or updating the
+///   symlink to the makefile.
+///
+/// # Panics
+///
+/// This function will panic if the `CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY`
+/// environment variable is not set
+fn load_rust_driver_makefile_internal<S: AsRef<str> + AsRef<Utf8Path> + AsRef<Path>>(
```
wmmc88: nit

```suggestion
/// Symlinks a [`wdk_build`] `cargo-make` makefile to the `target` folder where
/// it can be extended from a downstream `Makefile.toml`. This is necessary so that paths
/// in the [`wdk_build`] makefile can be relative to
/// `CARGO_MAKE_CURRENT_TASK_INITIAL_MAKEFILE_DIRECTORY`
///
/// # Errors
///
/// This function returns:
/// - [`ConfigError::CargoMetadataError`] if there is an error executing or
///   parsing `cargo_metadata`
/// - [`ConfigError::MultipleWDKBuildCratesDetected`] if there are multiple
///   versions of the WDK build crate detected
/// - [`ConfigError::IoError`] if there is an error creating or updating the
///   symlink to the makefile.
///
/// # Panics
///
/// This function will panic if the `CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY`
/// environment variable is not set
fn load_wdk_build_makefile<S: AsRef<str> + AsRef<Utf8Path> + AsRef<Path>>(
```

crates/wdk-build/rust-driver-sample-makefile.toml
```diff
@@ -0,0 +1,53 @@
...
+		// cargo_make will interpret returning an error from the rust-script condition_script as skipping the task
+		return Err(format!("Skipping InfVerif. InfVerif in WDK Build {version} is bugged and does not contain the /samples flag.").into());
+	}
+	Ok(())
+}).unwrap_or_else(|_|{
+	// panic contents would have already been printed to console
+	eprintln!("condition_script for infverif task panicked while executing. Defaulting to running inverif task.");
+	Ok(())
+})?
+'''
```
wmmc88: nit: end files in newline

crates/wdk-build/rust-driver-sample-makefile.toml
```diff
@@ -0,0 +1,53 @@
...
+install_crate = { crate_name = "rust-script", min_version = "0.30.0" }
+script_runner = "@rust"
+condition_script = '''
+//! ```cargo
+//! [dependencies]
+//! wdk-build = { path = ".", version = "0.2.0" }
+//! ```
+#![allow(unused_doc_comments)]
+
+// This range is inclusive of 25798. Will update with range end after /sample flag is added to InfVerif CLI
```
wmmc88:
```suggestion
// This range is inclusive of 25798. FIXME: update with range end after /sample flag is added to InfVerif CLI
```

crates/wdk-build/src/cargo_make.rs
wmmc88: I think you can remove the traitbound on `AsRef<Path>>`

wmmc88 (commented):
Looking good! pretty close to mergeable once the remaining issues are resolved

crates/wdk-build/rust-driver-sample-makefile.toml
NateD-MSFT: Yep, forgot to run toml format before this

crates/wdk-build/rust-driver-sample-makefile.toml
```diff
@@ -0,0 +1,53 @@
+# This file is used to extend the standarrd rust-driver-makefile to build official sample drivers. See examples at https://github.com/microsoft/Windows-rust-drivers-samples
+# Using this file requires extending both the standard makefile and this makefile in order, as follows:
+# extend = [ { path =  "target/rust-driver-makefile.toml" }, { path = "target/rust-driver-sample-makefile.toml" } ]
+
+[tasks.wdk-samples-setup]
+private = true
+dependencies = ["wdk-build-init"]
+install_crate = { crate_name = "rust-script", min_version = "0.30.0" }
```
wmmc88:
```suggestion
private = true
install_crate = { crate_name = "rust-script", min_version = "0.30.0" }
```

This dependency will cause init to run twice (once from dependency and once as the special init task

crates/wdk-build/rust-driver-sample-makefile.toml
```diff
@@ -0,0 +1,73 @@
...
+dependencies = ["wdk-samples-setup", "stampinf"]
+install_crate = { crate_name = "rust-script", min_version = "0.30.0" }
+script_runner = "@rust"
+plugin = "rust-condition-script-cargo_make_current_task_initial_makefile_directory-substitution"
+condition_script = '''
+//! ```cargo
+//! [dependencies]
+//! wdk-build = { path = "${CARGO_MAKE_CURRENT_TASK_INITIAL_MAKEFILE_DIRECTORY}", version = "0.2.0" }
+//! ```
+#![allow(unused_doc_comments)]
```
wmmc88:
```suggestion
[tasks.infverif]
dependencies = ["wdk-samples-setup", "stampinf"]
plugin = "rust-condition-script-cargo_make_current_task_initial_makefile_directory-substitution"
condition_script = '''
#!@rust

//! ```cargo
//! [dependencies]
//! wdk-build = { path = "${CARGO_MAKE_CURRENT_TASK_INITIAL_MAKEFILE_DIRECTORY}", version = "0.2.0" }
//! ```
#![allow(unused_doc_comments)]
```

Try this. I think it might solve your error

NateD-MSFT: Still no luck, I'm afraid.  From cargo make -v:

```
[cargo-make] VERBOSE - Changing working directory to: C:\Users\natede\source\repos\windows-drivers-rs\crates\sample-kmdf-driver
[cargo-make] VERBOSE - Working directory changed to: C:\Users\natede\source\repos\windows-drivers-rs\crates\sample-kmdf-driver
[cargo-make] VERBOSE - Running Task: infverif via plugin: rust-condition-script-cargo_make_current_task_initial_makefile_directory-substitution
[cargo-make] VERBOSE - Running Task: infverif via plugin: rust-condition-script-cargo_make_current_task_initial_makefile_directory-substitution script:
# Parse task json into variables
taskjson = json_parse ${task.as_json}

# Convert backslashes to forward slashes
rust-driver-toolchain-path = replace ${taskjson.env.CARGO_MAKE_CURRENT_TASK_INITIAL_MAKEFILE_DIRECTORY} "\\" "/"

# Replace the ${CARGO_MAKE_CURRENT_TASK_INITIAL_MAKEFILE_DIRECTORY} variable in the condition_script with the rust-driver-toolchain-path
taskjson.condition_script = replace ${taskjson.condition_script} "\${CARGO_MAKE_CURRENT_TASK_INITIAL_MAKEFILE_DIRECTORY}" "${rust-driver-toolchain-path}"

# Reencode variables into json
taskjson = json_encode taskjson

# Run the invoking task, with the modified condition_script
cm_plugin_run_custom_task ${taskjson}

[cargo-make] ERROR - Error while running plugin: Source: Unknown Line: 22 - invalid type: string "true", expected a boolean at line 1 column 1858
[cargo-make] WARN - Build Failed.
```

crates/wdk-build/src/cargo_make.rs
NateD-MSFT: You'd think so, but the compiler is unhappy when I try.

crates/sample-kmdf-driver/Makefile.toml
NateD-MSFT: Since this change in the samples repo depends on a new version of this crate, I haven't opened a PR there yet.  Is this OK?

NateD-MSFT: Created a PR in the samples repo for once this crate is updated and released: https://github.com/microsoft/Windows-rust-driver-samples/pull/18

NateD-MSFT:
> Could you also spin up an accompanying pr for the samples repo? It won't merge until we do another crates.io release for WDR, but you can test the changes using git or path deps in cargo

Created here: https://github.com/microsoft/Windows-rust-driver-samples/pull/18

crates/wdk-build/rust-driver-sample-makefile.toml
```diff
@@ -0,0 +1,78 @@
+# This file is used to extend the standard rust-driver-makefile to build official sample drivers. See examples at https://github.com/microsoft/Windows-rust-drivers-samples
+# Using this file requires extending both the standard makefile and this makefile in order, as follows:
+# extend = [ { path =  "target/rust-driver-makefile.toml" }, { path = "target/rust-driver-sample-makefile.toml" } ]
+
+# This plugin replaces the embedded `CARGO_MAKE_CURRENT_TASK_INITIAL_MAKEFILE_DIRECTORY` in the rust-script doc comment with its evaluated value. See https://github.com/sagiegurari/cargo-make/issues/1081 for more context.
+[plugins.impl.rust-condition-script-cargo_make_current_task_initial_makefile_directory-substitution]
```
wmmc88, line 6: With resolution of https://github.com/sagiegurari/cargo-make/issues/1081, this plugin may be no longer needed. This can be addressed in a followup

crates/wdk-build/rust-driver-sample-makefile.toml
```diff
@@ -0,0 +1,78 @@
...
+
+//! ```cargo
+//! [dependencies]
+//! wdk-build = { path = "${CARGO_MAKE_CURRENT_TASK_INITIAL_MAKEFILE_DIRECTORY}", version = "0.2.0" }
+//! ```
+#![allow(unused_doc_comments)]
+
+use core::ops::RangeFrom;
+
+// This range is inclusive of 25798. FIXME: update with range end after /sample flag is added to InfVerif CLI
```
wmmc88, line 61: update with range end after /sample flag is added to InfVerif CLI

wmmc88 (approved):

crates/wdk-build/rust-driver-sample-makefile.toml
```diff
@@ -0,0 +1,78 @@
...
+const MISSING_SAMPLE_FLAG_WDK_BUILD_NUMBER_RANGE: RangeFrom<u32> = 25798..;
+
+std::panic::catch_unwind(|| {
+	let wdk_version = std::env::var(wdk_build::cargo_make::WDK_VERSION_ENV_VAR).expect("WDK_BUILD_DETECTED_VERSION should always be set by wdk-build-init cargo make task");
```
wmmc88, line 65: the logic of this condition script is complex enough that i think it'd be better to move it all into a function in `cargo_make.rs`. this also avoids needing to set `utils` to `pub` unnecessarily

hamzamust (approved):

Issue #82, opened 2024-01-23
Support WDK Configuration in crates.io-distributed crate


wmmc88:
Currently, the crate that's distributed via crates.io has several things hardcoded in the configuration. Some examples of this include [the WdfFunctions_01033 functional table reference](https://github.com/microsoft/windows-drivers-rs/blob/9202e28d93232824b5469ac8f64f3980220024f5/crates/wdk-sys/src/lib.rs#L63C9-L63C63), and the [WdfMinimumVersionDesired symbol](https://github.com/microsoft/windows-drivers-rs/blob/9202e28d93232824b5469ac8f64f3980220024f5/crates/wdk-sys/src/wdf.rs#L23). The `wdk-build::Config` used in `wdk-sys` also defaults to [KMDF 1.33](https://github.com/microsoft/windows-drivers-rs/blob/9202e28d93232824b5469ac8f64f3980220024f5/crates/wdk-build/src/lib.rs#L618).

While the repo supports different WDF versions, different driver models (WDM, KMDF, UMDF), etc, it is currently cumbersome to use anything that is not KMDF v1.33. Our goal should be to expose allow the user to be able to use the crate off crates.io and be able to configure all of the common configuration options of the WDK. 

A possible way to do this would be to have mutually-exclusive [cargo-features](https://doc.rust-lang.org/cargo/reference/features.html#features) for `wdk-sys` and `wdk` that configure what bindings are created by the crate.
There are some [caveats](https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features) for using mutually exclusive features, but it might be the best way forward that's currently available. There is extensive discussion about better ways to accomplish this in the future [here](https://internals.rust-lang.org/t/pre-rfc-mutually-excusive-global-features/19618/11) and [here](https://github.com/rust-lang/cargo/issues/2980), but we must make do with what is available right now. 

Another possible approach is to use environment variables to control the various features. Env variables can only have one value at any given time, so they are already mutually exclusive. They could be set via normal environment variable in shell, via [`env` section in a `config.toml`](https://doc.rust-lang.org/nightly/cargo/reference/config.html#env), or in [any of the numerous ways cargo-make provides](https://github.com/sagiegurari/cargo-make?tab=readme-ov-file#usage-env). This does harm discoverability of options, and there are some issues with some ways to configure env variables in Cargo right now (ex. https://github.com/rust-lang/cargo/issues/10358. This one in particular seems like it will not be resolved for a while).

[...]

PR #159, opened 2024-05-17
Build perf: Make calls to bindgen run in parallel

NateD-MSFT:
When compiling this repo (or projects that depend on it), a large portion of the build time is spent waiting for bindgen to finish constructing bindings to the WDK.  This can severely slow down iterative development and builds.

This PR makes a simple change to how we call bindgen by calling each function wrapping a call to it in parallel.  On my machine this provides a speedup of running "cargo make" in windows-drivers-rs of about 20% (~500 seconds run time to ~400 seconds).

I wouldn't be surprised if we can refactor the calls even more to see greater speedups, but this was just a quick late afternoon project to speed up work on other features.

### DIFF
--- a/crates/wdk-build/src/bindgen.rs
+++ b/crates/wdk-build/src/bindgen.rs
@@ -15,7 +15,7 @@ pub trait BuilderExt {
     ///
     /// Implementation may return `wdk_build::ConfigError` if it fails to create
     /// a builder
-    fn wdk_default(c_header_files: Vec<&str>, config: Config) -> Result<Builder, ConfigError>;
+    fn wdk_default(c_header_files: Vec<&str>, config: &Config) -> Result<Builder, ConfigError>;
 }
 
 impl BuilderExt for Builder {
@@ -26,7 +26,7 @@ impl BuilderExt for Builder {
     ///
     /// Will return `wdk_build::ConfigError` if any of the resolved include or
     /// library paths do not exist
-    fn wdk_default(c_header_files: Vec<&str>, config: Config) -> Result<Self, ConfigError> {
+    fn wdk_default(c_header_files: Vec<&str>, config: &Config) -> Result<Self, ConfigError> {
         let mut builder = Self::default();
 
         for c_header in c_header_files {
--- a/crates/wdk-sys/build.rs
+++ b/crates/wdk-sys/build.rs
@@ -6,6 +6,7 @@
 use std::{
     env,
     path::{Path, PathBuf},
+    thread::{self, JoinHandle},
 };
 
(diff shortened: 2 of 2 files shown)


crates/wdk-build/src/bindgen.rs
```diff
@@ -26,7 +26,7 @@ impl BuilderExt for Builder {
     ///
     /// Will return `wdk_build::ConfigError` if any of the resolved include or
     /// library paths do not exist
-    fn wdk_default(c_header_files: Vec<&str>, config: Config) -> Result<Self, ConfigError> {
+    fn wdk_default(c_header_files: Vec<&str>, config: &Config) -> Result<Self, ConfigError> {
```
wmmc88, line 29: Haha. I made a similar change to this in a feature branch I haven't pr'ed yet

crates/wdk-sys/build.rs
```diff
@@ -135,11 +136,23 @@ fn main() -> anyhow::Result<()> {
...
 
+    let mut handles = Vec::<JoinHandle<Result<(), ConfigError>>>::new();
     for out_path in out_paths {
-        generate_constants(&out_path, config.clone())?;
-        generate_types(&out_path, config.clone())?;
-        generate_ntddk(&out_path, config.clone())?;
-        generate_wdf(&out_path, config.clone())?;
+        let temp_config = config.clone();
+        let handle: JoinHandle<Result<(), ConfigError>> = thread::spawn(move || {
+            generate_constants(&out_path, &temp_config)?;
```
wmmc88: Each of these 4 generates could actually be paralllelized

wmmc88 (approved):
This pr speeds up the build by parallelizing the generation of the bindings in outdir and in the generated_bindings folder. The latter was only added since it was easier to get diffs between prs when there are bindgen changes. I have a pr that removes this duplicate generation (since that pr also introduces feature in #82 so it doesnt make sense to only look at the km bindings diff).

I have written a cargo make task that will generate these binding diffs, and will clean it up and pr in the near future

I am still good to merge this until that lands

crates/wdk-build/src/bindgen.rs
NateD-MSFT: Great minds think alike!

crates/wdk-sys/build.rs
NateD-MSFT: D'you know, I think that was my original intent and I got distracted by various compile errors (still learning the nuances of Rust threads.)  I'll make that change and check that everything still works as expected.

wmmc88 (approved):

hamzamust (approved):

PR #175, opened 2024-06-15
ci: pin LLVM version in pipelines

wmmc88:
As documented in #115 , LLVM 18 is bugged for arm64 targets. 

This PR:
* adds an additional CI matrix entry to test on specific LLVM versions instead of using whatever version is in the [github windows runner image](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md)
    * currently only testing on latest LLVM 17
* switches all winget commands to use the new winget powershell module so its easier to parse output


hamzamust (approved):

ayodejiige:
See comments in [microsoft/Windows-rust-driver-sample#19 ](https://github.com/microsoft/Windows-rust-driver-samples/pull/19)

ayodejiige (approved):
Approved with some minor comments

PR #178, opened 2024-06-22
Remove ? operator from build.rs step in readme

pc386:
Removed the extraneous `?` operator from the `configure_binary_build()` call in `build.rs` in the README. It does not return a Result therefore ? cannot be used. Compiler error: the `?` operator can only be applied to values that implement `Try`
the trait `Try` is not implemented for `()`

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ The crates in this repository are available from [`crates.io`](https://crates.io
 
    ```rust
    fn main() -> Result<(), wdk_build::ConfigError> {
-      wdk_build::Config::from_env_auto()?.configure_binary_build()?;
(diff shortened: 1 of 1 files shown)


wmmc88:
The [latest code on main](https://github.com/microsoft/windows-drivers-rs/blob/27309815433e0a550902e835220d7d6a24822477/crates/wdk-build/src/lib.rs#L564) has `configure_binary_build` returning a `Result` type, so the `?` in the `README` is valid. The latest code on main is not released to crates.io yet. 

Please refer the readme in the tag of the release you're using (this is the automatic behavior if you look at the crate in crates.io

PR #183, opened 2024-07-09
feat: add more precise NTSTATUS const fns

NateD-MSFT:
The current NT_SUCCESS macro checks for NTSTATUS >= 0.  While this is semantically equivalent to the official behavior I think it is appropriate to match MSDN documentation exactly.

This PR also adds the other NT_ macros for information, warning, and error values.  

Reference: https://learn.microsoft.com/en-us/windows-hardware/drivers/kernel/using-ntstatus-values


crates/wdk-sys/src/lib.rs
```diff
@@ -64,8 +64,33 @@ lazy_static! {
 #[allow(missing_docs)]
 #[must_use]
 #[allow(non_snake_case)]
+#[allow(clippy::cast_sign_loss)]
 pub const fn NT_SUCCESS(nt_status: NTSTATUS) -> bool {
-    nt_status >= 0
+    nt_status as u32 <= 0x7FFF_FFFF
+}
+
+#[allow(missing_docs)]
```
wmmc88: Could you update all of these to just have a docstring to that mirrors the current mslearn descriptions

crates/wdk-sys/src/lib.rs
```diff
@@ -64,8 +64,33 @@ lazy_static! {
 #[allow(missing_docs)]
 #[must_use]
 #[allow(non_snake_case)]
+#[allow(clippy::cast_sign_loss)]
 pub const fn NT_SUCCESS(nt_status: NTSTATUS) -> bool {
-    nt_status >= 0
```
wmmc88, line 81: Greater than or equal to 0 actually matches the actual definition of `NT_SUCCESS` in `ntdef.h` in the WDK. 

I think that if we're going to make a change here that differs from what it is exactly in the WDK, we might as well make it clearer how these "magic ranges" are determined. https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/87fba13e-bf06-450e-83b1-9241dc81e781 outlines how the 2 MSB correspond with a severity code, and constants for `STATUS_SEVERITY_SUCCESS`, `STATUS_SEVERITY_WARNING`, `STATUS_SEVERITY_INFORMATIONAL`, `STATUS_SEVERITY_ERROR`. maybe write some inline const fn that masks only the severity portion of ntstatus and comparing that value wiht the `ERROR_SEVERITY_INFORMATIONAL`, `ERROR_SEVERITY_ERROR`, `ERROR_SEVERITY_WARNING`, and `ERROR_SEVERITY_SUCCESS` constants...

otherwise, I think we should strictly stick to the code in the WDK

crates/wdk-sys/src/lib.rs
```diff
@@ -64,8 +64,33 @@ lazy_static! {
...
+    nt_status as u32 <= 0x7FFF_FFFF
```
ChrisDenton: That seems a lot less clear?

ChrisDenton:
For reference, here's how they're defined in C:

```c
#define NT_SUCCESS(Status) (((NTSTATUS)(Status)) >= 0)

#ifdef _PREFAST_
#define NT_INFORMATION(Status) (((NTSTATUS)(Status)) >= (long)0x40000000)
#else
#define NT_INFORMATION(Status) ((((ULONG)(Status)) >> 30) == 1)
#endif

#ifdef _PREFAST_
#define NT_WARNING(Status) (((NTSTATUS)(Status) < (long)0xc0000000))
#else
#define NT_WARNING(Status) ((((ULONG)(Status)) >> 30) == 2)
#endif

#ifdef _PREFAST_
#define NT_ERROR(Status) (((NTSTATUS)(Status)) >= (unsigned long)0xc0000000)
#else
#define NT_ERROR(Status) ((((ULONG)(Status)) >> 30) == 3)
#endif
```

wmmc88 (changes requested):

crates/wdk-sys/src/lib.rs
NateD-MSFT: Fair feedback.  I did not look at the WDK code itself and only based it off the MSDN page, naively assuming the WDK code would match what was on MSDN.  I'll make adjustments.

hamzamust (dismissed):

crates/wdk-sys/src/lib.rs
NateD-MSFT: Reverted to >= 0 in the latest commit.

tests/wdk-sys-tests/tests/nt_status_functions.rs
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation
```
wmmc88, line 1: Having this need to be a separate crate in order to run the test in an environment that has a WDK config is unfortunate. I have something in the works that will allow these unit tests to live in the same module of the functions they're testing

wmmc88: tracked in #211

wmmc88 (approved):

hamzamust (approved):

NateD-MSFT:
@wmmc88 it looks like one of the docs jobs hanged.  Can you restart it or bypass it?

wmmc88:
> @wmmc88 it looks like one of the docs jobs hanged.  Can you restart it or bypass it?

Ah. Hanged on winget installation. Winget installation is primarily used to get the WDK. Do you think someone could port the pipelines to use the new nuget pkgs?

Issue #171, opened 2024-05-30
Replace `rust-condition-script-cargo_make_current_task_initial_makefile_directory-substitution` plugin with new cargo make feature


wmmc88:
With resolution of https://github.com/sagiegurari/cargo-make/issues/1081, this plugin may be no longer needed. This can be addressed in a followup

_Originally posted by @wmmc88 in https://github.com/microsoft/windows-drivers-rs/pull/143#discussion_r1621435724_

svasista-ms:
@wmmc88, I'm working on this issue. I'll raise a PR after #186 is merged.

wmmc88:
@svasista-ms On closer inspection, I believe this task is blocked by https://github.com/sagiegurari/cargo-make/pull/1132 merging. While https://github.com/sagiegurari/cargo-make/issues/1081 was resolved, it still fails to expand the args based of local testing. https://github.com/sagiegurari/cargo-make/pull/1132 should fix that.

wmmc88:
@svasista-ms you should be unblocked now

PR #186, opened 2024-07-18
feat: configure WDK configuration via parsing Cargo manifest metadata

### LINKED ISSUE #82: Support WDK Configuration in crates.io-distributed crate

wmmc88:
This PR enables configuring the WDK Metadata via `[metadata.wdk]` tags in Cargo manifests.

As a consequence of that new feature, several other changes were required:
* Added minimal Sample WDM and UMDF drivers
* Make all driver packaging tooling (ex. signing, cat generation, etc) support WDM and UMDF
* Serde Serializer added to export WDKMetadata configuration to a structured format
* Emit that structured format so that downstream code can `cfg` based off of certain WDK configuration parameters
* A `wdf.c` file is compiled and linked in to provide the `__declspec(selectany)` that are not generated by bindgen
* Have several templates that generate rust code in `wdk-sys`'s build.rs that is otherwise not possible to generate conditionally (ex. call_unsafe_wdf_function_binding definition that points to `OUT_DIR`, `WDFFUNCTIONS` symbol name that depends on configuration, `WDF_FUNCTION_TABLE` definition whose size depends on configuration, etc.)
* added UM support 
* remove local `generated_bindings` copy of bindings
* exporting via `links` keys in cargo manifests is deprecated
* switch to cargo-machete for unused dep detection (required since most code is not cfg-ed out)
* remove wdk-macros running of `cargo check` to obtain out dir (and replace with wdk_sys passing the outdir to the procmacro)

resolves #82


NateD-MSFT:
Detailed review is in progress, but quick design question: can you help me understand why we roll a custom (de)serializer rather than building off the TOML crate?

wmmc88:
> Detailed review is in progress, but quick design question: can you help me understand why we roll a custom (de)serializer rather than building off the TOML crate?

@NateD-MSFT the custom serde serializer takes the rust `metadata::Wdk` struct and turns it into a map that's eventually emitted as cargo prints to set `cfg's`. There is no similar "to_map" serializer.

There is no custom deserializer here. `cargo_metadata` [deserializes all the cargo metadata](https://github.com/wmmc88/windows-drivers-rs/blob/b95365c6a5e836f4e5b18da40f46089d1e4a073e/crates/wdk-build/src/lib.rs#L228) into `serde_json::Value`, and then we [deserialize that `serde_json::Value` into the `metadata::Wdk` struct using `serde_json`](https://github.com/wmmc88/windows-drivers-rs/blob/b95365c6a5e836f4e5b18da40f46089d1e4a073e/crates/wdk-build/src/metadata/mod.rs#L188).

crates/wdk-build/src/cargo_make.rs
```diff
@@ -754,16 +840,106 @@ fn load_wdk_build_makefile<S: AsRef<str> + AsRef<Utf8Path> + AsRef<Path>>(
...
+/// [`rust-driver-makefile.toml`](../rust-driver-makefile.toml)
+///
+/// # Errors
+///
+/// This function returns an error whenever it determins that the
+/// `package-driver-flow` `cargo-make` task should be skipped (i.e. when no WDK
+/// configurations are detected)
+pub fn package_driver_flow_condition_script() -> anyhow::Result<()> {
+    condition_script(|| {
+        match metadata::Wdk::try_from(&get_cargo_metadata()?) {
```
wmmc88: This doesn't work for workspaces that have WDK configs, but when the current crate being processed is not marked as a driver

README.md
```diff
@@ -40,76 +40,77 @@ The crates in this repository are available from [`crates.io`](https://crates.io
...
-   use wdk_alloc::WDKAllocator;
+   use wdk_alloc::WdkAllocator;
 
    #[cfg(not(test))]
    #[global_allocator]
-   static GLOBAL_ALLOCATOR: WDKAllocator = WDKAllocator;
+   static GLOBAL_ALLOCATOR: WdkAllocator = WdkAllocator;
    ```
 
    This is only required if you want to be able to use the [`alloc` modules](https://doc.rust-lang.org/alloc/) in the rust standard library. You are also free to use your own implementations of global allocators.
```
NateD-MSFT: Not directly related to this PR, but do we actually want to have language stating people can roll their own global allocators? For kernel mode I expect we want to use (a variant of) ExAllocatePool2 or bust.

crates/wdk-build/rust-driver-makefile.toml
```diff
@@ -41,31 +41,79 @@ if is_file ${filepath}
...
+                value = array_get ${parts} 1
+                set_env ${key} ${value}
+            end
+        end
     end
+    assert ${looking_for_start_delimiter} "A matching \"END OF FORWARDING ARGS TO CARGO-MAKE\" for a \"FORWARDING ARGS TO CARGO-MAKE:\" was not found in script output."
 end
 '''
 
+# This plugin replaces the embedded `CARGO_MAKE_CURRENT_TASK_INITIAL_MAKEFILE_DIRECTORY` in the rust-script doc comment with its evaluated value. See https://github.com/sagiegurari/cargo-make/issues/1081 for more context.
```
NateD-MSFT, line 98: Just checking - is this task still necessary given https://github.com/sagiegurari/cargo-make/issues/1081 is now resolved and released?

crates/wdk-build/rust-driver-makefile.toml
```diff
@@ -283,11 +395,12 @@ wdk_build::cargo_make::copy_to_driver_package_folder(
 
 [tasks.inf2cat]
 private = true
-dependencies = ["copy-sys-to-package", "copy-inf-to-package"]
+dependencies = ["copy-driver-binary-to-package", "copy-inf-to-package"]
+env = { "WDK_BUILD_INF2CAT_OS" = { source = "${CARGO_MAKE_CRATE_TARGET_TRIPLE}", default_value = "UNKNOWN", mapping = { "x86_64-pc-windows-msvc" = "10_x64", "aarch64-pc-windows-msvc" = "Server10_arm64" } } }
```
NateD-MSFT, line 438:
```suggestion
env = { "WDK_BUILD_INF2CAT_OS" = { source = "${CARGO_MAKE_CRATE_TARGET_TRIPLE}", default_value = "UNKNOWN", mapping = { "x86_64-pc-windows-msvc" = "10_NI_X64", "aarch64-pc-windows-msvc" = "10_NI_ARM64" } } }
```

NateD-MSFT: I assume we want inf2cat to use the rules for the current version of Windows.

crates/wdk-build/src/lib.rs
```diff
@@ -472,118 +488,203 @@ impl Config {
...
-        // Disable Manifest File Generation
-        println!("cargo::rustc-cdylib-link-arg=/MANIFEST:NO");
+        // Emit linker search paths
+        for path in library_paths {
+            println!("cargo::rustc-link-search={}", path.display());
+        }
 
         match &self.driver_config {
-            DriverConfig::WDM() => {
+            DriverConfig::Wdm => {
```
NateD-MSFT, line 680: Not something that needs to be changed in this PR, but in the future let's see if we can refactor all this target-specific logic out of raw code and into a declarative manner.  (Maybe I need to build a deserializer for the MSBuild proj files directly?)

crates/wdk-build/src/lib.rs
```diff
@@ -605,116 +714,269 @@ impl Config {
...
-            "x86_64" => Some(Self::AMD64),
-            "aarch64" => Some(Self::ARM64),
+            "x86_64" => Some(Self::Amd64),
+            "aarch64" => Some(Self::Arm64),
             _ => None,
         }
     }
 }
 
+/// Find the path the the toplevel Cargo manifest of the currently executing
```
NateD-MSFT:
```suggestion
/// Find the path of the toplevel Cargo manifest of the currently executing
```

crates/wdk-build/src/lib.rs
```diff
@@ -605,116 +714,269 @@ impl Config {
...
+/// Cargo subcommand. This should resolve to either:
+/// 1. the `Cargo.toml` of the package where the Cargo subcommand (build, check,
+///    etc.) was run
+/// 2. the `Cargo.toml` provided to the `--manifest-path` argument to the Cargo
+///    subcommand
+/// 3. the `Cargo.toml` of the workspace that contains the package pointed to by
+///    1 or 2
+///
+/// The returned path should be a manifest in the same directory of the
+/// lockfile. This does not support invokations that use non-default target
```
NateD-MSFT:
```suggestion
/// lockfile. This does not support invocations that use non-default target
```

crates/wdk-build/src/lib.rs
```diff
@@ -605,116 +714,269 @@ impl Config {
...
+/// Panics if the invoked from outside a Cargo build environment
+pub fn configure_wdk_binary_build() -> Result<(), ConfigError> {
+    Config::from_env_auto()?.configure_binary_build()
+}
+
+// This currently only exports the driver type, but may export more metadata in
+// the future. `EXPORTED_CFG_SETTINGS` is a mapping of cfg key to allowed cfg
+// values
+lazy_static::lazy_static! {
+    // FIXME: replace lazy_static with std::Lazy once available: https://github.com/rust-lang/rust/issues/109736
```
NateD-MSFT, line 974: Lazy is now stable if I'm not mistaken.  Not something that needs to be fixed right this instant though.

README.md
wmmc88: At the langauge level, rust has this bring-your-own-allocator construct for no_std programs (ex. km driver). There are still quite a few shortcomings with the `wdk-alloc` crate (ex. panic on oom) so for now I'd like to not force any sort of restrictions here. 

But given that the README is meant as a "getting started", I can change the wording to not explicitly encourage custom allocators

crates/wdk-build/rust-driver-makefile.toml
wmmc88: Nope. I just didn't want to make the change in this PR since scope is already extremely large here. Will address in a followup. Tracking this in https://github.com/microsoft/windows-drivers-rs/issues/171 already

crates/wdk-build/rust-driver-makefile.toml
wmmc88: I used `10_x64` as that was the default value that the WDK resolves to when building C drivers. 

Is there a new value every WDK release (I see `NI` in your suggestion)? If so, there'd need to be some logic here to resolve the mapping here based on the WDK config.

I'm inclined to leave it as is for this PR, since it matches parity with building C drivers

crates/wdk-build/src/lib.rs
wmmc88: Yes. Ideally we'd be able to consume the WDK's various xml files (vcxproj, targets, etc) and get all of this information. There'd still need to be code here to do things to emit the correctly formatted print statements to pass the information to cargo, but all that information would be parsed from the xmls instead of hardcoded here.

> Maybe I need to build a deserializer for the MSBuild proj files directly

There would need to be work done to be able to merge all the xmls together, since many of the msbuild files reference other xml files and merge them together in some deterministic fashion. I'd imagine merging all that together, and then deserializing that merged xml using some serde xml deserializer(ex. serde_xml_rs) could work.

crates/wdk-build/src/lib.rs
wmmc88: > Lazy is now stable if I'm not mistaken. Not something that needs to be fixed right this instant though.

Its available in stable as of 1.80.0 (released 5 days ago). Although not formally stated (or enforced) I've been trying to keep support for Stable and Stable-1. 

There are also other places in the repo where lazy_static is used, that aren't supported by std::sync::Lazy. (ex. LazyLock is not available in no_std, so in those situations, we should migrate to once_cell w/ the `racy` feature)

crates/wdk-build/rust-driver-makefile.toml
NateD-MSFT: When you say the default value, did you get that from WDK props, or...?

In any event see https://learn.microsoft.com/en-us/windows-hardware/drivers/devtest/inf2cat here for the full list of supported OS flags.  You can also provide multiple values to validate for multiple different OSes.  I think at the very least we should go to 10_NI for our default here since Rust drivers are at this early stage experimental and based on our latest OS planning.

README.md
NateD-MSFT: Yeah my concern is longer term - that we may have security or compliance reasons that we require people to use ExAllocatePool* and a specific MS allocator.  But your point about the current allocator being pretty bare bones is perfectly fair.  We can revisit this down the line.

crates/wdk-build/rust-driver-makefile.toml
wmmc88: This particular inf2cat setting corresponds with the `Inf2CatWindowsVersionList` property. 

On arm64, it defaults to `Server10_$(DDKPlatform)` where `DDKPlatform` resolves to `arm64`.
On x64, it defaults to `10_$(DDKPlatform)` where `DDKPlatform` resolves to `x64`.
In GE WDK, these defaults are set in `content\Program Files\Windows Kits\10\build\10.0.26100.0\WindowsDriver.OS.Props`.

If you open one of the sample windows drivers (ex. `umdf2toaster.sln`) in VS22, it confirms the above:
**x64:**
image
image
**arm64**
image
image

NateD-MSFT: For others reading this review: we resolved this offline and this is not blocking.

NateD-MSFT (approved):

crates/wdk-build/src/cargo_make.rs
wmmc88: fixed

README.md
wmmc88: I've updated the verbiage a bit to not specifically call out the functionality to bring in your own rust allocators

crates/wdk-build/src/lib.rs
wmmc88: tracking this in https://github.com/microsoft/windows-drivers-rs/issues/192

wmmc88:
Last round of changes for this PR fixed a logic bug that broke automatic skipping of the packaging task for non-binary(ie driver) wdk packages, added more tests for that specific situation, fixed various issues identified by pipelines(ie. formatting, clippy), and expanded coverage of the pipelines to include the workspace-level tests and examples folders.

crates/wdk-build/rust-driver-makefile.toml
```diff
@@ -391,64 +523,48 @@ run_task = "signtool-verify"
...
 
 [tasks.package-driver-flow]
-# Only run flow if the current package is marked as a driver
+# Note: Dependencies are always run, regardless of the condition_script result. This allows `cargo make` in mixed driver/non-driver workspaces
+dependencies = ["build"]
+# Only run package-driver flow if the current package is marked as a driver
+plugin = "nested-cargo-workspace-in-cargo-make-emulated-workspace-support"
+condition_script_runner_args = [
+  "--base-path",
+  "${CARGO_MAKE_CURRENT_TASK_INITIAL_MAKEFILE_DIRECTORY}",
```
wmmc88: The proper expansion of this is blocked on https://github.com/sagiegurari/cargo-make/pull/1132

wmmc88: added workaround to unblock pr

crates/wdk-build/rust-driver-makefile.toml
```diff
@@ -41,28 +44,115 @@ if is_file ${filepath}
...
+
+    # Re-run the task in a forked subprocess, allowing cargo-make to run in a workspace context (i.e. running on each of the members of the Cargo workspace)
+    echo Executing \"${task.name}\" Task in a forked subprocess to run on Cargo workspace: ${CARGO_MAKE_CRATE_CURRENT_WORKSPACE_MEMBER}
+
+    # Unset the CARGO_MAKE_CRATE_CURRENT_WORKSPACE_MEMBER env var to allow cargo-make's workspace detection to run 
+    unset_env CARGO_MAKE_CRATE_CURRENT_WORKSPACE_MEMBER
+
+    cm_plugin_run_custom_task "{\"run_task\":{\"name\":\"${task.name}\",\"fork\":true}}"
+else
+    # FIXME: modifying the taskjson here is not needed after upstream fix is merged: https://github.com/sagiegurari/cargo-make/pull/1132
```
wmmc88, line 137: This should be removed once https://github.com/sagiegurari/cargo-make/pull/1132 is merged. The entirety of [2d199bc9e2b12e29f329fc9377ef31f498d35335 commit](https://github.com/microsoft/windows-drivers-rs/commit/2d199bc9e2b12e29f329fc9377ef31f498d35335) can be reverted after upstream release

wmmc88: tracked in #194

crates/wdk-build/src/metadata/error.rs
```diff
@@ -0,0 +1,57 @@
...
+        value_being_serialized: String,
+    },
+
+    /// error emitted when duplicate key names are found during
+    /// serialization. Serializing into a
+    /// [`metadata::Map`](crate::metadata::Map) requires unique key names
+    #[error(
+        "duplicate keys found during serialization:\nkey: {key}\nvalue 1: {value_1}\nvalue 2: \
+         {value_2}"
+    )]
```
ayodejiige, line 40: what about if there are more than two values of the same key? is the intention to just catch the first duplicate?

crates/wdk-build/src/cargo_make.rs
```diff
@@ -754,16 +842,155 @@ fn load_wdk_build_makefile<S: AsRef<str> + AsRef<Utf8Path> + AsRef<Path>>(
...
+        Ok(())
+    })
+}
+
+/// `cargo-make` condition script for `package-driver-flow` task in
+/// [`rust-driver-makefile.toml`](../rust-driver-makefile.toml)
+///
+/// # Errors
+///
+/// This function returns an error whenever it determins that the
```
ayodejiige: determines

ayodejiige (dismissed):

crates/wdk-build/src/metadata/error.rs
wmmc88: It fails as soon as it finds the first duplicate. In the context of where this sits in the serializer apis and how those apis are invoked, I don't actually think its possible for a duplicate key to be output from the serializer. `-` is used as the seperator between rust identifiers, and rust identifiers aren't allowed to have `-` in them (even raw identifiers can't).

wmmc88: I had written this as an error instead of a panic because I actually wrote a deserializer as well that could have introduced this as a possible error condition. I ended up removing the deserializer from the scope of this PR, but plan to add it in a future followup

crates/wdk-build/src/cargo_make.rs
wmmc88: fixed

ayodejiige (approved):

hamzamust (approved):

PR #188, opened 2024-07-27
fix: bug in provided file content for `build.rs`

### LINKED ISSUE #187: Minor bug in `build.rs` content provided by README
The section of the README describing how to create a new `windows-driver-rs` project has a minor bug in step 6.
It provides the following code for the `build.rs` file.

```rust
fn main() -> Result<(), wdk_build::ConfigError> {
   wdk_build::Config::from_env_auto()?.configure_binary_build()?;
   Ok(())
}
``` 
The issue here is the last `?` after `configure_binary_build`. This function returns `()` and therefore does not implement `Try` which results in an error. Simply removing the last `?` resolves the issue.

kedom1337:
fixes #187

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ The crates in this repository are available from [`crates.io`](https://crates.io
 
    ```rust
    fn main() -> Result<(), wdk_build::ConfigError> {
-      wdk_build::Config::from_env_auto()?.configure_binary_build()?;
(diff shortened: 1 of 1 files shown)


wmmc88:
dupe of #178 . See my [comment in the issue](https://github.com/microsoft/windows-drivers-rs/issues/187#issuecomment-2256422279)

PR #229, opened 2024-10-05
ci: fix wrong `RUSTFLAGS` being used by `cargo-make` in CI

wmmc88:
Previously, [a bug in cargo make](https://github.com/sagiegurari/cargo-make/discussions/1169) caused the `RUSTFLAGS` env variable set in CI to be ignored. This meant that certain targets were compiled without `-D warnings`. Now that the bug has been fixed in `cargo-make`, the `RUSTFLAGS` env var is properly applying but the static crt setting was missing


wmmc88:
Nightly Clippy failures to be addressed by #230

ankurung (approved):

ayodejiige (approved):

PR #232, opened 2024-10-07
feat: automatically install test-certificate and validate they are installed on LocalMachine for `signtool-verify` tasks

wmmc88:
Users often make the mistake of installing the test-signed certificate to `Current User` store instead of `Local Machine`. The `signtool-verify` cargo-make tasks were originally added to help diagnose these issues.

Using `/kp` with `signtool` to validate kernel signing policy is not possible since the test-cert will terminate at a root that isn't trusted by windows: https://learn.microsoft.com/en-us/windows-hardware/drivers/install/test-signing . Using `/pa` is not sufficient since it will pass even when the certificate is only installed in `Current User`. Additional checks were added to make sure that on top of `/pa` passing, the certificate is also installed in the `Local Machine` stores

### Certificate Management Enhancements:
* Added new tasks for installing and validating certificates in the `rust-driver-makefile.toml` file. This includes the `install-certificate` and `validate-certificate-installed` tasks, which ensure that the `WDRLocalTestCert` certificate is correctly installed and validated per Windows requirements. (`crates/wdk-build/rust-driver-makefile.toml`) [[1]](diffhunk://#diff-257cffc3109c5d92855f7d0f253ea648b0bb4317b4d2708bdb3dfb7cf249d3a3R437-R486) [[2]](diffhunk://#diff-257cffc3109c5d92855f7d0f253ea648b0bb4317b4d2708bdb3dfb7cf249d3a3R539)
* Implemented functions in `cargo_make.rs` for handling certificate installation, validation, and checking if the process is running with admin privileges. (`crates/wdk-build/src/cargo_make.rs`)
* Updated `README.md` to include instructions for installing and validating certificates. (`README.md`)

### Workflow and Utility Enhancements:
* Updated the GitHub Actions workflow to enable `WDK_BUILD_ENABLE_SIGNTOOL_VERIFY` for signature verification. (`.github/workflows/build.yaml`)
* Added utility functions in `utils.rs` for checking admin privileges using Windows APIs. (`crates/wdk-build/src/utils.rs`)

closes https://github.com/microsoft/Windows-rust-driver-samples/issues/30

### DIFF
--- a/crates/wdk-build/src/cargo_make.rs
+++ b/crates/wdk-build/src/cargo_make.rs
@@ -18,7 +18,7 @@ use std::{
     io,
     panic::UnwindSafe,
     path::{Path, PathBuf},
-    process::Command,
+    process::{self, Command},
 };
 
 use anyhow::{bail, Context};
@@ -1271,21 +1271,12 @@ fn is_wdrlocaltestcert_installed() -> anyhow::Result<bool> {
 
     // Export WDRLocalTestCert from "Trusted Root Certificate Authorities" store
     // (Local Machine)
-    let output = Command::new("certmgr.exe")
-        .arg("-put")
-        .arg("-c")
-        .arg("-s")
-        .arg("-r")
-        .arg("localMachine")
-        .arg("ROOT")
-        .arg("-n")
-        .arg("WDRLocalTestCert")
-        .arg(&exported_root_cert_path)
-        .output()?;
+    let output = export_certificate("localMachine", "ROOT", &exported_root_cert_path)?;
     if !output.status.success() {
         println!(
             "WDRLocalTestCert not found in Trusted Root Certificate Authorities store (Local \
-             Machine): \n\nstdout:\n{}\nstderr:\n{}",
+             Machine). It is expected that the generate-certificate task has installed the \
+             signing certificate there: \n\nstdout:\n{}\nstderr:\n{}",
             String::from_utf8_lossy(&output.stdout),
             String::from_utf8_lossy(&output.stderr)
         );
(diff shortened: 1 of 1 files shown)


crates/wdk-build/rust-driver-makefile.toml
```diff
@@ -487,7 +487,7 @@ run_task = "signtool-sign"
 private = true
 condition = { env_true = ["WDK_BUILD_ENABLE_SIGNTOOL_VERIFY"] }
 command = "signtool"
-args = ["verify", "/v", "/pa", "${WDK_BUILD_SIGNTOOL_VERIFY_INPUT_FILE}"]
+args = ["verify", "/v", "/kp", "${WDK_BUILD_SIGNTOOL_VERIFY_INPUT_FILE}"]
```
wmmc88: /kp is actually not appropriate to use given that test-signed certs will result in the following error: 

```
SignTool Error: Signing Cert does not chain to a Microsoft Root Cert.
```

crates/wdk-build/src/utils.rs
```diff
@@ -276,6 +289,84 @@ pub fn get_wdk_version_number<S: AsRef<str> + ToString + ?Sized>(
...
+        assert!(
+            MIN_SUBAUTHORITY_COUNT <= ADMINISTRATORS_GROUP_SUBAUTHORITIES.len()
+                && ADMINISTRATORS_GROUP_SUBAUTHORITIES.len() <= MAX_SUBAUTHORITY_COUNT
+        )
+    };
+
+    let subauthorities = const {
+        let mut subauthorities = [0_u32; MAX_SUBAUTHORITY_COUNT];
+
+        // While loop since iterators are not const evaluatable
```
wmmc88:
```suggestion
        // Use while loop since iterators are not const evaluable
```

crates/wdk-build/src/cargo_make.rs
```diff
@@ -1106,6 +1112,265 @@ pub fn driver_sample_infverif_condition_script() -> anyhow::Result<()> {
...
+/// This function returns an error whenever it determines that the
+/// `install-certifiacte` `cargo-make` task should be skipped (i.e. when the
+/// certificate is already installed)
+///
+/// # Panics
+/// Panics if `CARGO_MAKE_CURRENT_TASK_NAME` is not set in the environment
+pub fn install_certificate_condition_script() -> anyhow::Result<()> {
+    condition_script(|| {
+        if is_wdrlocaltestcert_installed()? {
+            let cargo_make_task_name = env::var(CARGO_MAKE_CURRENT_TASK_NAME_ENV_VAR)
```
wmmc88, line 1129: Cargo make needs uprev to min 0.37.22 to get  bugfix

PR #233, opened 2024-10-08
Use stack-based formatter for debug-printing.

Zero-Tang:
This will enable `print!` and `println!` macros to be used in `<=DIRQL` levels.
If `alloc` crate is used, `ExAllocatePool2` will fail in `>DISPATCH_IRQL` levels.


Zero-Tang:

Zero-Tang:
It's been a few weeks since this PR is opened. Someone please review this PR.

Zero-Tang:
Ping: It's been a month since I opened this PR. Someone please review this PR.

wmmc88:
> Ping: It's been a month since I opened this PR. Someone please review this PR.

Hey @Zero-Tang. Thanks for the PR. I've already gone thru it as a surface level and it LGTM, but we're having some discussions about it internally, as well as going thru some testing with it. I will get this merged in when that has concluded.

crates/wdk/src/print.rs
```diff
@@ -33,20 +85,220 @@ macro_rules! println {
...
-    unsafe {
-        #[cfg(any(driver_model__driver_type = "WDM", driver_model__driver_type = "KMDF"))]
-        {
-            wdk_sys::ntddk::DbgPrint(formatted_string.as_ptr());
+pub fn _print(args: fmt::Arguments) {
+    cfg_if::cfg_if! {
+        if #[cfg(any(driver_model__driver_type = "WDM", driver_model__driver_type = "KMDF"))] {
+            let mut buffered_writer = dbg_print_buf_writer::DbgPrintBufWriter::new();
+
+            // TODO: handle internal bytes?
```
wmmc88: @leon-xd

Rust strings passed to `print!`/`println!` can potentially include null bytes in them, as that is valid in Rust strings. The current implementation doesn't handle these internal null byte, and I think the string just ends up truncated by DbgPrint on the first null. 

`DbgPrintBufWriter`'s `fmt::Write` implementation can be updated to handle these situations similar to how I changed this PR to handle it in UMDF scenario

crates/wdk/src/print.rs
```diff
@@ -33,20 +85,220 @@ macro_rules! println {
...
+mod dbg_print_buf_writer {
+    use core::fmt;
+
+    /// Max size that can be transmitted by `DbgPrint` in single call:
+    /// <https://learn.microsoft.com/en-us/windows-hardware/drivers/debugger/reading-and-filtering-debugging-messages#dbgprint-buffer-and-the-debugger>
+    const DBG_PRINT_MAX_TXN_SIZE: usize = 512;
+
+    // We will allocate the format buffer on stack instead of heap
+    // so that debug printer won't be subject to DISPATCH_IRQL restriction.
+
```
wmmc88:
```suggestion
```

crates/wdk/src/print.rs
```diff
@@ -33,20 +85,220 @@ macro_rules! println {
...
+            // that gets called for flushing (real impl uses Dbgprint and test impl uses a
+            // mock with a counter and some way to validate contents being sent to the flush
+            // closure)
+
+            writer.flush();
+            assert_eq!(writer.used, 0);
         }
     }
+
+    // FIXME: add tests for no internal null bytes, string w/ exactly USABLE
```
wmmc88: @leon-xd please add these extra test scenarios

crates/wdk/src/print.rs
```diff
@@ -33,20 +85,220 @@ macro_rules! println {
...
+
+            let mut writer = DbgPrintBufWriter::new();
+            fmt::write(&mut writer, format_args!("{TEST_STRING}"))
+                .expect("fmt::write should succeed");
+            assert_eq!(writer.used, expected_unflushed_string_contents.len());
+            assert_eq!(
+                &writer.buffer[..writer.used],
+                expected_unflushed_string_contents.as_bytes()
+            );
+            // FIXME: When this test is compiled, rustc automatically links the
```
wmmc88, line 321: @leon-xd this one doesn't necessarily need to be addressed in this PR. can handle based off your bandwidth

wmmc88 (changes requested):
Thanks for the initial PR @Zero-Tang - once this goes in, it'll contribute to being a much better experience for developing Rust Windows Drivers! 

We ran into some issues while testing this PR with various corner-cases, and I've made some enhancements to it to accommodate that, along with several extra new features. My colleague @leon-xd will take the last bit of this PR to completion before we make a new release of windows-drivers-rs crates this week

crates/wdk/src/print.rs
leon-xd: fixed

crates/wdk/src/print.rs
leon-xd: added

crates/wdk/src/print.rs
leon-xd: Bandwidth was low so going to create an issue and link to this task

crates/wdk/src/print.rs
wmmc88: remove the TODO?

Cargo.lock
```diff
@@ -304,9 +304,9 @@ dependencies = [
 
 [[package]]
 name = "linux-raw-sys"
-version = "0.4.15"
+version = "0.4.14"
```
wmmc88: Why downgrade this?

crates/wdk/src/print.rs
```diff
@@ -161,25 +158,42 @@ mod dbg_print_buf_writer {
         fn write_str(&mut self, s: &str) -> fmt::Result {
             let mut str_byte_slice = s.as_bytes();
             let mut remaining_buffer = &mut self.buffer[self.used..Self::USABLE_BUFFER_SIZE];
-            let mut remaining_buffer_size = remaining_buffer.len();
+            let mut remaining_buffer_len = remaining_buffer.len();
+            
+            if let Some(first_non_null_byte_pos) = str_byte_slice.iter().position(|&b| b != 0u8) {
```
wmmc88:
```suggestion
            if let Some(first_non_null_byte_pos) = str_byte_slice.iter().position(|&b| b != b"\0") {
```

crates/wdk/src/print.rs
```diff
@@ -196,6 +210,9 @@ mod dbg_print_buf_writer {
         }
 
         pub fn flush(&mut self) {
+            // Null-terminate the string
+            self.buffer[self.used] = 0;
+
             // SAFETY: This is safe because:
             // 1. `self.buffer` contains a valid C-style string with the data placed in
             //    [0..self.used] by the `write_str` implementation
```
wmmc88, line 217: this is actually guaranteed by the above null term now

crates/wdk/src/print.rs
```diff
@@ -277,14 +294,15 @@ mod dbg_print_buf_writer {
             let expected_unflushed_string_contents =
                 &TEST_STRING[UNFLUSHED_STRING_CONTENTS_STARTING_INDEX..];
 
-            let mut writer = DbgPrintBufWriter::new();
+            let mut writer: DbgPrintBufWriter = DbgPrintBufWriter::new();
```
wmmc88:
```suggestion
            let mut writer = DbgPrintBufWriter::new();
```

crates/wdk/src/print.rs
```diff
@@ -277,14 +294,15 @@ mod dbg_print_buf_writer {
...
             fmt::write(&mut writer, format_args!("{TEST_STRING}"))
                 .expect("fmt::write should succeed");
             assert_eq!(writer.used, expected_unflushed_string_contents.len());
             assert_eq!(
                 &writer.buffer[..writer.used],
                 expected_unflushed_string_contents.as_bytes()
             );
+            let old_writer_used = writer.used;
```
wmmc88:
```suggestion
            let expected_null_byte_position = writer.used;
```

wmmc88: ditto on other tests

Cargo.lock
leon-xd: This slipped by me in fixing the lockfiles, bumping back to 15

crates/wdk/src/print.rs
leon-xd: This only works if you convert a char to a byte, not a string (`b'\0'` instead of `b"\0"`). Fixed.

crates/wdk/src/print.rs
leon-xd: removed in all tests

crates/wdk/src/print.rs
leon-xd: fixed. think the follow up comment was meant to be put on a different comment since this is in the `flush` method and not in any tests.

leon-xd: fixed

crates/wdk/src/print.rs
```diff
@@ -33,20 +85,396 @@ macro_rules! println {
...
-        .expect("CString should be able to be created from a String.");
+pub fn _print(args: fmt::Arguments) {
+    cfg_if::cfg_if! {
+        if #[cfg(any(driver_model__driver_type = "WDM", driver_model__driver_type = "KMDF"))] {
+            let mut buffered_writer = dbg_print_buf_writer::DbgPrintBufWriter::new();
+
+            if let Ok(_) = fmt::write(&mut buffered_writer, args) {
+                buffered_writer.flush();
+            } else {
+                unreachable!("DbgPrintBufWriter should never fail write");
```
gurry:
```suggestion
                unreachable!("DbgPrintBufWriter should never fail to write");
```

crates/wdk/src/print.rs
```diff
@@ -33,20 +85,396 @@ macro_rules! println {
...
         }
+    }
+
+    impl fmt::Write for DbgPrintBufWriter {
+        fn write_str(&mut self, s: &str) -> fmt::Result {
+            let mut str_byte_slice = s.as_bytes();
+            let mut remaining_buffer = &mut self.buffer[self.used..Self::USABLE_BUFFER_SIZE];
+            let mut remaining_buffer_len = remaining_buffer.len();
+
+            if let Some(first_non_null_byte_pos) = str_byte_slice.iter().position(|&b| b != b'\0') {
```
gurry: This code is duplicated below. Consider refactoring into a function. That function can be declared locally within this function itself.

crates/wdk/src/print.rs
```diff
@@ -33,20 +85,396 @@ macro_rules! println {
...
+            Self {
+                // buffer is initialized to all null
+                buffer: [0; DBG_PRINT_MAX_TXN_SIZE],
+                used: 0,
+            }
         }
+    }
+
+    impl fmt::Write for DbgPrintBufWriter {
+        fn write_str(&mut self, s: &str) -> fmt::Result {
```
gurry, line 160: Consider adding a comment on this function explaining how it works

crates/wdk/src/print.rs
```diff
@@ -33,20 +85,396 @@ macro_rules! println {
...
+    pub struct DbgPrintBufWriter {
+        buffer: [u8; DBG_PRINT_MAX_TXN_SIZE],
+        used: usize,
+    }
+
+    impl Default for DbgPrintBufWriter {
+        fn default() -> Self {
+            Self {
+                // buffer is initialized to all null
+                buffer: [0; DBG_PRINT_MAX_TXN_SIZE],
```
gurry, line 149: Just a thought, instead of 512 bytes can we get away with something smaller like 256 and save on stack space. Here's 256 x's just to get a sense:

`xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx`

I believe a vast majority of print messages would be shorter than this.

crates/wdk/src/print.rs
```diff
@@ -33,20 +85,396 @@ macro_rules! println {
...
+        /// strings
+        ///
+        /// The last byte is reserved for the null terminator
+        const USABLE_BUFFER_SIZE: usize = DBG_PRINT_MAX_TXN_SIZE - 1;
+
+        pub fn new() -> Self {
+            Self::default()
+        }
+
+        pub fn flush(&mut self) {
```
gurry, line 206: Consider adding a comment on this function, only if you think it will help otherwise ignore this

crates/wdk/src/print.rs
```diff
@@ -33,20 +85,396 @@ macro_rules! println {
...
+            // 1. `self.buffer` contains a valid C-style string with the data placed in
+            //    [0..self.used] by the `write_str` implementation
+            // 2. The `write_str` method ensures `self.used` never exceeds
+            //    `USABLE_BUFFER_SIZE`, leaving the last byte available for null termination
+            // 3. The "%s" format specifier is used as a literal string to prevent
+            //    `DbgPrint` from interpreting format specifiers in the message, which could
+            //    lead to memory corruption or undefined behavior if the buffer contains
+            //    printf-style formatting characters
+            unsafe {
+                wdk_sys::ntddk::DbgPrint(
```
gurry, line 225: We can avoid calling `DbgPrint` if `self.used` is `0`. Will save on an unnecessary call.

crates/wdk/src/print.rs
```diff
@@ -33,20 +85,396 @@ macro_rules! println {
...
-pub fn _print(args: core::fmt::Arguments) {
-    let formatted_string = CString::new(alloc::format!("{args}"))
-        .expect("CString should be able to be created from a String.");
+pub fn _print(args: fmt::Arguments) {
+    cfg_if::cfg_if! {
+        if #[cfg(any(driver_model__driver_type = "WDM", driver_model__driver_type = "KMDF"))] {
+            let mut buffered_writer = dbg_print_buf_writer::DbgPrintBufWriter::new();
+
+            if let Ok(_) = fmt::write(&mut buffered_writer, args) {
+                buffered_writer.flush();
```
gurry, line 94: It may be better if `DbgPrintBufWriter::write_str` itself calls `flush` at the end instead of leaving this responsibility on the caller who can forget to do so.

crates/wdk/src/print.rs
```diff
@@ -33,20 +85,396 @@ macro_rules! println {
...
+
+                if let Some(first_non_null_byte_pos) =
+                    str_byte_slice.iter().position(|&b| b != b'\0')
+                {
+                    str_byte_slice = &str_byte_slice[first_non_null_byte_pos..];
+                } else {
+                    str_byte_slice = &str_byte_slice[str_byte_slice.len()..];
+                }
+
+                // Flush buffer if full , otherwise update amount used
```
gurry:
```suggestion
                // Flush buffer if full, otherwise update amount used
```

crates/wdk/src/print.rs
wmmc88: i think it's fine to keep outside since this matches how `std::io::BufWriter` works (requires an explicit flush at the end). While `dbg_print_buf_writer` is currently not exported privately, it may make sense to export it publicly in the future to more generally allow KM string formatting without heap allocs (rn its only available for these dbgprints)

crates/wdk/src/print.rs
leon-xd: 512 is the buffer size cap for `DbgPrint`, hence why we chose this number.  We are exposing this through the print macros here so it's probably best to keep full functionality.

[DbgPrint documentation](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-dbgprint)

crates/wdk/src/print.rs
leon-xd: Fixed

crates/wdk/src/print.rs
leon-xd: Agree with @wmmc88, this was intentionally kept and this behavior is specifically checked by our test cases.

crates/wdk/src/print.rs
leon-xd: Added

crates/wdk/src/print.rs
leon-xd: Added

crates/wdk/src/print.rs
leon-xd: Added

crates/wdk/src/print.rs
leon-xd: Refactored

crates/wdk/src/print.rs
leon-xd: Fixed

crates/wdk/src/print.rs
gurry: 256 being smaller will not violate `DbgPrint`'s 512 cap and it should not restrict the functionality in any way. I'm not wedded to this idea, but as stack space is limited in KM, it seems like a good tradeoff.

@wmmc88 What are your thoughts?

crates/wdk/src/print.rs
gurry: One key difference from `std::io::BufWriter` is that `std::io::BufWriter` is never sure if there's more data to come in subsequent calls, That's why it has to expose `flush` to allow the user to indicate they are done and the data can be pushed upstream.

In our case every call is self contained. It carries the full message to be written and the writer can therefore go ahead and flush before it returns. In fact it _has to_ flush because otherwise it won't function correctly (although we're relying on the user to do that for us).

That said, I don't want to unnecessarily delay the PR. As this is an internal type we can always come back to it if needed.

crates/wdk/src/print.rs
```diff
@@ -153,16 +153,16 @@ mod dbg_print_buf_writer {
...
             let mut str_byte_slice = s.as_bytes();
             let mut remaining_buffer = &mut self.buffer[self.used..Self::USABLE_BUFFER_SIZE];
             let mut remaining_buffer_len = remaining_buffer.len();
 
-            if let Some(first_non_null_byte_pos) = str_byte_slice.iter().position(|&b| b != b'\0') {
-                str_byte_slice = &str_byte_slice[first_non_null_byte_pos..];
-            } else {
-                return Ok(());
-            }
+            str_byte_slice = advance_slice_to_next_non_null_byte(str_byte_slice);
```
gurry, line 165: Before the refactoring we were returning from `write_str` here if we could not find any non-null char. Perhaps we should retain that behaviour.

crates/wdk/src/print.rs
wmmc88: I think we can stick with 512B for now. IIRC kernel stack is 24KB, so the call will consume ~2% of the stack.

In the future, we can revisit this. I would like in the future for this `fmt::write` implementation to have a configurable generic buffer size when it gets refactored to allow general (ie. non Dbgprint-specific) string formatting on the stack

gurry: Makes sense. Agreed.

gurry (approved):

wmmc88 (dismissed):

leon-xd:
Discovered a bug in my logic -- conditionally updating `used` prevented `flush` from null terminating the string in the correct location. This is updated in my latest commit.

All print functions are working as anticipated, tested via UMDF, KMDF, and WDM sample drivers.

wmmc88 (dismissed):

gurry:
> Discovered a bug in my logic -- conditionally updating `used` prevented `flush` from null terminating the string in the correct location. This is updated in my latest commit.
> 
> All print functions are working as anticipated, tested via UMDF, KMDF, and WDM sample drivers.

Consider adding a test that would have failed due to this bug so that we don't have the bug again in the future.

gurry (dismissed):

gurry (dismissed):

wmmc88 (approved):

wmmc88 (approved):

gurry (approved):

Zero-Tang:
I'm glad to see this PR is coming to a close.

leon-xd:
> I'm glad to see this PR is coming to a close.

@Zero-Tang thank you for your contribution (and your patience)!

PR #243, opened 2024-11-13
Rust Versioning and package metadata integration

CHMANG:
Includes automation capability to generate Resource files for both WDM and WDF driver development models by using cargo metadata.
Compile the Resource file using Resource compiler and link the compiled resource file into the application driver executable using the linker.


crates/wdk-build/Cargo.toml
```diff
@@ -11,6 +11,8 @@ categories = ["development-tools::build-utils", "development-tools::ffi"]
 
 [build-dependencies]
 rustversion.workspace = true
+serde = "1.0"
+serde_json = "1.0"
```
wmmc88, line 15: These should be workspace dependencies

crates/wdk-build/src/lib.rs
```diff
@@ -313,6 +315,39 @@ impl Config {
...
+    /// the architecture platform
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if any of the required paths do not
+    /// exist.
+    pub fn get_rc_root_path(&self) -> Result<String, ConfigError> {
+        let bin_directory = self.wdk_content_root.join("bin");
+    
+        // Add Windows SDK library paths based on logic from WindowsDriver.KernelMode.props &
```
wmmc88: This comment is copy-pasted. please clean up so that it makes sense

crates/wdk-build/src/lib.rs
```diff
@@ -313,6 +315,39 @@ impl Config {
...
     }
 
+    /// Returns Resource Compile path to build and link based off of
+    /// the architecture platform
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if any of the required paths do not
+    /// exist.
+    pub fn get_rc_root_path(&self) -> Result<String, ConfigError> {
```
wmmc88, line 325: This looks like the same logic as in: https://github.com/CHMANG/windows-drivers-rs/blob/37f74cae7207571347706806121088917f5040ba/crates/wdk-build/src/cargo_make.rs#L516-L517 

Can this logic be commonized to a single place in `utils` as a free function that takes in a wdk content root path? that way it can be called in both `setup_path` and in this function.

also can this Config method be named `get_bin_path` since all the arch-specific bins are here.

crates/wdk-build/src/lib.rs
```diff
@@ -323,54 +358,51 @@ impl Config {
...
             });
         }
         include_paths.push(
-            kit_shared_include_path
+            um_include_path
                 .canonicalize()?
                 .strip_extended_length_path_prefix()?,
         );
-
+
```
wmmc88, line 402: Why are you changing the include path logic? This changes how all the bindings are generated and break it.

The logic here strictly matches what happens in the vcxproj files of the WDK. If RC compilation requires different includes, it should be handled elsewhere

crates/wdk-build/src/resource_compile.rs
```diff
@@ -0,0 +1,217 @@
...
+        productversion,
+        description, 
+        fileversion, 
+        name, 
+        &includeargs,
+        rc_exe_rootpath,
+    );
+}
+// function to get and set RC File with package metadata
+fn getandset_rcfile(
```
wmmc88: please follow rust conventions for function names

examples/sample-wdm-driver/Cargo.toml
```diff
@@ -12,6 +12,9 @@ publish = false
 
 [package.metadata.wdk.driver-model]
 driver-type = "WDM"
+companyname = "Microsoft Corporation"
+copyright = "(C) 2024 Microsoft. All rights reserved."
+productname = "Surface"
```
wmmc88, line 17: These fields should be using kebab case per toml conventions

wmmc88: They also should not be a part of the driver-model metadata section. The `driver-model` section contains things in the driver model tab of visual studio. I think you probably want a separate metadata section for this? What do you think @ayodejiige

crates/wdk-build/src/resource_compile.rs
```diff
@@ -0,0 +1,217 @@
...
+    copyright: String, 
+    productname: String, 
+    productversion:String, 
+    description:String, 
+    fileversion:String, 
+    name:String, 
+    includeargs: &Vec<String>, 
+    rc_exe_rootpath:String,
+) {
+    println!("Set and create rc file... ");
```
wmmc88, line 54: rather than using `println` and `eprintln`, please using `tracing` crate's [macros for info, error, warn, etc.](https://docs.rs/tracing/latest/tracing/#macros)

crates/wdk-build/src/resource_compile.rs
```diff
@@ -0,0 +1,217 @@
...
+            // Append "/I" and the include path to the modified vector
+            includeargs.push("/I".to_string());
+            includeargs.push(include_str.to_string());
+        } else {
+            println!("Non-Unicode path is not supported: {:?}", include_path);
+        }
+    }
+
+    let (companyname, copyright, productname) = get_packagemetadatadetails();
+    let (productversion, description, fileversion, name) = get_packagedetails();
```
wmmc88: Rather than use tuples, please use purpose-built structs for type safety

crates/wdk-build/src/resource_compile.rs
```diff
@@ -0,0 +1,217 @@
...
+    (companyname, copyrightname, productname)
+}
+// function to get package details
+fn get_packagedetails() -> (String, String, String, String) {
+    let mut fileversion = String::new();
+    let mut description = String::new();
+    let mut productversion = String::new();
+    let mut name = String::new();
+
+    match fs::read_to_string("Cargo.toml") {
```
wmmc88, line 195: This is very hacky to get the crate version. You should use the env vars that cargo provides you: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

crates/wdk-build/src/resource_compile.rs
```diff
@@ -0,0 +1,217 @@
...
+#undef  VER_PRODUCTNAME_STR
+#define VER_PRODUCTNAME_STR    {productname}
+
+#include "common.ver""#,
+        file_type = ver_filetype,
+        file_subtype = ver_filesubtype,
+        original_filename = ver_originalfilename
+    );
+   
+    std::fs::write("resources.rc", rc_content).expect("Unable to write RC file");
```
wmmc88, line 111: error handling is needed in this module rather than panicking

crates/wdk-build/src/resource_compile.rs
```diff
@@ -0,0 +1,217 @@
...
+            eprintln!("Error running rc.exe: {}", err);
+            std::process::exit(1); // Exit with a non-zero status code
+        }
+    }
+}
+// function to get package metadata details
+fn get_packagemetadatadetails() -> (String, String, String) {
+    // Run the 'cargo metadata' command and capture its output
+    let path = env::var("CARGO_MANIFEST_DIR").unwrap();
+    let meta = MetadataCommand::new()
```
wmmc88, line 155: Rather than do additional parsing here, do you think it would be better to bake this into the WDK config struct @ayodejiige ?

crates/wdk-build/src/resource_compile.rs
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    let ver_filetype = "VFT_DRV";
+    let ver_filesubtype = "VFT2_DRV_SYSTEM";
+    let ver_originalfilename = "VER_INTERNALNAME_STR";
+
+    // Create the RC file content
+    let rc_content = format!(
+        r#"#include <windows.h>
+#include <ntverp.h>
```
wmmc88, line 74: Is there an MSDN doc that this template is pulled from? Are these all required? optional?

crates/wdk-build/src/resource_compile.rs
```diff
@@ -0,0 +1,217 @@
+use std::{
+    env,
+    fs,
+    path::{Path, PathBuf},
+    process::Command,
+};
+use cargo_metadata::MetadataCommand;
```
wmmc88, line 7: I think this file should probably be restructured such that it is some ResourceCompile builder, rather than a set of free functions. I think the process of compiling the rc file should be separated from the process of writing the rc file from a template to filesystem. This way, we could reuse it for different rc files, if needed

wmmc88 (changes requested):

PR #246, opened 2024-11-16
feat: Enable build for INF only drivers

ayodejiige:


crates/wdk-build/src/metadata/driver_settings.rs
```diff
@@ -0,0 +1,129 @@
...
+)]
+pub enum DriverConfig {
+    /// Windows Driver Model
+    Wdm,
+    /// Kernel Mode Driver Framework
+    Kmdf(KmdfConfig),
+    /// User Mode Driver Framework
+    Umdf(UmdfConfig),
+    /// INF only drivers e.g. null drivers and extension INFs
+    Package
```
wmmc88: Circling back on our earlier conversations, the WDK does use Package as a "driver type": https://github.com/microsoft/Windows-driver-samples/blob/ea026ed7e82fe5b45f0d978661e68c7f3249f3fc/general/toaster/umdf2/Package/package.VcxProj#L32C6-L32C16

crates/wdk-build/rust-driver-makefile.toml
```diff
@@ -303,6 +303,24 @@ args = [
 private = true
```
wmmc88, line 303: https://github.com/microsoft/windows-drivers-rs/blob/10a8e3742b7834e816f6a2ce630b46533bcec357/crates/wdk-build/src/cargo_make.rs#L824

Previously, there was an assumption that there must be wdk-build in the workspace. This is no longer true now there can be a null driver in the workspace without any actual driver code being built (so no wdk-build in the graph).

In the wdk-build only case, we should also see if putting a workspace dep on wdk-build yields anything in the cargo metadata output.

maybe also worth putting a specific message for the case of no-wdk in workspace since that would typically be a null-driver-only workspace scenario

PR #250, opened 2024-12-04
refactor!: Remove lazy static instances

### LINKED ISSUE #192: Remove all lazy_static instances
> Lazy is now stable if I'm not mistaken. Not something that needs to be fixed right this instant though.

Its available in stable as of 1.80.0 (released 5 days ago). Although not formally stated (or enforced) I've been trying to keep support for Stable and Stable-1. 

There are also other places in the repo where lazy_static is used, that aren't supported by std::sync::Lazy. (ex. LazyLock is not available in no_std, so in those situations, we should migrate to once_cell w/ the `racy` feature)

_Originally posted by @wmmc88 in https://github.com/microsoft/windows-drivers-rs/pull/186#discussion_r1697631038_

leon-xd:
Replaced lazy static instances with std::sync::LazyLock in following locations:
- crates/wdk-build/src/lib.rs
- crates/wdk-sys/build.rs
- tests/wdk-macros-tests/src/lib.rs

Replaced lazy static instances with custom FunctionTable struct in generated code in `crates/wdk-sys/build.rs`

### Breaking Changes
- Removes `wdf_function_table.rs` from generated code of WDK_BUILD. Users should access the function table instead via pointer at `wdk_sys::WdfFunctions`, and checking bounds via `wdk_sys::wdf::get_wdf_function_count()` which is generated in `wdf_function_count.rs`
- lazy-static is no longer dependency in the workspace and is replaced with `sync::LazyLock` as introduced in Rust 1.83

closes #192


leon-xd:
@microsoft-github-policy-service agree company="Microsoft"

crates/wdk-sys/build.rs
```diff
@@ -47,35 +47,64 @@ const WDF_FUNCTION_COUNT_DECLARATION_EXTERNAL_SYMBOL: &str = "
...
-            core::slice::from_raw_parts(wdf_function_table, wdf_function_count)
+static WDF_FUNCTION_TABLE_TEMPLATE: LazyLock<String> = LazyLock::new(|| {
+    format!(
+        r"
+extern crate alloc;
+
+/// Struct to encapsulate a function table.
+/// Lazily initializes on access via index operation.
+pub struct FunctionTable {{
+    inner: once_cell::race::OnceBox<alloc::boxed::Box<[crate::WDFFUNC]>>
```
leon-xd: OnceBox implicitly requires its template parameter to be Sized -- therefore I wrapped the function table in a Box so that OnceBox can initialize.

leon-xd: We aren't able to use LazyLock here since the generated code is `no-std`, therefore we must use the `race` module in the `once_cell` crate. documentation can be found [here](https://docs.rs/once_cell/latest/once_cell/race/struct.OnceBox.html).

crates/wdk-sys/build.rs
```diff
@@ -47,35 +47,64 @@ const WDF_FUNCTION_COUNT_DECLARATION_EXTERNAL_SYMBOL: &str = "
...
+}}
+
+impl core::ops::Index<usize> for FunctionTable {{
+    type Output = crate::WDFFUNC;
+
+    fn index(&self, index: usize) -> &Self::Output {{
```
leon-xd: each time the table is accessed, we call `self.get_function_table()` in order to call the `get_or_init` function to lazily initialize the table.

Cargo.toml
```diff
@@ -45,7 +45,9 @@ cfg-if = "1.0.0"
 clap = "4.5.9"
 clap-cargo = "0.14.1"
 itertools = "0.13.0"
-lazy_static = "1.5.0"
+once_cell = { version = "1.20.2", default-features = false, features = [
```
leon-xd: this links std with default-features turned on

crates/wdk-build/src/lib.rs
```diff
@@ -981,14 +981,8 @@ pub fn configure_wdk_binary_build() -> Result<(), ConfigError> {
     Config::from_env_auto()?.configure_binary_build()
 }
 
-// This currently only exports the driver type, but may export more metadata in
```
wmmc88, line 1149: keep this comment and actually change it to a doc comment (ie. triple backslash)

crates/wdk-sys/Cargo.toml
```diff
@@ -74,4 +74,4 @@ redundant_explicit_links = "warn"
 unescaped_backticks = "warn"
 
 [package.metadata.cargo-machete]
-ignored = ["lazy_static"] # lazy_static is used in code generated by build.rs
```
wmmc88, line 81: keep the comment

Cargo.toml
```diff
@@ -45,7 +45,9 @@ cfg-if = "1.0.0"
...
+  "alloc",
```
wmmc88: hmm is alloc required here? this forces there to be an allocator pulled in always. packages without them would fail to compile

wmmc88: To be addressed in separate issue: we should add some samples to the top level test that explicitly test this no alloc available case. The current examples all pull in `wdk-alloc` but samples should work without it still

wmmc88: I think as it stands, this pr breaks the pure no-std case

crates/wdk-sys/build.rs
```diff
@@ -47,35 +47,64 @@ const WDF_FUNCTION_COUNT_DECLARATION_EXTERNAL_SYMBOL: &str = "
...
-        //            larger than `isize::MAX`. This is proven by the below `debug_assert!`.
-        unsafe {{
-            debug_assert!(isize::try_from(wdf_function_count * core::mem::size_of::<crate::WDFFUNC>()).is_ok());
-            core::slice::from_raw_parts(wdf_function_table, wdf_function_count)
+static WDF_FUNCTION_TABLE_TEMPLATE: LazyLock<String> = LazyLock::new(|| {
+    format!(
+        r"
+extern crate alloc;
+
+/// Struct to encapsulate a function table.
```
wmmc88: I think this type should be moved to be private somewhere in wdk-sys. you can use it in generated code still via `crate::`

crates/wdk-sys/build.rs
wmmc88: wrapping it in a box makes it do a heap alocation, which is not desireable

crates/wdk-sys/build.rs
wmmc88: also makes it so you can run tests for it

wmmc88 (changes requested):

crates/wdk-sys/Cargo.toml
leon-xd: this entire line is gone -- there's no reference to lazy_static. why do we need to keep the comment?

wmmc88: there is still a machete entry for once_cell. I want a comment documenting why machete needs to ignore once_cell (presumably because its only used in generated code that machete doesn't scan , resulting in a false positive).

wmmc88: without the comment, machete fails, right?

Cargo.toml
wmmc88: @leon-xd  this was marked as resolved, but i believe its still the case?

leon-xd: I deleted this, there's no reference to once_cell at all in the code.

crates/wdk-sys/Cargo.toml
leon-xd: i double checked both on github and my own repository, once_cell is completely removed and so is the machete-ignore entry. I think you may have been viewing from an old commit. Is there something in the build logs that you're seeing that is generating this?

tests/wdk-macros-tests/tests/outputs/nightly/macrotest/bug_unused_imports.expanded.rs
```diff
@@ -47,8 +47,26 @@ pub extern "system" fn driver_entry(
                     driver__: *mut WDFDRIVER,
                 ) -> NTSTATUS {
                     let wdf_function: wdk_sys::PFN_WDFDRIVERCREATE = Some(unsafe {
+                        let wdf_function_table = wdk_sys::WdfFunctions;
+                        let wdf_function_count = wdk_sys::wdf::get_wdf_function_count();
+                        if true {
```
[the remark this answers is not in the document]
leon-xd: this is caused by macro expansion.

crates/wdk-sys/Cargo.toml
wmmc88: Ah yes. the comment was originally from when the approach was for racy once_cell. resolving this as its no longer relevant.

tests/wdk-macros-tests/tests/outputs/nightly/macrotest/bug_unused_imports.expanded.rs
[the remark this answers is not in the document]
wmmc88: ah interesting. i didn't know debug_assert used a runtime check instead of condition compilation

crates/wdk-sys/src/lib.rs
```diff
@@ -5,8 +5,6 @@
 
 #![no_std]
 
-#[cfg(any(driver_model__driver_type = "KMDF", driver_model__driver_type = "UMDF"))]
```
wmmc88, line 8: nit: we _try_ to follow [conventional commits spec](https://www.conventionalcommits.org/en/v1.0.0/). Since you're removing a public item exported from the wdk-sys crate, this is considered a breaking change and the commit should indicate it via `!`. (PR title ends up being the message of the squashed commit)

crates/wdk-sys/build.rs
```diff
@@ -427,7 +397,7 @@ fn main() -> anyhow::Result<()> {
                         .expect("Scoped Thread should spawn successfully"),
                 );
 
-                info_span!("wdf_function_table.rs generation").in_scope(|| {
+                info_span!("wdf_function_count.rs generation").in_scope(|| {
                     generate_wdf_function_table(&out_path, &config)?;
```
wmmc88:
```suggestion
                    generate_wdf_function_count(&out_path, &config)?;
```

crates/wdk-sys/build.rs
```diff
@@ -259,15 +240,14 @@ fn generate_wdf(out_path: &Path, config: &Config) -> Result<(), ConfigError> {
     }
 }
 
-/// Generates a `wdf_function_table.rs` file in `OUT_DIR` which contains the
-/// definition of `WDF_FUNCTION_TABLE`. This is required to be generated here
+/// Generates a `wdf_function_count.rs` file in `OUT_DIR` which contains the
```
wmmc88, line 324: Seems like this comment is no longer accurate since theres no `WDF_FUNCTION_COUNT` generated

crates/wdk-sys/build.rs
```diff
@@ -31,51 +31,30 @@ use wdk_build::{
...
+
+const WDF_FUNCTION_COUNT_DECLARATION_EXTERNAL_SYMBOL: &str =
+    "(unsafe { crate::WdfFunctionCount }) as usize";
+
+const WDF_FUNCTION_COUNT_DECLARATION_TABLE_INDEX: &str =
+    "crate::_WDFFUNCENUM::WdfFunctionTableNumEntries as usize";
+
+static WDF_FUNCTION_COUNT_FUNCTION_TEMPLATE: LazyLock<String> = LazyLock::new(|| {
+    format!(
+        r"/// function to access the value of the number of functions in the WDF function table.
```
wmmc88:
```suggestion
        r"/// Returns the number of functions available in the WDF function table.
```

nit: style so that the doccomment matches typical rust style (and other rustdoc comments in this repo)

crates/wdk-sys/build.rs
```diff
@@ -31,51 +31,30 @@ use wdk_build::{
...
-        unsafe {{
-            debug_assert!(isize::try_from(wdf_function_count * core::mem::size_of::<crate::WDFFUNC>()).is_ok());
-            core::slice::from_raw_parts(wdf_function_table, wdf_function_count)
-        }}
-    }};
-}}"#
-    );
-    static ref CALL_UNSAFE_WDF_BINDING_TEMPLATE: String = format!(
+const WDF_FUNCTION_COUNT_PLACEHOLDER: &str =
+    "<PLACEHOLDER FOR EVALUATION CONTAINING WDF_FUNCTION_COUNT VALUE";
```
wmmc88:
```suggestion
    "<PLACEHOLDER FOR EXPRESSION FOR NUMBER OF WDF FUNCTIONS IN `wdk_sys::WdfFunctions`";
```

crates/wdk-sys/build.rs
```diff
@@ -31,51 +31,30 @@ use wdk_build::{
...
+
+const WDF_FUNCTION_COUNT_DECLARATION_EXTERNAL_SYMBOL: &str =
```
wmmc88, line 44: re-add the removed safety comment

wmmc88 (changes requested):
I really dislike adding a public api for the function count, since no such public api exists in the WDK, and i'd like for the sys crate to be purely the wdk apis. Ideally, the proc_macro should generate code that either dereferences the symbol OR uses the enum by conditionally compiling for the relevant cases. This essentially means turning the `is_wdf_function_count_generated` expression into a CFG.

This is currently not possible since:
1. we do not expose WDF version number in `EXPORTED_CFG_SETTINGS` in `emit_cfg_settings` 
2. it would be very tedious and manual to cfg over all the supported values, since there is no rust cfg equivalent to `UMDF_VERSION_MINOR > 1234`. We would need to figure out a way to write a cfg in this fashion (is this even possible?)

I think those things are out of the scope of this PR, and the best we can do right now is hide the `get_wdf_function_count` api from the user and make it clear that it is subject to future removal once we've figured out the above.

wmmc88:
> I really dislike adding a public api for the function count, since no such public api exists in the WDK, and i'd like for the sys crate to be purely the wdk apis. Ideally, the proc_macro should generate code that either dereferences the symbol OR uses the enum by conditionally compiling for the relevant cases. This essentially means turning the `is_wdf_function_count_generated` expression into a CFG.
> 
> This is currently not possible since:
> 
> 1. we do not expose WDF version number in `EXPORTED_CFG_SETTINGS` in `emit_cfg_settings`
> 2. it would be very tedious and manual to cfg over all the supported values, since there is no rust cfg equivalent to `UMDF_VERSION_MINOR > 1234`. We would need to figure out a way to write a cfg in this fashion (is this even possible?)
> 
> I think those things are out of the scope of this PR, and the best we can do right now is hide the `get_wdf_function_count` api from the user and make it clear that it is subject to future removal once we've figured out the above.

For more on #2, see https://users.rust-lang.org/t/cfg-w-arbitrary-expressions/109371

I think it may be possible to write a new attribute macro that expands into a cfg

crates/wdk-sys/src/wdf.rs
```diff
@@ -20,4 +20,4 @@ mod bindings {
     include!(concat!(env!("OUT_DIR"), "/wdf.rs"));
 }
 
-include!(concat!(env!("OUT_DIR"), "/wdf_function_table.rs"));
+include!(concat!(env!("OUT_DIR"), "/wdf_function_count.rs"));
```
wmmc88: Put this inside a module named `__private`, and have the module be `#[doc(hidden)]`. I don't like our sys crate having a public api that isn't present in the WDK

wmmc88: to be clear, this doesn't actually make it private. the modules still needs to be `pub` in order for the proc_macro to access it. This just hides it from the generated docs (and rust-analyzer auto-complete) and makes it clear that its an internal implementation detail.

Similar approaches are used in crates like serde to expose internal-only apis to their proc_macros: https://github.com/serde-rs/serde/blob/930401b0dd58a809fce34da091b8aa3d6083cb33/serde/src/lib.rs#L324

We also use a similar approach in our proc_macro

crates/wdk-sys/src/wdf.rs
```diff
@@ -20,4 +20,10 @@ mod bindings {
     include!(concat!(env!("OUT_DIR"), "/wdf.rs"));
 }
 
-include!(concat!(env!("OUT_DIR"), "/wdf_function_table.rs"));
+// This is a temporary workaround to expose the generated function count. When
+// we are able to parse the configuration at runtime, we will be able to remove
+// the function count generation from `build.rs`.
+#[allow(missing_docs)]
+pub mod __private {
```
wmmc88, line 30:
```suggestion
#[doc(hidden)]
pub mod __private {
```

crates/wdk-sys/src/wdf.rs
```diff
@@ -20,4 +20,10 @@ mod bindings {
     include!(concat!(env!("OUT_DIR"), "/wdf.rs"));
 }
 
-include!(concat!(env!("OUT_DIR"), "/wdf_function_table.rs"));
+// This is a temporary workaround to expose the generated function count. When
+// we are able to parse the configuration at runtime, we will be able to remove
+// the function count generation from `build.rs`.
```
wmmc88:
```suggestion
// This is a workaround to expose the generated function count to the `call_unsafe_wdf_function_binding` proc-macro, so that the macro-generated code can determine the slice size at runtime. When
// we are able to conditionally compile based off a cfg range for WDF version, this module can be removed and the runtime check can be replaced with a conditional compilation: https://github.com/microsoft/windows-drivers-rs/issues/276
```

wmmc88 (changes requested):

wmmc88 (dismissed):

hamzamust (dismissed):

wmmc88 (approved):

hamzamust (dismissed):

wmmc88 (dismissed):

wmmc88 (dismissed):

wmmc88 (approved):

hamzamust (approved):

PR #259, opened 2025-01-09
refactor: port certificate-generation condition script to Rust

### LINKED ISSUE #93: Port `duckscript` `condition_script` blocks to Rust
With https://github.com/sagiegurari/cargo-make/issues/1025 being resolved in cargo make v0.37.8, these condition scripts could be refactored to be rust-based for better maintainability, but since its fully functional right now, I'll leave that as a future improvement.

_Originally posted by @wmmc88 in https://github.com/microsoft/windows-drivers-rs/pull/35#discussion_r1462521681_

Also see https://github.com/microsoft/windows-drivers-rs/pull/35#discussion_r1462521821

leon-xd:
Ported last Duckscript condition script to Rust

closes #93

### DIFF
--- a/crates/wdk-build/src/cargo_make.rs
+++ b/crates/wdk-build/src/cargo_make.rs
@@ -1015,15 +1015,18 @@ pub fn package_driver_flow_condition_script() -> anyhow::Result<()> {
 ///
 /// This functions returns an error whenever it determines that the
 /// `generate-certificate` `cargo-make` task should be skipped. This only
-/// occurs when the `WdrLocalTestCert` cannot be added from the build directory
-/// to `WDRTestCertStore`.
+/// occurs when `WdrLocalTestCert` already exists in `WDRTestCertStore`.
 ///
 /// # Panics
 ///
-/// Panics if certmgr is unable to be run.
+/// Panics if `CARGO_MAKE_CURRENT_TASK_NAME` is not set in the environment.
 pub fn generate_certificate_condition_script() -> anyhow::Result<()> {
     condition_script(|| {
-        let output = Command::new("certmgr")
+        let cert_save_location = get_wdk_build_output_directory().join("WDRLocalTestCert.cer");
+
+        let mut command = Command::new("certmgr");
+
+        command
             .args([
                 "-put",
                 "-s",
@@ -1032,13 +1035,23 @@ pub fn generate_certificate_condition_script() -> anyhow::Result<()> {
                 "-n",
                 "WdrLocalTestCert",
             ])
-            .arg(get_wdk_build_output_directory().join("WDRLocalTestCert.cer"))
-            .output()
-            .expect("Failed to run certmgr.exe.");
+            .arg(cert_save_location);
+
+        let output = command.output().unwrap_or_else(|err| {
+            panic!(
+                "Failed to run certmgr.exe {} due to error: {}",
+                command
+                    .get_args()
+                    .map(|arg| arg.to_string_lossy())
+                    .collect::<Vec<_>>()
+                    .join(" "),
+                err
+            )
+        });
 
         match output.status.code() {
             Some(0) => Err(anyhow::anyhow!(
-                "WDRLocalTestCert already in WDRTestCertStore. Skipping certificate generation."
+                "WDRLocalTestCert found in WDRTestCertStore. Skipping certificate generation."
             )),
             Some(1) => Ok(()),
             Some(_) => {
@@ -1046,8 +1059,7 @@ pub fn generate_certificate_condition_script() -> anyhow::Result<()> {
                 Ok(())
             }
             None => {
-                eprintln!("No status code found from certmgr. Generating certificate.");
-                Ok(())
+                unreachable!("Unreachable, no status code found from certmgr.");
             }
         }
     })
--- a/crates/wdk-build/src/cargo_make.rs
+++ b/crates/wdk-build/src/cargo_make.rs
@@ -16,6 +16,7 @@ use std::{
     env,
     panic::UnwindSafe,
     path::{Path, PathBuf},
+    process::Command,
 };
 
 use anyhow::Context;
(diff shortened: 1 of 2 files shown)


crates/wdk-build/src/cargo_make.rs
```diff
@@ -1007,6 +1008,51 @@ pub fn package_driver_flow_condition_script() -> anyhow::Result<()> {
...
 
+/// `cargo-make` condition script for `generate-certificate` task in
+/// [`rust-driver-makefile.toml`](../rust-driver-makefile.toml)
+///
+/// # Errors
+///
+/// This functions returns an error whenever it determines that the
+/// `generate-certificate` `cargo-make` task should be skipped. This only
+/// occurs when the `WdrLocalTestCert` cannot be added from the build directory
+/// to `WDRTestCertStore`.
```
wmmc88:
```suggestion
/// occurs when `WdrLocalTestCert` already exists in `WDRTestCertStore`.
```

Docs for `certmgr -put`:
> Saves an X.509 certificate, CTL, or CRL from a certificate store to a file.

`certmgr.exe -put -s WDRTestCertStore -c -n WDRLocalTestCert ${WDK_BUILD_OUTPUT_DIRECTORY}/WDRLocalTestCert.cer` tries to save a cert called `WDRLocalTestCert` from the `WDRTestCertStore` store to `${WDK_BUILD_OUTPUT_DIRECTORY}/WDRLocalTestCert.cer`. This fails (ie. returns non-zero) if the cert doesn't exist.

crates/wdk-build/src/cargo_make.rs
```diff
@@ -1007,6 +1008,51 @@ pub fn package_driver_flow_condition_script() -> anyhow::Result<()> {
...
+///
+/// # Panics
+///
+/// Panics if certmgr is unable to be run.
+pub fn generate_certificate_condition_script() -> anyhow::Result<()> {
```
wmmc88:
```suggestion
/// Panics if `CARGO_MAKE_CURRENT_TASK_NAME` is not set in the environment
pub fn generate_certificate_condition_script() -> anyhow::Result<()> {
```

This function only panics if `condition_script` itself panics, since `condition_script` catches all other panics. `condition_script` only panics when `CARGO_MAKE_CURRENT_TASK_NAME` is not set. 

I've made this mistake elsewhere in the this file too. All functions that are just `condition_script` should only panics when `CARGO_MAKE_CURRENT_TASK_NAME` is not set. Could you fix this doc error for the rest of the comments in this file?

crates/wdk-build/src/cargo_make.rs
```diff
@@ -1007,6 +1008,51 @@ pub fn package_driver_flow_condition_script() -> anyhow::Result<()> {
...
+                "-put",
+                "-s",
+                "WDRTestCertStore",
+                "-c",
+                "-n",
+                "WdrLocalTestCert",
+            ])
+            .arg(get_wdk_build_output_directory().join("WDRLocalTestCert.cer"))
+            .output()
+            .expect("Failed to run certmgr.exe.");
```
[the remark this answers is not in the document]
wmmc88: @leon-xd I think the expect prints the Err variant contents, but doesn't print anything to do with the Command. It probably makes sense to print all the args here to make it more clear what command failed, since it will fail and then continue to try to run the task. Maybe do smth like:

```rust
let cert_save_location = get_wdk_build_output_directory().join("WDRLocalTestCert.cer");
let args = [
                "-put",
                "-s",
                "WDRTestCertStore",
                "-c",
                "-n",
                "WdrLocalTestCert",
                cert_save_location.as_str()
            ];
let output = Command::new("certmgr")
            .args(&args)
            .output()
            .unwrap_or_else(|| panic!("Failed to run 'certmgr.exe {}'", &args.join(" ")));

```

crates/wdk-build/src/cargo_make.rs
```diff
@@ -1007,6 +1008,51 @@ pub fn package_driver_flow_condition_script() -> anyhow::Result<()> {
...
+
+        match output.status.code() {
+            Some(0) => Err(anyhow::anyhow!(
+                "WDRLocalTestCert already in WDRTestCertStore. Skipping certificate generation."
```
wmmc88:
```suggestion
                "WDRLocalTestCert found in WDRTestCertStore. Skipping certificate generation."
```
let's keep the existing error message

crates/wdk-build/src/cargo_make.rs
```diff
@@ -1007,6 +1008,51 @@ pub fn package_driver_flow_condition_script() -> anyhow::Result<()> {
...
+            )),
+            Some(1) => Ok(()),
+            Some(_) => {
+                eprintln!("Unknown status code found from certmgr. Generating certificate.");
+                Ok(())
+            }
+            None => {
+                eprintln!("No status code found from certmgr. Generating certificate.");
+                Ok(())
+            }
```
wmmc88: I believe this None case is only applicable for Unix since Windows doesn't use signals. You can replace this branch with an `unreachable!`

crates/wdk-build/src/cargo_make.rs
```diff
@@ -1007,6 +1008,51 @@ pub fn package_driver_flow_condition_script() -> anyhow::Result<()> {
...
+            ])
+            .arg(get_wdk_build_output_directory().join("WDRLocalTestCert.cer"))
+            .output()
+            .expect("Failed to run certmgr.exe.");
+
+        match output.status.code() {
+            Some(0) => Err(anyhow::anyhow!(
+                "WDRLocalTestCert already in WDRTestCertStore. Skipping certificate generation."
+            )),
+            Some(1) => Ok(()),
```
wmmc88: The original script had a `WDRLocalTestCert not found in WDRTestCertStore. Generating new certificate.` message. this branch should retain it

wmmc88 (changes requested):
some minor comments, but good to merge after

crates/wdk-build/src/cargo_make.rs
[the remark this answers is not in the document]
leon-xd: makes sense -- I kept this as two calls to args/arg since i thought it was a bit messy converting from an OsString to a str but I suppose displaying these arguments is worth it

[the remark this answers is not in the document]
leon-xd: @wmmc88 I ended up doing it slightly differently. cert_save_location is a `PathBuf` which is impossible to reduce to a `&str` without either an [`unwrap` after a conversion](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.to_str) or a [lossy str conversion](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.to_string_lossy). Given that it's possible for path strings to be non-UTF8 I figured keeping the lossy conversion until the unwrap was more complete, in case a developer's build path contains a non-UTF8 string.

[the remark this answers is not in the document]
wmmc88: I think its useful to look at why it was working before: https://doc.rust-lang.org/std/process/struct.Command.html#method.arg takes in a `AsRef<OsStr>`. Since slices need to have all members be of the same type, what you could do is actually just call [`as_os_str`](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.as_os_str) on the `PathBuf`, and then have all the other `str` in the slice be `OsStr` via `AsRef<OsStr> for str`

Example: https://play.rust-lang.org/?version=nightly&mode=release&edition=2024&gist=c8c49b3308e432a31ed278c2d89467c3

crates/wdk-build/src/cargo_make.rs
```diff
@@ -1007,6 +1008,68 @@ pub fn package_driver_flow_condition_script() -> anyhow::Result<()> {
...
+            )
+        });
+
+        match output.status.code() {
+            Some(0) => Err(anyhow::anyhow!(
+                "WDRLocalTestCert found in WDRTestCertStore. Skipping certificate generation."
+            )),
+            Some(1) => {
+                println!(
+                    "WDRLocalTestCert not found in WDRTestCertStore. Generating new certificate."
```
wmmc88, line 1057: Does this print show up properly? I think I used eprintln for all the other prints for some reason to do with how cargo-make normally outputs things

leon-xd: fixed in latest push

crates/wdk-build/src/cargo_make.rs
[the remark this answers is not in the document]
leon-xd: gotcha - used `as_ref` with all the `&str`s in latest push.

wmmc88 (approved):

hamzamust (approved):

PR #260, opened 2025-01-11
feat!: expand `wdk-sys` coverage to include hid-related headers

wmmc88:
This pull request introduces several significant changes to the `wdk-sys` and `wdk-build` crates. The main changes include adding headers related to Windows HID support via a new `hid` feature, and laying the groundwork for future features to be added to further expand API coverage. Bindgen input has moved from a static input header to a dynamically generated header passed to [bindgen's header_contents method](https://docs.rs/bindgen/latest/bindgen/struct.Builder.html#method.header_contents), allowing for the inputs to more easily be adaptable to different cargo features.

### HID Support:
* Introduced the `ApiSubset` enum and related methods in the `Config` struct to handle different subsets of APIs required for various driver types. [[1]](diffhunk://#diff-43524f733cccc1a175c3bf83309866acb53b8cadb1628478ca1c87d31dda2b26R193-R203) [[2]](diffhunk://#diff-43524f733cccc1a175c3bf83309866acb53b8cadb1628478ca1c87d31dda2b26L544-R560) [[3]](diffhunk://#diff-43524f733cccc1a175c3bf83309866acb53b8cadb1628478ca1c87d31dda2b26R639-R771) [[4]](diffhunk://#diff-43524f733cccc1a175c3bf83309866acb53b8cadb1628478ca1c87d31dda2b26L686-R833)
* [`crates/wdk-sys/build.rs`](diffhunk://#diff-2006bfd5c1bb8d3bf7c95ff14e2a7ddd94fa13d1ce6cbf293b90138a903d413fR237-R262): Added the `generate_hid` function to generate bindings for HID-related headers and avoid duplicate definitions.
* [`crates/wdk-sys/src/hid-input.h`](diffhunk://#diff-3b214824cdcd0ea0baa1bcd7c4f8d03fb3cc87d866ac2d2105c5257fd26416cfR1-R16): Added a new header file for HID input with the necessary includes and comments for future improvements.
* [`crates/wdk-sys/src/hid.rs`](diffhunk://#diff-3547ef0e00ae9e4370e8acbc4593c65efe2127ada2fafd3166363221e06b8fdcR1-R14): Added a new source file for HID bindings, including the necessary modules and imports.
* [`crates/wdk-sys/src/lib.rs`](diffhunk://#diff-2a6d6fddbacd5610778650cafba5da93eb702ddeb5144ee73a34d254a4300eb7R26-R40): Added a new module for HID support, ensuring it is included for all driver models.

### Kernel Mode Definitions:
* [`crates/wdk-build/src/lib.rs`](diffhunk://#diff-43524f733cccc1a175c3bf83309866acb53b8cadb1628478ca1c87d31dda2b26L539-R547): Updated the `Config` implementation to include `_KERNEL_MODE` definition for both WDM and KMDF driver configurations.

### Workflow Enhancements:
* Updated various GitHub workflow files (`.github/workflows/build.yaml`, `.github/workflows/codeql.yml`, `.github/workflows/docs.yaml`, `.github/workflows/lint.yaml`, `.github/workflows/test.yaml`) to include the `--all-features` flag in Cargo commands to ensure all features are built and tested. [[1]](diffhunk://#diff-d0777657fa3fd81d23aaf7273e58aee453b04e67882517900c56daeef9b3e4c1L76-R90) [[2]](diffhunk://#diff-12783128521e452af0cfac94b99b8d250413c516ec71fe6d97dbea666ff7ba27L92-R92) [[3]](diffhunk://#diff-d54d69dbb27e75dae25cb4b2384310cb57707e419377cf572d5cb0ecc1f16877L75-R80) [[4]](diffhunk://#diff-4b122024a3a28ded65da76a2f1bface1f3a27328374438d1298d25585fe0603bL74-R78) [[5]](diffhunk://#diff-245392b692a50c38ecab4381b118862db514035c10983f3bd4f4b7f1f4be4692L76-R76)

### Breaking Changes:

BREAKING CHANGE: `wdk_default` no longer takes in a list of c header files. Users are expected to specify `headers` or `header_contents` via their respective `bindgen::Builder` APIs

BREAKING CHANGE: `Config::get_include_paths` has been renamed to `Config::include_paths`

BREAKING CHANGE: `Config::get_library_paths` has been renamed to `Config::library_paths`

BREAKING CHANGE: `Config::get_preprocessor_definitions_iter` has been renamed to `Config::preprocessor_definitions`


crates/wdk-build/src/lib.rs
```diff
@@ -623,6 +629,134 @@ impl Config {
...
+        }
+        .into_iter()
+        .map(std::string::ToString::to_string)
+    }
+
+    pub fn wdf_headers(&self) -> impl Iterator<Item = String> {
+        ["wdf.h"].into_iter().map(std::string::ToString::to_string)
+    }
+
+    #[cfg(feature = "hid")]
```
wmmc88: These shouldn't be gated by conditional compilation

crates/wdk-build/src/lib.rs
```diff
@@ -623,6 +629,134 @@ impl Config {
...
+  unsigned __int64 Alignment;
+} KIDTENTRY64, *PKIDTENTRY64;
+",
+            );
+        }
+
+        header_contents
+    }
+
+    pub fn bindgen_wdf_header_contents(&self) -> Option<String> {
```
wmmc88: this should probably be a common `bindgen_header_contents` function that takes in some enum value for each of the header groups

examples/sample-kmdf-driver/Cargo.toml
```diff
@@ -31,6 +31,9 @@ wdk-sys = { path = "../../crates/wdk-sys", version = "0.3.0" }
 
 [features]
 default = []
+
+hid = ["wdk-sys/hid"]
```
wmmc88, line 35: `--all-features` should probably be passed to all the CI jobs for now. Eventually, we should probably use something like `cargo-hack` for more thorough testing

.github/workflows/lint.yaml
```diff
@@ -71,11 +75,7 @@ jobs:
           targets: ${{ matrix.target_triple }}
 
       - name: Run Cargo Clippy
-        run: cargo +${{ matrix.rust_toolchain }} clippy --locked --profile ${{ matrix.cargo_profile }} --target ${{ matrix.target_triple }} --all-targets -- -D warnings
-
-      - name: Run Cargo Clippy (--features nightly)
-        if: matrix.rust_toolchain == 'nightly'
-        run: cargo +${{ matrix.rust_toolchain }} clippy --locked --profile ${{ matrix.cargo_profile }} --target ${{ matrix.target_triple }} --all-targets --features nightly -- -D warnings
+        run: cargo +${{ matrix.rust_toolchain }} clippy --locked --profile ${{ matrix.cargo_profile }} --target ${{ matrix.target_triple }} --all-targets --all-features
```
leon-xd, line 78: the `--all-features` flag should be added to the instructions in `CONTRIBUTING.md` for running `clippy`

.github/workflows/codeql.yml
```diff
@@ -89,7 +89,7 @@ jobs:
 
     - if: matrix.build-mode == 'manual'
       working-directory: ./examples
-      run: cargo +${{ matrix.rust_toolchain }} make default --locked --profile ${{ matrix.cargo_profile }} --target ${{ matrix.target_triple }} --workspace
+      run: cargo +${{ matrix.rust_toolchain }} make default --locked --profile ${{ matrix.cargo_profile }} --target ${{ matrix.target_triple }} --workspace --all-features
```
leon-xd, line 92: if `--all-features` is required to build examples we should also add this to the `README` in each sample driver

.github/workflows/lint.yaml
leon-xd: same thing applies to test & doc

crates/wdk-build/src/lib.rs
```diff
@@ -623,6 +634,139 @@ impl Config {
...
+            }
+        }
+        .into_iter()
+        .map(std::string::ToString::to_string)
+    }
+
+    /// Returns a [`String`] containing the contents of a header file designed
+    /// for [`bindgen`](https://docs.rs/bindgen) to processs
+    ///
+    /// The contents contains `#include`'ed headers based off the [`ApiSubset`]
```
leon-xd:
```suggestion
    /// The contents contain `#include`'ed headers based off the [`ApiSubset`]
```

crates/wdk-sys/build.rs
```diff
@@ -203,8 +218,20 @@ fn initialize_tracing() -> Result<(), ParseError> {
 fn generate_constants(out_path: &Path, config: &Config) -> Result<(), ConfigError> {
     info!("Generating bindings to WDK: constants.rs");
 
-    Ok(bindgen::Builder::wdk_default(vec!["src/input.h"], config)?
+    let header_contents = config.bindgen_header_contents([
+        ApiSubset::Base,
+        ApiSubset::Wdf,
```
leon-xd, line 222: if both `Base` and `Wdf` are always included in builds, is there any advantage to separating these, other than keeping things modular? Is there any confusion that could arise -- I don't think this code should be tinkered with by driver developers so I doubt it, but want to bring it up just in case.

crates/wdk-sys/build.rs
```diff
@@ -213,40 +240,66 @@ fn generate_constants(out_path: &Path, config: &Config) -> Result<(), ConfigErro
...
         .expect("Bindings should succeed to generate")
         .write_to_file(out_path.join("types.rs"))?)
 }
 
 fn generate_base(out_path: &Path, config: &Config) -> Result<(), ConfigError> {
     let outfile_name = match &config.driver_config {
-        DriverConfig::Wdm | DriverConfig::Kmdf(_) => "ntddk.rs",
-        DriverConfig::Umdf(_) => "windows.rs",
+        DriverConfig::Wdm | DriverConfig::Kmdf(_) => "ntddk",
+        DriverConfig::Umdf(_) => "windows",
```
leon-xd, line 264: looked into this a bit but couldn't find anything super conclusive. but wondering if there's a better way using `#[repr()]` and a custom struct to contain all of the specific strings/bits of config across the two enums that we have (`ApiSubset`/`DriverConfig`). I know that there are dynamic strings that we are able to set with Kmdf/Umdf based on configuration, not sure if it's idiomatic to have static stuff involved here. But also this might be too OOP for this setting, I see the advantage of having these match statements sprinkled around.

crates/wdk-sys/build.rs
```diff
@@ -259,6 +312,43 @@ fn generate_wdf(out_path: &Path, config: &Config) -> Result<(), ConfigError> {
...
+            let bindgen_builder = {
+                 let mut builder = bindgen::Builder::wdk_default(config)?
+                .with_codegen_config((CodegenConfig::TYPES | CodegenConfig::VARS).complement())
+                .header_contents("hid-input.h", &header_contents);
+
+            // Only allowlist files in the hid-specific files to avoid duplicate definitions
+            for header_file in config.headers(ApiSubset::Hid)
+            {
+                builder = builder.allowlist_file(format!("(?i).*{header_file}.*"));
+            }
```
leon-xd: Not sure if `fmt` or `clippy` caught this but i'm prety sure 325-332 should be indented

leon-xd: also why is `bindgen_builder` created from `builder` in a separate block here? not quite sure I understand why this is separate -- is it to keep `bindgen_builder` from being `mut`?

leon-xd (changes requested):
main change is to add some relevant changes to `CONTRIBUTING.md`, everything else is some general questions. great work!

crates/wdk-sys/build.rs
wmmc88: > Not sure if `fmt` or `clippy` caught this but i'm prety sure 325-332 should be indented

Nice catch. `rustfmt` can't really format code within macros, but does special case some things in certain macros (like `cfg-if`). [This cfg-if issue](https://github.com/rust-lang/cfg-if/issues/35) tracks that problem, but probably won't ever get fixed since people rather focus on stabilizing [`cfg_match`](https://github.com/rust-lang/rust/issues/115585) (which we should pivot too once its ready)

> also why is `bindgen_builder` created from `builder` in a separate block here? not quite sure I understand why this is separate -- is it to keep `bindgen_builder` from being `mut`?

You're right. Its a stylistic choice to limit the scope of `mut`. `mut` is required here in order to be able to do the loop to allowlist every header (since allowlist uses `self` and returns a new `Builder`). I guess since its regex, i could programatically generate some regex that allowlists all the files in one go, so that I can just call `allowlist_file` once, but i figure that regex is always more confusing to read and get right 😆

.github/workflows/codeql.yml
wmmc88: All features is not required to build samples. `all-features` just enables binding generation for the optional features (like hid). 

Technically nothing in the examples does anything with `hid` at all. Enabling `wdk-sys/hid` via `all-features` is just a sanity check to make sure hid features compile correctly in all configuration, until #269 is implemented

crates/wdk-sys/build.rs
wmmc88: I originally wanted to have some sort of mapping thing to unify something from `Config` to <some list of configurations and stuff>, but couldn't find a way to make it cleaner and more readable than what i wrote. Its sort of tricky because some of the binding generation comes from cargo features, and other things come from WDK config.

Also the headers to process via bindgen are not always necessarily the same. Example: the types and constants want all enabled features, whereas the api generate functions will need only the relevant subsets

wmmc88: I think I'll keep it as is for now, and see how adding more tech-specific api subsets evolves this file. If there is some pattern later, we can refactor

crates/wdk-sys/build.rs
wmmc88: It's now mostly to break down the subsets to align with mslearn documentation for organizational benefit. In an earlier revision of this PR, the config was checked for every `generate_*` function before adding `wdf.h` but that logic was moved to the `Config::headers`

crates/wdk-build/src/lib.rs
wmmc88: fixed

.github/workflows/lint.yaml
wmmc88: fixed

.github/workflows/codeql.yml
leon-xd: makes sense!

crates/wdk-sys/build.rs
leon-xd: gotcha, understood.

crates/wdk-sys/build.rs
leon-xd: makes sense that this is unclear as of now, we can keep it in the back of our minds!

leon-xd (dismissed):

crates/wdk-build/src/lib.rs
```diff
@@ -623,6 +634,139 @@ impl Config {
...
+        .into_iter()
+        .map(std::string::ToString::to_string)
+    }
+
+    /// Returns a [`String`] containing the contents of a header file designed
+    /// for [`bindgen`](https://docs.rs/bindgen) to processs
+    ///
+    /// The contents contain `#include`'ed headers based off the [`ApiSubset`]
+    /// and [`Config`], as well as any additional definitions required for the
+    /// headers to be processed sucessfully
```
ankurung: successfully.

ankurung (dismissed):

crates/wdk-build/src/lib.rs
wmmc88: fixed!

ankurung (approved):

leon-xd (approved):

PR #280, opened 2025-01-29
fix: generated types from `DECLARE_HANDLE` macro are marked as opaque, non-copy, and non-default

leon-xd:
This pull request takes care of a `bindgen` oversight.


wmmc88:
The opaque handles from[ this doc page](https://learn.microsoft.com/en-us/windows-hardware/drivers/kernel/eprocess) should also be handled (if not already handled)

PR #295, opened 2025-02-14
feat: Cache parameters & return type during `call_unsafe_wdf_function_binding` macro expansion

leon-xd:
### Overview
Currently, each invocation of the `call_unsafe_wdf_function_binding` macro loads the entire `types.rs` file into memory, and parses it for relevant information. This is very inefficient because `types.rs` is generated to be over 200,000 lines long, and we are searching the entire file for just a few lines of information.

This PR instead introduces a caching mechanism that stores the relevant information from `types.rs` in storage, so that we don't have to re-parse the entire file every time we need to look up a type or return value. This is done in `generate_derived_ast_fragments` by generating a temporary `scratch` directory, dumping the relevant information into a file in that directory, and then loading that file into memory when we need to look up the parameters & return value. This generated file is only 32 KB, as opposed to `types.rs`, which is 5 MB.

### Results
Using the `crox` tool we can evaluate the difference in performance of this change. The previous version of this implementation took around 3 seconds per invocation, as seen here:
image

After this change, the first invocation of this macro increases to a bit under 5 seconds:
image

But after each file is cached, each subsequent invocation of the macro takes around 3 milliseconds.
image

The time saved per build scales with the number of WDF function invocations, meaning this greatly improves the developer experience.


crates/wdk-macros/src/lib.rs
```diff
@@ -318,6 +459,10 @@ impl IntermediateOutputASTFragments {
     }
 }
 
+fn map_to_syn_error<E: std::fmt::Display>(span: Span, error: E) -> Error {
```
wmmc88: `map_to_syn_error` is more verbose than just `Error::new` everywhere

crates/wdk-macros/tests/unit-tests-input/generated-types.rs
```diff
@@ -3,6 +3,12 @@
 
 //! Snippet of a bindgen-generated file containing types information used by tests for [`wdk_macros::call_unsafe_wdf_function_binding!`]
 
+pub mod _WDFFUNCENUM {
+    pub const WdfDriverCreateTableIndex: Type = 116;
+    pub const WdfVerifierDbgBreakPointTableIndex: Type = 367;
+}
+
+
```
wmmc88:
```suggestion
```

nit

crates/wdk-macros/src/lib.rs
```diff
@@ -95,6 +97,14 @@ struct IntermediateOutputASTFragments {
     inline_wdf_fn_invocation: ExprCall,
 }
 
+/// Struct storing string representations of the information we want to cache
+/// from `types.rs`.
+#[derive(Serialize, Deserialize)]
+struct SavedFunctionInfo {
```
wmmc88:
```suggestion
struct CachedFunctionInfo {
```

nit: suggestion

crates/wdk-macros/src/lib.rs
```diff
@@ -202,6 +230,119 @@ impl Inputs {
             inline_wdf_fn_name,
         })
     }
+
+    // Motivation for this function is to reduce build time. Rather than parse
+    // types.rs for relevant information to construct function table call on each
+    // macro invocation, we store all possible function table information for each
+    // function on first invocation. We cache this using the `scratch` crate, and
+    // read from this cache for each subsequent invocation.
+    fn get_fragment_info_map(&self) -> Result<HashMap<String, SavedFunctionInfo>> {
```
wmmc88:
```suggestion
    /// This function aims to reduce build time by caching function table information.
    /// Instead of parsing `types.rs` for relevant data on every macro invocation,
    /// this function stores all possible function table information during the first invocation
    /// using the `scratch` crate. Subsequent invocations can read from this cache.
    fn get_fragment_info_map(&self) -> Result<HashMap<String, SavedFunctionInfo>> {
    fn get_fragment_info_map(&self) -> Result<HashMap<String, SavedFunctionInfo>> {
```

Copilot suggested reword with justification:

> The improved comment is better because it provides a clear and concise explanation of the function's purpose, which is to reduce build times by caching function table information during the first invocation and reading from the cache in subsequent invocations. This added clarity helps developers understand the motivation and mechanism behind the function, making the code more maintainable and easier to comprehend.

crates/wdk-macros/src/lib.rs
```diff
@@ -202,6 +230,119 @@ impl Inputs {
...
+            serde_json::from_str(&generated_map_string)
+                .map_err(|e| map_to_syn_error(self.wdf_function_identifier.span(), e))?;
+        Ok(map)
+    }
+
+    // To generate the cache we parse the types.rs file for `mod _WDFFUNCENUM`. This
+    // stores each function table name with the suffix "TableIndex", which we parse
+    // and trim. We then use this function name to parse for the function pointer
+    // type alias, which we then parse for the function parameters and return type.
+    fn generate_fragment_info_map_from_types(&self) -> Result<HashMap<String, SavedFunctionInfo>> {
```
wmmc88: This function should also have tests

crates/wdk-macros/src/lib.rs
wmmc88: If you want to reduce code dedup here, you probably want to instead have an extension trait you impl for the main Result types you're mapping, and add a method for `map_to_syn_err(&self, span)`. then these you can replace all the `map_err` with `map_to_syn_err`

crates/wdk-macros/src/lib.rs
```diff
@@ -202,6 +230,119 @@ impl Inputs {
...
+        let types_ast = parse_types_ast(&self.types_path)?;
+
+        let func_enum_mod: Option<&syn::ItemMod> = types_ast.items.iter().find_map(|item| {
+            if let Item::Mod(mod_alias) = item {
+                if mod_alias.ident == "_WDFFUNCENUM" {
```
wmmc88: maybe const for this somewhere since u use it multiple times

crates/wdk-macros/src/lib.rs
```diff
@@ -202,6 +230,119 @@ impl Inputs {
...
+
+        let generated_map_string = std::fs::read_to_string(&fragment_info_map_path)
+            .map_err(|e| map_to_syn_error(self.wdf_function_identifier.span(), e))?;
+        let map: HashMap<String, SavedFunctionInfo> =
+            serde_json::from_str(&generated_map_string)
+                .map_err(|e| map_to_syn_error(self.wdf_function_identifier.span(), e))?;
+        Ok(map)
+    }
+
+    // To generate the cache we parse the types.rs file for `mod _WDFFUNCENUM`. This
```
wmmc88: Even though this is a private function, you should write its comment like a doc comment (you can provide an option to cargo doc to enable generation for these functions. Also IIRC it shows in ide autocomplete).

Also generally for doc comments, you want to write in a certain style so its easier for others to read:

Writing good Rustdoc comments can really make your code stand out and be easily understood by others (or even yourself, months down the line). Here are some tips for achieving clarity, precision, and usefulness in your Rustdoc comments. See this copilot-generated summary (especially #1 & #2):


### 1. **Write in Present Tense**
   Use the present tense to keep things concise and direct. Instead of saying, “This function will calculate...,” say, “This function calculates...”.

### 2. **Use Active Voice**
   Active voice makes your comments easier to read and understand. For example, write, “This function checks if a number is prime,” rather than, “This function is used to check if a number is prime.”

### 3. **Be Clear and Concise**
   Aim for clarity and brevity. Explain what the code does, what its parameters are, and what it returns, without unnecessary details.

### 4. **Follow Rustdoc Syntax**
   Use Rustdoc’s special syntax for formatting. Use triple slashes `///` for outer comments and `//!` for module-level documentation.

### 5. **Document Parameters and Return Values**
   Use `# Arguments` and `# Returns` sections to clarify what each parameter does and what the function returns. Example:
   ```rust
   /// Adds two numbers together.
   ///
   /// # Arguments
   ///
   /// * `a` - The first number.
   /// * `b` - The second number.
   ///
   /// # Returns
   ///
   /// The sum of `a` and `b`.
   fn add(a: i32, b: i32) -> i32 {
       a + b
   }
   ```

### 6. **Include Examples**
   Examples are very helpful for illustrating how to use your function or module. Use the `# Examples` section.
   ```rust
   /// Calculates the factorial of a number.
   ///
   /// # Arguments
   ///
   /// * `n` - The number to calculate the factorial of.
   ///
   /// # Returns
   ///
   /// The factorial of `n`.
   ///
   /// # Examples
   ///
   /// ```
   /// let result = factorial(5);
   /// assert_eq!(result, 120);
   /// ```
   fn factorial(n: u32) -> u32 {
       (1..=n).product()
   }
   ```

### 7. **Highlight Panics and Errors**
   If your function can panic or return errors, make sure to document this in the `# Panics` and `# Errors` sections.
   ```rust
   /// Divides two numbers.
   ///
   /// # Panics
   ///
   /// Panics if the divisor is zero.
   ///
   /// # Examples
   ///
   /// ```
   /// let result = divide(10, 2);
   /// assert_eq!(result, 5);
   /// ```
   fn divide(a: i32, b: i32) -> i32 {
       if b == 0 {
           panic!("division by zero");
       }
       a / b
   }
   ```

### 8. **Consistency is Key**
   Keep the style and structure of your comments consistent throughout your codebase.

By keeping these tips in mind, your Rustdoc comments will not only be helpful but also professional and easy to read. Happy documenting! 📜

wmmc88:
> Low priority question: what does the benchmarking look like with
> 
> ```
> [profile.dev.build-override]
> opt-level = 3
> 
> [profile.release.build-override]
> opt-level = 3
> ```
> 
> set in cargo.toml to optimize macros used during build? This saved me significant time compiling a complex driver using lots of calls to call_unsafe_wdf_function_binding; I expect your change easily outperforms this, but I'm curious if enabling this in the cargo.toml still provides meaningful speedups with your change.

Oh that's interesting. I didn't know proc-macros always compile with no optimizations. If having `opt-level = 3` does improve on top of what this pr does (I'm guessing if a driver uses one wdf call, it will use many... so `call_unsafe_wdf_function_binding` speedup **should** dominate over slowdown of the proc-macro's compilation), we should consider:

1. Recommending everyone use this setting
2. Adding it as part of our template for `cargo-wdk`

Something else of note from the cargo docs:
> Note: When a dependency is both a normal dependency and a build dependency, Cargo will try to only build it once when --target is not specified. When using build-override, the dependency may need to be built twice, once as a normal dependency and once with the overridden build settings. This may increase initial build times.

Because of this, it may be worth using a more targeted approach:
```toml
[profile.dev.package.<PKG NAME>]
opt-level = 3
```

We can have the above snippet for all packages that:
* are known to likely only be used in build scripts (ex. bindgen)
* are known to take long time executing in our build in proc-macros or in build.rs (ex. bindgen, syn, ripgrep?, etc)

crates/wdk-macros/src/lib.rs
```diff
@@ -202,6 +230,119 @@ impl Inputs {
...
         })
     }
+
+    // Motivation for this function is to reduce build time. Rather than parse
+    // types.rs for relevant information to construct function table call on each
+    // macro invocation, we store all possible function table information for each
+    // function on first invocation. We cache this using the `scratch` crate, and
+    // read from this cache for each subsequent invocation.
+    fn get_fragment_info_map(&self) -> Result<HashMap<String, SavedFunctionInfo>> {
+        let scratch_dir = scratch::path("ast_fragments");
```
wmmc88: the scratch path `suffix` arg should be enough to uniquely identify the scrath folder since any crate can access it. I would suggest prepending `env!(CARGO_CRATE_NAME)` so that there's less chance of a possible collision

crates/wdk-macros/src/lib.rs
wmmc88: Some new tests should be added for this function (see structure of existing test collateral)

wmmc88: Note the active voice and present tense

wmmc88: also this doc comment should probably have a short summary at the top that also mentions that this function fetches the map from file, and generates the file cache if it doesnt already exist

wmmc88: nit: naming bikeshed :)

where does the `fragment_info_map` name come from?

crates/wdk-macros/src/lib.rs
```diff
@@ -202,6 +230,119 @@ impl Inputs {
...
+        let flock = std::fs::File::create(scratch_dir.join(".lock"))
+            .map_err(|e| map_to_syn_error(self.wdf_function_identifier.span(), e))?;
+
+        let fragment_info_map_path = scratch_dir.join("fragment_info_map.json");
+
+        if !fragment_info_map_path.exists() {
+            FileExt::lock_exclusive(&flock)
+                .map_err(|e| map_to_syn_error(self.wdf_function_identifier.span(), e))?;
+
+            if !fragment_info_map_path.exists() {
```
wmmc88: although this double lock checking pattern is very common, it might be good to have a comment here saying the 2nd check is to make sure it wasnt created between the first check and lock acquisition

crates/wdk-macros/src/lib.rs
```diff
@@ -202,6 +230,119 @@ impl Inputs {
...
+                let generated_map = self.generate_fragment_info_map_from_types()?;
```
wmmc88:
```suggestion
                // create cache since it doesn't already exist
                let generated_map = self.generate_fragment_info_map_from_types()?;
```

wmmc88: or better yet make the code self describing by putting all the code in this block in a function called `generate_wdf_function_info_file_cache` or similar

crates/wdk-macros/src/lib.rs
```diff
@@ -202,6 +230,119 @@ impl Inputs {
...
+                let generated_map_string = serde_json::to_string(&generated_map)
+                    .map_err(|e| map_to_syn_error(self.wdf_function_identifier.span(), e))?;
+                std::fs::write(&fragment_info_map_path, generated_map_string)
+                    .map_err(|e| map_to_syn_error(self.wdf_function_identifier.span(), e))?;
+            }
+            FileExt::unlock(&flock)
+                .map_err(|e| map_to_syn_error(self.wdf_function_identifier.span(), e))?;
```
wmmc88: after this file unlock, you should early return with `generated_map` to avoid needing to read it back from the file again needlessly (its already in memory and in scope)

crates/wdk-macros/src/lib.rs
```diff
@@ -202,6 +230,119 @@ impl Inputs {
...
+                    return Some(mod_alias);
+                }
+            }
+            None
+        });
+
+        let func_enum_mod = func_enum_mod.ok_or_else(|| {
+            Error::new(
+                self.wdf_function_identifier.span(),
+                "Failed to find _WDFFUNCENUM module in types file",
```
wmmc88:
```suggestion
                "Failed to find _WDFFUNCENUM module in types.rs file",
```

crates/wdk-macros/src/lib.rs
```diff
@@ -202,6 +230,119 @@ impl Inputs {
...
+            )
+        })?;
+
+        let func_enum_mod_contents = &func_enum_mod
+            .content
+            .as_ref()
+            .ok_or_else(|| {
+                Error::new(
+                    self.wdf_function_identifier.span(),
+                    "Failed to find _WDFFUNCENUM module contents in types file",
```
wmmc88:
```suggestion
                    "Failed to find _WDFFUNCENUM module contents in types.rs file",
```

crates/wdk-macros/src/lib.rs
```diff
@@ -202,6 +230,119 @@ impl Inputs {
...
+                )
+            })?
+            .1;
```
wmmc88:
```suggestion
        let (_brace, func_enum_mod_contents) = &func_enum_mod
            .content
            .as_ref()
            .ok_or_else(|| {
                Error::new(
                    self.wdf_function_identifier.span(),
                    "Failed to find _WDFFUNCENUM module contents in types file",
                )
            })?;
```

this tuple destructuring makes it clearer what you're doing

crates/wdk-macros/src/lib.rs
```diff
@@ -202,6 +230,119 @@ impl Inputs {
...
+
+        let mut const_func_enum_types: Vec<String> = vec![];
+        for func_enum_mod_item in func_enum_mod_contents {
+            if let Item::Const(const_alias) = func_enum_mod_item {
+                const_func_enum_types.push(const_alias.ident.to_string());
+            }
+        }
```
wmmc88: do you need to create this intermediate `const_func_enum_types` vec at all? why not just iterate over this to begin with?

wmmc88: (you can use `.filter_map` and similar functions)

crates/wdk-macros/src/lib.rs
```diff
@@ -202,6 +230,119 @@ impl Inputs {
...
+
+        let mut function_name_to_info_map: HashMap<String, SavedFunctionInfo> = HashMap::new();
+        for const_func_enum_type in const_func_enum_types {
+            let Some(wdf_function_name) = const_func_enum_type.strip_suffix("TableIndex") else {
+                continue;
```
wmmc88: rather than iterate thru all and the skip the `WdfFunctionTableNumEntries` here, you can just `.filter` the iterator using [`ends_with`](https://doc.rust-lang.org/std/string/struct.String.html#method.ends_with)

crates/wdk-macros/src/lib.rs
wmmc88: i modelled all the other functions(except the main top-level ones) as free functions taking in parameters, instead of methods, so that the discrete chunks can be more easily unit tested with needing a stateful object. This function and `get_fragment_info_map` should follow that

crates/wdk-macros/src/lib.rs
```diff
@@ -202,6 +230,119 @@ impl Inputs {
...
+                "PFN_{uppercase_c_function_name}",
+                uppercase_c_function_name = wdf_function_name.to_uppercase(),
+                span = self.wdf_function_identifier.span()
+            );
+
+            let (parameters, return_type) =
+                match generate_parameters_and_return_type(&types_ast, &function_pointer_type) {
+                    Ok((parameters, return_type)) => (parameters, return_type),
+                    Err(err) => {
+                        if err
```
wmmc88: Parsing the error like this is very brittle. Since we own the offending code, we should just make `generate_parameters_and_return_type` return a `Result<Option<(Punctuated<BareFnArg, Token![,]>, ReturnType)>>`

crates/wdk-macros/src/lib.rs
```diff
@@ -202,6 +230,119 @@ impl Inputs {
...
+            .1;
+
+        let mut const_func_enum_types: Vec<String> = vec![];
+        for func_enum_mod_item in func_enum_mod_contents {
+            if let Item::Const(const_alias) = func_enum_mod_item {
+                const_func_enum_types.push(const_alias.ident.to_string());
+            }
+        }
+
+        let mut function_name_to_info_map: HashMap<String, SavedFunctionInfo> = HashMap::new();
```
wmmc88: For small N, BTreeMap performs significantly better than HashMap with default hasher. OTTOMH, most benchmarks i saw said for N<1000, use BTreeMap. This is also why `serde_json` internally uses a BTreeMap, and [we also use a btreemap for our small maps](https://github.com/microsoft/windows-drivers-rs/blob/1dd243ef07e470ec2f6a4973754bb30fe930f69d/crates/wdk-build/src/lib.rs#L312)

wmmc88: BTreeMap should also be smaller when serialized since it doesnt need to store hashes

crates/wdk-macros/src/lib.rs
```diff
@@ -202,6 +230,119 @@ impl Inputs {
...
+                            .to_string()
+                            .contains("Failed to find type alias definition for")
+                        {
+                            continue;
+                        }
+                        return Err(err);
+                    }
+                };
+
+            function_name_to_info_map.insert(
```
wmmc88: rather than insert, the more idiomatic way to write this logic is use rust's functional iterator chaining. some combo of `filter`, `map`, `filter_map` should get you to an iterator that yields only the FunctionInfo you want. Then you can use `collect::<Map>` to collect the iterator into a map. I think all the std Map types also [specialize the collect from an iterator](https://doc.rust-lang.org/std/collections/btree_map/struct.BTreeMap.html#impl-FromIterator%3C(K,+V)%3E-for-BTreeMap%3CK,+V%3E), so its probably more efficient too!

Note: Collecting into a `HashMap` and `BTreeMap` will silently ignore collisions (IIRC it always takes the latest value). This is fine in this case because we are guaranteed to not have collisions (types.rs would fail to compile if there were multiple fn pointers with the same name). If you cared about dupes, you'd have to do something like what [i did for our custom serializer](https://github.com/microsoft/windows-drivers-rs/blob/1dd243ef07e470ec2f6a4973754bb30fe930f69d/crates/wdk-build/src/metadata/ser.rs#L130)

wmmc88 (changes requested):
Thanks for digging into this and making this change! I've definitely been meaning to make this change for a long while now, but never had the time to :) Glad you could implement it and that its showing significant benefit

I requested various comments mostly on testing/structuring for testing, some small style things, and some potential perf bumps.

crates/wdk-macros/src/lib.rs
leon-xd: extension traits strike again!! that makes a whole lotta sense.

crates/wdk-macros/src/lib.rs
```diff
@@ -202,6 +230,119 @@ impl Inputs {
...
+                .map_err(|e| map_to_syn_error(self.wdf_function_identifier.span(), e))?;
+
+            if !fragment_info_map_path.exists() {
+                let generated_map = self.generate_fragment_info_map_from_types()?;
+                let generated_map_string = serde_json::to_string(&generated_map)
+                    .map_err(|e| map_to_syn_error(self.wdf_function_identifier.span(), e))?;
+                std::fs::write(&fragment_info_map_path, generated_map_string)
+                    .map_err(|e| map_to_syn_error(self.wdf_function_identifier.span(), e))?;
+            }
+            FileExt::unlock(&flock)
```
wmmc88: I was wondering why you used the Fully Qualified Syntax here... I didn't realize Flocks are being added to std! (it actually just got its stabilization pr merged yesterday!)

wmmc88: created issue to track @ https://github.com/microsoft/windows-drivers-rs/issues/300

crates/wdk-macros/src/lib.rs
leon-xd: > putting all the code in this block in a function called generate_wdf_function_info_file_cache

This is kind of the idea of creating the second function -- this function is purely for the file lock stuff, and the other function is to generate the cache. renamed the function to `generate_wdf_function_info_file_cache` since that's a better name

wmmc88: My initial intention was for `generate_wdf_function_info_file_cache` to include all the flock stuff, since in my mind, "generating the file cache" includes actually writing it to disk. the function that creates the map in memory should still be named something like generate xyz map.

You can have `generate_wdf_function_info_file_cache` return the in-memory map of the cache and use that to early return from this function (after Flock unlock ofc)

crates/wdk-macros/src/lib.rs
leon-xd: added

crates/wdk-macros/src/lib.rs
leon-xd: changed

crates/wdk-macros/src/lib.rs
leon-xd: updated

crates/wdk-macros/src/lib.rs
leon-xd: added

crates/wdk-macros/src/lib.rs
leon-xd: added

crates/wdk-macros/src/lib.rs
leon-xd: added -- thank you for the extensive feedback!

crates/wdk-macros/src/lib.rs
leon-xd: fixed -- good catch :D

crates/wdk-macros/src/lib.rs
leon-xd: added

leon-xd: Fixed. Comes from outer space I think.

crates/wdk-macros/src/lib.rs
leon-xd: Added with an attempt at active voice

crates/wdk-macros/src/lib.rs
leon-xd: Still need to address this follow up

crates/wdk-macros/src/lib.rs
leon-xd: fixed

crates/wdk-macros/src/lib.rs
leon-xd: fixed

crates/wdk-macros/src/lib.rs
leon-xd: fixed

crates/wdk-macros/src/lib.rs
leon-xd: Fixed -- the old way made me raise my eyebrows for sure.

crates/wdk-macros/src/lib.rs
leon-xd: Removed

crates/wdk-macros/src/lib.rs
leon-xd: This plays nicely with `filter_map` so I kept it in in the revision. If you feel strongly still I can remove.

crates/wdk-macros/src/lib.rs
leon-xd: Moved

crates/wdk-macros/src/lib.rs
leon-xd: This comment made things a lot easier, fixed

crates/wdk-macros/src/lib.rs
leon-xd: Fixed. Shoulda just trusted my gut -- was initially using BTreeMap at first and then read [this](https://doc.rust-lang.org/std/collections/index.html):

> To get this out of the way: you should probably just use [Vec](https://doc.rust-lang.org/std/vec/struct.Vec.html) or [HashMap](https://doc.rust-lang.org/std/collections/struct.HashMap.html). These two collections cover most use cases for generic data storage and processing. They are exceptionally good at doing what they do. All the other collections in the standard library have specific use cases where they are the optimal choice, but these cases are borderline niche in comparison. Even when Vec and HashMap are technically suboptimal, they’re probably a good enough choice to get started.

crates/wdk-macros/src/lib.rs
leon-xd: Added!

crates/wdk-macros/src/lib.rs
leon-xd: whelp, looks like I'll be back in this neck of the woods in June!

crates/wdk-macros/src/lib.rs
wmmc88: yeah normally I'd agree with that, but since this is an internal detail + they're basically drop in replacements for what we're using them for + I previously did some legwork for evaluated these for similar use case, I thought I'd chime in :)

crates/wdk-macros/tests/unit-tests-input/generated-types-robust.rs
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft Corporation
+// License: MIT OR Apache-2.0
+
+//! Snippet of a bindgen-generated file containing types information used by tests for [`wdk_macros::call_unsafe_wdf_function_binding!`]
+
+pub mod _WDFFUNCENUM {
+    pub const WdfDriverCreateTableIndex: Type = 116;
+    pub const WdfVerifierDbgBreakPointTableIndex: Type = 367;
+    // Some entries in `WDFFUNCENUM` do not correspond to bindgen-generated functions.
```
wmmc88:
```suggestion
    // Some entries in `WDFFUNCENUM` do not correspond to bindgen-generated functions(ex. private WDF APIs). 
```

crates/wdk-macros/tests/unit-tests-input/no-wdf-func-enum.rs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation
+// License: MIT OR Apache-2.0
+
+//! Snippet of a bindgen-generated file containing types information used by tests for [`wdk_macros::call_unsafe_wdf_function_binding!`]
```
wmmc88: rename file to "missing-wdf-func-enum" to be more explicit

crates/wdk-macros/tests/unit-tests-input/no-wdf-func-enum-contents.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation
```
wmmc88: rename file to "missing-wdf-func-enum-contents" to be more explicit

crates/wdk-macros/tests/unit-tests-input/generated-types-robust.rs
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft Corporation
+// License: MIT OR Apache-2.0
+
+//! Snippet of a bindgen-generated file containing types information used by tests for [`wdk_macros::call_unsafe_wdf_function_binding!`]
+
+pub mod _WDFFUNCENUM {
```
wmmc88: Does this file actually need to exist or can you just add these extra entries to `generate-types.rs`?

crates/wdk-macros/src/lib.rs
```diff
@@ -41,6 +43,9 @@ use syn::{
     TypePath,
 };
 
+// A const &str to represent the name of the _WDFFUNCENUM module.
```
wmmc88:
```suggestion
/// Name of the `bindgen`-generated Rust module that contains the `TableIndex` constants for the `WDF`'s function table
```

crates/wdk-macros/src/lib.rs
```diff
@@ -374,6 +428,113 @@ fn parse_types_ast(path: &LitStr) -> Result<File> {
...
+/// it exists. If not, this function creates the cache by reading the
+/// `types.rs` file. The information is returned as a `BTreeMap`, where
+/// `key` is the function name and `value` is the cached function table
+/// information.
+///
+/// This function reduces build time by caching function table
+/// information. Instead of parsing `types.rs` for relevant data on
+/// every macro invocation, this function stores all possible function
+/// table information during the first invocation using the `scratch`
+/// and `serde` crates. Subsequent invocations can read from this cache.
```
wmmc88:
```suggestion
/// Fetch the function table information from the cache, if
/// it exists. If not, create the cache by reading the
/// `types.rs` file. Returns a `BTreeMap`, where
/// `key` is the function name and `value` is the cached function table
/// information.
///
/// Instead of parsing `types.rs` for relevant data on
/// every macro invocation, all relevant function
/// table information is extracted during the first `proc-macro` invocation and serialized to a location accessible by all proc-macro invocations. Subsequent invocations fetching from the cache significantly reduces compilation time
```

Led you a little astray with the copilot suggestion oops.

crates/wdk-macros/src/lib.rs
```diff
@@ -374,6 +428,113 @@ fn parse_types_ast(path: &LitStr) -> Result<File> {
...
+fn get_wdf_function_info_map(
+    types_path: &LitStr,
+    span: Span,
+) -> Result<BTreeMap<String, CachedFunctionInfo>> {
+    let scratch_dir = scratch::path(concat!(env!("CARGO_CRATE_NAME"), "_ast_fragments"));
+    let flock = std::fs::File::create(scratch_dir.join(".lock")).to_syn_result(span)?;
+
+    let cached_function_fragments_map_path = scratch_dir.join("cached_function_fragments_map.json");
```
wmmc88:
```suggestion
    let cached_function_info_map_path = scratch_dir.join("cached_function_info_map.json");
```

nit: just for consistency with new names

crates/wdk-macros/src/lib.rs
```diff
@@ -374,6 +428,113 @@ fn parse_types_ast(path: &LitStr) -> Result<File> {
...
+    Ok(map)
+}
+
+/// This function parses from `self.types_path` to generate a `BTreeMap` of
+/// function information, where `key` is the function name and `value` is
+/// the cached function table information. The function parses the
+/// `types.rs` file for `mod _WDFFUNCENUM`, which stores each function with
+/// the suffix `TableIndex`. It uses this to parse the function pointer type
+/// alias, then parses for the function parameters and return type.
+fn generate_wdf_function_info_file_cache(
```
wmmc88:
```suggestion
/// Parse file from `types_path` to generate a `BTreeMap` of
/// function information, where `key` is the function name and `value` is
/// the cached function table information.
fn generate_wdf_function_info_file_cache(
```

a bit too much implementation detail here. Most is implied from the type structure of `CachedFunctionInfo`. Too much implementation info makes comments that easily become outdated

crates/wdk-macros/src/lib.rs
```diff
@@ -58,6 +63,19 @@ trait StringExt {
     fn to_snake_case(&self) -> String;
 }
 
+/// A trait to provide mapping of `std::result::Result` to `syn::Result`
+trait ToSynResult<T, E> {
+    fn to_syn_result(self, span: Span) -> syn::Result<T>;
+}
```
wmmc88:
```suggestion
/// A trait to provide additional functionality to `std::result::Result`
trait ResultExt<T, E> {
    fn to_syn_result(self, span: Span) -> syn::Result<T>;
}
```

Convention is to name extensions traits as <Type being extended>Ext

crates/wdk-macros/src/lib.rs
```diff
@@ -131,6 +149,24 @@ impl StringExt for String {
     }
 }
 
+impl<T, E: std::fmt::Display> ToSynResult<T, E> for std::result::Result<T, E> {
```
wmmc88: Doesnt really matter that much here, but is there a reason you chose the loose bound of `E: Display` instead of `E: Error`? `Error` is already a subtrait of `Display`

crates/wdk-macros/tests/unit-tests-input/generated-types-robust.rs
leon-xd: Added

crates/wdk-macros/src/lib.rs
wmmc88: no this is good. I misread the let as an if let.

crates/wdk-macros/src/lib.rs
leon-xd: Added line break we discussed

crates/wdk-macros/src/lib.rs
```diff
@@ -392,14 +557,19 @@ fn parse_types_ast(path: &LitStr) -> Result<File> {
...
-    parse_fn_pointer_definition(fn_pointer_definition, function_pointer_type.span())
+) -> Result<Option<CachedFunctionInfo>> {
+    match find_type_alias_definition(types_ast, function_pointer_type) {
+        Ok(type_alias_definition) => {
+            let fn_pointer_definition =
+                extract_fn_pointer_definition(type_alias_definition, function_pointer_type.span())?;
+            parse_fn_pointer_definition(fn_pointer_definition, function_pointer_type.span())
+                .map(Some)
+        }
+        Err(_err) => Ok(None),
```
wmmc88, line 589: add a short comment on why Err == None

crates/wdk-macros/src/lib.rs
```diff
@@ -525,12 +695,12 @@ fn extract_fn_pointer_definition(type_alias: &ItemType, error_span: Span) -> Res
 fn parse_fn_pointer_definition(
     fn_pointer_typepath: &TypePath,
     error_span: Span,
-) -> Result<(Punctuated<BareFnArg, Token![,]>, ReturnType)> {
+) -> Result<CachedFunctionInfo> {
```
wmmc88: you should keep the return type as `Result<(Punctuated<BareFnArg, Token![,]>, ReturnType)>`  and just call .into at the callsite of this function

crates/wdk-macros/src/lib.rs
leon-xd: Actually, I ended up implementing the function we discussed -- the line break triggered a `cargo fmt` warning.

crates/wdk-macros/src/lib.rs
```diff
@@ -970,9 +1155,254 @@ mod tests {
...
             }
         }
+
+        mod get_wdf_function_info_map {
+            use serial_test::serial;
+
+            use super::*;
+
+            #[test]
+            #[serial]
```
wmmc88: Per offline discussion, I don't like the additional dep introduced just for these 2 tests. I expect for stateful tests the environment is reset after the test.

wmmc88: Here's an example of a function used to make sure that the stateful tests that execute do not interfere with eachother: https://github.com/leon-xd/windows-drivers-rs/blob/optimize-macro-expansion/crates/wdk-build/src/lib.rs#L1149

Similarily, here, you can:
1. grab flock
2. do test
3. reset env
4. release flock

this also gets rid of the "delete cache if it exists" logic that happens before your real test logic starts. You can wrap the above 4 steps into a reusuable function like i did in `with_env`

crates/wdk-macros/tests/unit-tests-input/no-wdf-func-enum.rs
leon-xd: Changed

crates/wdk-macros/tests/unit-tests-input/no-wdf-func-enum-contents.rs
leon-xd: Changed

crates/wdk-macros/tests/unit-tests-input/generated-types-robust.rs
leon-xd: You're right -- added.

crates/wdk-macros/src/lib.rs
leon-xd: Changed

crates/wdk-macros/src/lib.rs
leon-xd: Changed

crates/wdk-macros/src/lib.rs
leon-xd: Changed

crates/wdk-macros/src/lib.rs
leon-xd: Changed

crates/wdk-macros/src/lib.rs
leon-xd: Changed

crates/wdk-macros/src/lib.rs
leon-xd: True, I changed it to `std::error::Error`. I assumed the different crates & errors we were dealing with were not implemented by `std::error::Error` but it seems to work fine.

crates/wdk-macros/src/lib.rs
leon-xd: Added

crates/wdk-macros/src/lib.rs
leon-xd: Changed

crates/wdk-macros/src/lib.rs
```diff
@@ -334,6 +390,129 @@ fn call_unsafe_wdf_function_binding_impl(input_tokens: TokenStream2) -> TokenStr
...
+            );
+            generate_cached_function_info(&types_ast, &function_pointer_type)
+                .transpose()
+                .map(|cached_function_info_result| {
+                    cached_function_info_result
+                        .map(|cached_function_info| (function_name, cached_function_info))
+                })
+        })
+        .collect()
+}
```
wmmc88:
```suggestion
    func_enum_mod_contents
        .iter()
        .filter_map(|item| {
            if let Item::Const(function_table_enum_const) = item {
                function_table_enum_const
                    .ident
                    .to_string()
                    .strip_suffix("TableIndex")
                    .and_then(|function_name| {
                        let function_pointer_type = format_ident!(
                            "PFN_{uppercase_c_function_name}",
                            uppercase_c_function_name = function_name.to_uppercase(),
                            span = span
                        );

                        generate_cached_function_info(&types_ast, &function_pointer_type)
                            .transpose()
                            .map(|generate_cached_function_info_result| {
                                generate_cached_function_info_result.map(|cached_function_info| {
                                    (function_name.to_string(), cached_function_info)
                                })
                            })
                    })
            } 
            None
        })
        .collect()
}
```

I think this is better? got rid of one of the maps, got rid of some level of nesting

leon-xd: ah shoot yeah the back to back filter_maps I tried to avoid, looks like one slipped through.

crates/wdk-macros/src/lib.rs
leon-xd: Done

crates/wdk-macros/src/lib.rs
leon-xd: Added this change

crates/wdk-macros/src/lib.rs
```diff
@@ -392,14 +571,21 @@ fn parse_types_ast(path: &LitStr) -> Result<File> {
...
-    let fn_pointer_definition =
-        extract_fn_pointer_definition(type_alias_definition, function_pointer_type.span())?;
-    parse_fn_pointer_definition(fn_pointer_definition, function_pointer_type.span())
+) -> Result<Option<CachedFunctionInfo>> {
+    match find_type_alias_definition(types_ast, function_pointer_type) {
+        Ok(type_alias_definition) => {
+            let fn_pointer_definition =
+                extract_fn_pointer_definition(type_alias_definition, function_pointer_type.span())?;
+            parse_fn_pointer_definition(fn_pointer_definition, function_pointer_type.span())
+                .map(|fn_pointer_definition| Some(fn_pointer_definition.into()))
```
wmmc88:
```suggestion
            Some(parse_fn_pointer_definition(fn_pointer_definition, function_pointer_type.span())?.into())
```

wmmc88 (approved):

crates/wdk-macros/src/lib.rs
leon-xd: This didn't work -- changed it to:
```
Ok(Some(
    parse_fn_pointer_definition(fn_pointer_definition, function_pointer_type.span())?
        .into(),
))
```

wmmc88 (approved):

gurry:
> 2\. Adding it as part of our template for `cargo-wdk`

We can certainly add it to our templates. Just need to investigate and establish which build dependencies should receive this treatment.

FYI @krishnakumar4a4 and @svasista-ms. No action needed right now but something to consider for later.

hamzamust (approved):

PR #297, opened 2025-02-19
fix: use absolute paths for items used in PAGED_CODE macro

leon-xd:
Drivers that use the `wdk_sys::PAGED_CODE` macro used to have to separately import `wdk_sys::ntddk::KeGetCurrentIrql` and `wdk_sys::APC_LEVEL` in order to get the expanded macro to compile.

This change introduces absolute paths in `wdk_sys::PAGED_CODE` so that developers do not have to introduce these `use` statements to their code.


wmmc88 (approved):
(i edited your pr title a bit to be more specific) but nice catch! Thanks for the fix :)

wmmc88:
It'd be nice to have a test for this, but it'd rely on the automatic mockgeneration work that is still being explored

NateD-MSFT (approved):

PR #298, opened 2025-02-19
Added conventional commits spec to `CONTRIBUTING.md`

leon-xd:
As discussed a bit ago, added a link to the conventional commits spec in the `CONTRIBUTING.md` file.
Also reworded and cleaned up the whole Submitting a Pull Request section.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,6 +70,8 @@ Before you submit your Pull Request (PR) consider the following guidelines:
 * Commit your changes using a descriptive commit message
 * Push your fork to GitHub:
 * In GitHub, create a pull request
+  * Make sure to use the [Conventional Commits](https://www.conventionalcommits.org) specification
+   for the title of your PR.
 * If we suggest changes then:
   * Make the required updates.
   * Rebase your fork and force push to your GitHub repository (this will update your Pull Request):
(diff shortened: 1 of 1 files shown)


NateD-MSFT (approved):

CONTRIBUTING.md
```diff
@@ -62,14 +62,16 @@ You can file new issues by providing the above information at the corresponding
...
-* Make your changes in a new git fork:
+* Fork the repository and make changes on your fork locally.
 
-* Commit your changes using a descriptive commit message
-* Push your fork to GitHub:
-* In GitHub, create a pull request
+* Commit your changes using a descriptive commit message.
+* Push your commit to your forked repository.
+* In GitHub, create a pull request to merge your fork to `main`.
+  * Make sure to use the [Conventional Commits](https://www.conventionalcommits.org) specification
```
wmmc88: I wonder if we should enforce this automatically?

wmmc88: Maybe with https://github.com/marketplace/actions/semantic-pull-request? Not a blocker for this pr but worth considering

leon-xd: I could do this for this PR - sounds like a pretty easy fix and it's within scope. Will take a bit longer as we should discuss the full set of types that we would like to allow

PR #301, opened 2025-02-20
test: avoid flock contention in `wdk-macro-tests`

wmmc88:
This pull request includes several updates to the `tests/wdk-macros-tests/src/lib.rs` file to improve the handling of symbolic links and ensure proper file locking. The most important changes are to add a second condition check before attempting to lock the flock, and explicitly unlocking the flock as early as possible. These changes should marginally improve the performance of the `wdk-macro-tests`

Improvements to symbolic link handling and file locking:

* Refactored `_create_symlink_if_nonexistent` to include a new closure `link_needs_update` for checking if a symlink needs updating, ensuring the symlink is created or updated only when necessary.
* Added extra condition check to prevent needless flock locking
* Added explicit unlocking of `target_file` to avoid waiting for Windows to automatically release the lock, ensuring the lock is released promptly.

Minor formatting adjustments:

* Adjusted formatting of the `expand_args` function call to improve readability.
* Updated comments to use consistent capitalization.

### DIFF
--- a/tests/wdk-macros-tests/src/lib.rs
+++ b/tests/wdk-macros-tests/src/lib.rs
@@ -71,7 +71,7 @@ macro_rules! generate_macrotest_tests {
                         let symlink_target = &$crate::MACROTEST_INPUT_FOLDER_PATH.join(format!("{}.rs", stringify!($filename)));
                         let symlink_path = &$crate::MACROTEST_OUTPUT_FOLDER_PATH.join(format!("{}.rs", stringify!($filename)));
                         $crate::_create_symlink_if_nonexistent(symlink_path, symlink_target);
-                        $crate::expand(                            symlink_path);
+                        $crate::expand(symlink_path);
                     }
                 )?
 
@@ -84,8 +84,7 @@ macro_rules! generate_macrotest_tests {
                             let symlink_target = &$crate::MACROTEST_INPUT_FOLDER_PATH.join(format!("{}.rs", stringify!($filename)));
                             let symlink_path = &$crate::MACROTEST_OUTPUT_FOLDER_PATH.join(format!("{}.rs", stringify!($filename)));
                             $crate::_create_symlink_if_nonexistent(symlink_path, symlink_target);
-                            $crate::expand_args(
-                                symlink_path, &["--features", "nightly"]);
+                            $crate::expand_args(symlink_path, &["--features", "nightly"]);
                         }
                     )?
                 }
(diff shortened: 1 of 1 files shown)


tests/wdk-macros-tests/src/lib.rs
```diff
@@ -230,37 +229,48 @@ macro_rules! generate_call_unsafe_wdf_binding_tests {
...
+            || std::fs::read_link(link).expect("read_link of symlink should succeed") != target
+    };
+    if !link.exists() || link_needs_update() {
+        // create flock based off target_file, so tests can run in parallel
+        target_file
+            .lock_exclusive()
+            .expect("exclusive lock should be successfully acquired");
+
+        if !link.exists() {
+            std::os::windows::fs::symlink_file(relative_target_path, link)
```
ankurung, line 257: can this fail?

tests/wdk-macros-tests/src/lib.rs
```diff
@@ -230,37 +229,48 @@ macro_rules! generate_call_unsafe_wdf_binding_tests {
...
-
-        std::os::windows::fs::symlink_file(relative_target_path, link)
-            .expect("symlink creation should succeed");
-    } else {
-        // symlink already exists and points to the correct place
+    let link_needs_update = || {
+        !link.is_symlink()
+            || std::fs::read_link(link).expect("read_link of symlink should succeed") != target
+    };
+    if !link.exists() || link_needs_update() {
```
ankurung, line 250: Do we really need to call these 2 functions 2 times (one for checking here and then inside the block)

tests/wdk-macros-tests/src/lib.rs
wmmc88: In the happy path, there's no reason why it should fail. But theoretically it could fail due to stuff like permission issues. This function returns a result, and the expect on the next line would translate any error into a panic

tests/wdk-macros-tests/src/lib.rs
wmmc88: yes. `!link.exists() || link_needs_update()` is what determines if we need to create or update the symlinks. 

Its checked once first to make sure we don't lock unless we actually need to create/update the symlinks. Its checked a second time after the flock to make sure that we **still** need to create/update the symlinks (i.e. while this thread was blocking on the flock, the thread that owned the flock could have created/updated the symlink already). 

This is an example of the common [double-checked locking idiom](https://en.wikipedia.org/wiki/Double-checked_locking) and is for more efficient locked resource access

ankurung (approved):

NateD-MSFT (approved):

PR #278, opened 2025-01-27
feat: expand wdk-sys coverage to include gpio and parallel ports related headers

wmmc88:
This pull request adds support for GPIO and Parallel Ports drivers to the `wdk-sys` crate. The changes include updates to the API subsets, configuration, feature flags, and binding generation for these new driver types.

### API Subset Additions:
* Added `Gpio` and `ParallelPorts` to the `ApiSubset` enum in `crates/wdk-build/src/lib.rs`.

### Configuration Updates:
* Updated the `Config` implementation to include header files for `Gpio` and `ParallelPorts` based on the driver configuration in `crates/wdk-build/src/lib.rs` [[1]](diffhunk://#diff-43524f733cccc1a175c3bf83309866acb53b8cadb1628478ca1c87d31dda2b26R665-R673) [[2]](diffhunk://#diff-43524f733cccc1a175c3bf83309866acb53b8cadb1628478ca1c87d31dda2b26R687-R695).

### Feature Flags:
* Added `gpio` and `parallel-ports` features to `Cargo.toml` files for `wdk-sys` and example drivers [[1]](diffhunk://#diff-75c56254d2df2c16d14d51844a91e099874f56e1400992d6ec2bbc4d44796d9fR38-R41) [[2]](diffhunk://#diff-9b71ff41d80a1c3d6748742d1335fc1a4e78a79cd5e8b1695ac6f930938c31faR35-R38) [[3]](diffhunk://#diff-4da291dd13b35fc31bf456955fb12aa866aa18404cf0b8b1bb500971b0c64efbR33-R36) [[4]](diffhunk://#diff-45d1b36fd9f89d478f6dc3442004243ba9a67136f30f0de15effa9d1c92e1242R34-R37).

### Binding Generation:
* Implemented binding generation functions `generate_gpio` and `generate_parallel_ports` in `crates/wdk-sys/build.rs` [[1]](diffhunk://#diff-2006bfd5c1bb8d3bf7c95ff14e2a7ddd94fa13d1ce6cbf293b90138a903d413fR299-R334) [[2]](diffhunk://#diff-2006bfd5c1bb8d3bf7c95ff14e2a7ddd94fa13d1ce6cbf293b90138a903d413fR371-R406).
[...]

PR #307, opened 2025-03-07
Add `fltkernel.h` to base headers

wedsonaf:
This allows rust drivers to use fltmgr.


wmmc88 (commented):
It looks like the change is causing redefinitions of the GUIDs when other cargo features are enabled (ex. HID + STORAGE). this seems to be caused by `fltkernel.h` adding `initguid.h` which is now causing all the other guids in other headers to be both declared and (re-)defined.

In `bindgen_header_contents`, you can special-case the `ApiSubset` case that includes `fltkernel.h` to also prepend something like:

```c
// Force DEFINE_GUID to only declare GUIDS, and not define them (regardless if downstream headers try to include initguid.h)
#include <initguid.h>
#undef INITGUID
#include <guiddef.h>
```

wmmc88:
Another thing of note is that `fltkernel.h` seems to be part of the [`Filesystem` subset of the WDK](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/_ifsk/). I think it makes sense to separate it out as its own cargo feature, similar to what was done in #260, #263, #278, #287, etc.

PR #314, opened 2025-03-20
fix!: specify rust version & edition to wdk-default bindgen::builder

leon-xd:
BREAKING CHANGES:
- introduces MSRV via `rust-version` to virtual manifest, currently set to ~`1.82.0`~ `1.84.0`

# Overview

With the introduction of [Rust 2024](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/index.html), `bindgen` implemented [new options](https://github.com/rust-lang/rust-bindgen/issues/2998) for supporting different editions of Rust. `bindgen`'s default behavior is to generate Rust code for the latest edition and toolchain. 

We need to overwrite this default behavior with whatever edition `wdk-build` uses, as we would like to manually update from 2021 -> 2024. This PR also introduces logic to specify the rust version & toolchain to `bindgen`.

# Files changed

`cargo.toml` 
- added `rust-version`, set to ~`1.82.0`~ `1.84.0`
- bump `resolver` to `3`
- bump `bindgen` to `0.71.0` for `Builder::rust-edition`
- bump `cargo-metadata` to `0.19.2` for `Edition::E2024`
- add `semver` crate
- bump `thiserror` to `2.0.12` for `cargo-metadata` version dependency

`crates/wdk-build/cargo.toml`
- added relevant crates
- added nightly feature

`crates/wdk-build/src/bindgen.rs`
- `impl TryFrom` for conversion between `cargo-metadata` and `bindgen` rust edition enums
- `fn get_rust_target` determines proper rust version/toolchain for `bindgen`
- `fn get_rust_edition` determines proper rust edition for `bindgen`
- added calls to target and edition for `wdk_default`

`crates/wdk-build/src/cargo_make.rs`
- fixed code that was incompatible with older version of `cargo-metadata`

`crates/wdk-build/src/lib.rs`
- added `ConfigError` variants for new errors thrown by updated code


Cargo.toml
```diff
@@ -22,6 +22,7 @@ lto = true
 
 [workspace.package]
 edition = "2021"
+rust-version = "1.82.0"
```
wmmc88: since we are introducing msrv, we should probably also be using [the msrv aware solver in our own packages](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions) (so that we don't declare our msrv as 1.82, but actually pull in dependencies that have higher msrvs).

leon-xd:
I have created an issue to the `bindgen` team to discuss why this build is failing: https://github.com/rust-lang/rust-bindgen/issues/3177

Switching from `bindgen` 0.69 -> 0.71 means size and alignment checks happen at compile-time rather than runtime. Looking into the old versions of the USB header bindings, these were generated in old builds unsuccessfully, but since we never used them this was never discovered.

crates/wdk-build/src/bindgen.rs
```diff
@@ -123,6 +123,11 @@ impl BuilderExt for Builder {
             .blocklist_item(".*WHEA_ARM_BUS_ERROR(?:__bindgen.*)?")
             .blocklist_item(".*WHEA_ARM_PROCESSOR_ERROR")
             .blocklist_item(".*WHEA_ARM_CACHE_ERROR")
+            // FIXME: bindgen unable to generate for anonymous structs
```
wmmc88, line 126: link the relevant issue for tracking please

Cargo.toml
leon-xd: updated

crates/wdk-build/src/bindgen.rs
leon-xd: added

Cargo.toml
```diff
@@ -54,8 +55,9 @@ rustversion = "1.0.19"
 scratch = "1.0"
 serde = "1.0"
 serde_json = "1.0"
+semver = "1.0.26"
 syn = "2.0.100"
-thiserror = "1.0.69"
+thiserror = "2.0.12"
```
wmmc88, line 60: was there a reason you wanted to increase this here (rather than let dependabot bump it)

crates/wdk-build/src/cargo_make.rs
```diff
@@ -977,7 +977,7 @@ pub fn package_driver_flow_condition_script() -> anyhow::Result<()> {
         if !current_package
             .targets
             .iter()
-            .any(|target| target.kind.iter().any(|kind| kind == "cdylib"))
+            .any(|target| target.kind.contains(&cargo_metadata::TargetKind::CDyLib))
```
wmmc88, line 980: nice that `cargo_metadata` made this typesafe now

wmmc88, line 980: nice that `cargo_metadata` made this typesafe now

crates/wdk-build/src/bindgen.rs
```diff
@@ -137,3 +172,67 @@ impl WdkCallbacks {
...
+        return Ok(bindgen::RustTarget::nightly());
+    }
+
+    // Warn user if only one of these two conditions are enabled.
+    if (nightly_feature && !nightly_toolchain) || (!nightly_feature && nightly_toolchain) {
+        eprintln!(
+            "Nightly bindgen features are only enabled with nightly feature enablement and \
+             nightly toolchain use."
+        );
+    }
```
wmmc88: I think this is more readable as a match on tuples:

```suggestion
    let nightly_feature = cfg!(feature = "nightly");
    let nightly_toolchain = rustversion::cfg!(nightly);

    match (nightly_feature, nightly_toolchain) {
        (true, true) => Ok(bindgen::RustTarget::nightly()),
        (false, false) => {
            let package_msrv = semver::Version::parse(env!("CARGO_PKG_RUST_VERSION"))?;

            let bindgen_msrv = bindgen::RustTarget::stable(package_msrv.minor, package_msrv.patch)
                .map_err(|e| ConfigError::MsrvNotSupportedByBindgen {
                    msrv: package_msrv.to_string(),
                    reason: e.to_string(),
                })?;
            Ok(bindgen_msrv)
        }
        (true, false) | (false, true) => {
            eprintln!(
                "Nightly bindgen features are only enabled with nightly feature enablement and \
                 nightly toolchain use."
            );
        }
        
    }
```

crates/wdk-build/src/lib.rs
```diff
@@ -130,7 +130,34 @@ pub enum ConfigError {
...
+    },
+
+    /// Error returned when `bindgen` does not support `rust-version` in Cargo
+    /// manifest
+    #[error("Rust version {msrv} not supported by Bindgen: {reason}")]
+    MsrvNotSupportedByBindgen {
+        /// MSRV that was not supported by Bindgen
+        msrv: String,
+        /// Reason why the MSRV was not supported
+        reason: String,
```
wmmc88, line 151: rather than use the string, you can include the actual error. it helps with error backtraces and propogating other error trait info up. This is an example of where we do this: https://github.com/microsoft/windows-drivers-rs/blob/39b7290fb469fe97c3cd3ae4dcc6c1b3213e4814/crates/wdk-build/src/metadata/mod.rs#L80-L81

crates/wdk-build/src/bindgen.rs
```diff
@@ -137,3 +172,67 @@ impl WdkCallbacks {
...
+// semver version.
+fn get_rust_target() -> Result<bindgen::RustTarget, ConfigError> {
+    let nightly_feature = cfg!(feature = "nightly");
+    let nightly_toolchain = rustversion::cfg!(nightly);
+    if nightly_feature && nightly_toolchain {
+        return Ok(bindgen::RustTarget::nightly());
+    }
+
+    // Warn user if only one of these two conditions are enabled.
+    if (nightly_feature && !nightly_toolchain) || (!nightly_feature && nightly_toolchain) {
```
wmmc88: I'm wondering if we want to have this check at all. This breaks build if you try to compile with nightly feature on stable toolchain, right? I think elsewhere we have it so that nightly on stable just falls back to stable behavior

wmmc88: its also possible that some of the features we gate with "nightly" are more just experimental and unstable, rather than actually nightly features

crates/wdk-build/src/bindgen.rs
```diff
@@ -137,3 +172,67 @@ impl WdkCallbacks {
...
+//
+// # Errors
+//
+// Returns `ConfigError::CargoMetadataPackageNotFound` if the `wdk-build`
+// package is not found, or `ConfigError::UnsupportedRustEdition` if the edition
+// is not supported.
+fn get_rust_edition() -> Result<bindgen::RustEdition, ConfigError> {
+    let wdk_sys_cargo_metadata = MetadataCommand::new().exec()?;
+
+    let package_name = "wdk-build";
```
wmmc88: this can be a `const &str`

crates/wdk-build/src/bindgen.rs
```diff
@@ -137,3 +172,67 @@ impl WdkCallbacks {
...
+
+    let wdk_sys_package_metadata = wdk_sys_cargo_metadata
+        .packages
+        .iter()
+        .find(|package| package.name == package_name)
+        .ok_or_else(|| ConfigError::CargoMetadataPackageNotFound {
```
wmmc88: If this error is only used in this one place, it should probably be more specific

crates/wdk-build/src/lib.rs
```diff
@@ -130,7 +130,34 @@ pub enum ConfigError {
...
+    },
+
+    /// Error returned when `semver` parsing fails
+    #[error(transparent)]
+    SemverError(#[from] semver::Error),
```
wmmc88: Since this is only for semver parsing of wdkbuild, it may be better to have a more specific error name, and have this error just encapsulate the inner error (similar to the MsrvNotSupportedByBindgen suggestion)

Cargo.toml
```diff
@@ -12,7 +12,7 @@ exclude = [
   "tests/wdk-macros-tests",
   "tests/wdk-sys-tests",
 ]
-resolver = "2"
+resolver = "3"
```
wmmc88, line 15: Does this need also a pipeline change

wmmc88: https://doc.rust-lang.org/cargo/reference/config.html#resolverincompatible-rust-versions

Cargo.toml
leon-xd: bumping one of the dependencies (i forget which one) meant that it wouldn't build unless I bumped this in the main cargo.toml. I could have an incorrect understanding of how to resolve this kind of issue, however.

leon-xd: yeah in the pr description:
> bump thiserror to 2.0.12 **for cargo-metadata version dependency**

crates/wdk-build/src/cargo_make.rs
leon-xd: so nice he said it twice

Cargo.toml
wmmc88: ah i missed that

crates/wdk-build/src/bindgen.rs
leon-xd: hm.. if i follow the instructions of the comments below then we won't need to check for the true + false cases. so only true, true will be a special case and the rest of the code will follow "default" (stable) behavior. not sure if it's still idiomatic to use `match` in that case?

wmmc88: yeah, although you may want to emit a tracing `warn`. unfortunately, those aren't surfaced by default when building since all buildscript output is silenced by default (you need to enable via `cargo build -vv` at least afaik)

crates/wdk-build/src/lib.rs
leon-xd: the issue with this is that `bindgen` doesn't actually export the error type in the context of this error being called.

in `bindgen.rs`:
```
let bindgen_msrv = bindgen::RustTarget::stable(package_msrv.minor, package_msrv.patch)
        .map_err(|e| ConfigError::MsrvNotSupportedByBindgen {
            msrv: package_msrv.to_string(),
            reason: e.to_string(),
        })?;
```

the function [`stable`](https://docs.rs/bindgen/latest/bindgen/struct.RustTarget.html#method.stable) returns a `Result<RustTarget, InvalidRustTarget>`. `InvalidRustTarget` is in the [`features` mod](https://docs.rs/bindgen/latest/src/bindgen/features.rs.html#70-72) which is [selectively exposed](https://docs.rs/bindgen/latest/src/bindgen/lib.rs.html#53).

since `InvalidRustTarget` isn't even an `Error` (it only impls `Display`) I can't think of a way to do this other than the current implementation.

crates/wdk-build/src/bindgen.rs
leon-xd: removed this if statement

crates/wdk-build/src/bindgen.rs
leon-xd: added

crates/wdk-build/src/bindgen.rs
leon-xd: renamed

crates/wdk-build/src/lib.rs
leon-xd: changed

crates/wdk-build/src/bindgen.rs
leon-xd: given the hoops you'd jump through in order to see this output i just removed it entirely. are `warn`s prevalent throughout the build process?

wmmc88: There's a handful of warnings we emit from `wdk-build`. One example OTTOMH is when you build `wdk-sys` or anything other `wdk` library without selecting any WDK config

crates/wdk-build/src/lib.rs
wmmc88: ah thats unfortunate

Cargo.toml
wmmc88: I guess we don't technically verify with latest Dependencies in a job rn since we do - - locked. This can be punted to a separate task:

We should have:
* jobs that run with `--locked`
* jobs that run with `CARGO_RESOLVER_INCOMPATIBLE_RUST_VERSIONS=allow` w/o `--locked`: this checks latest dependencies work will work in workspaces that don't use the msrv-aware resolver
* jobs that run with `CARGO_RESOLVER_INCOMPATIBLE_RUST_VERSIONS=fallback` w/o `--locked: this checks latest dependencies work will work in workspaces that use the msrv-aware resolver
* jobs that specifically test on MSRV

leon-xd: created an issue linking to this comment

crates/wdk-build/src/bindgen.rs
leon-xd: fixed

hamzamust (dismissed):

hamzamust (approved):

crates/wdk-build/src/bindgen.rs
```diff
@@ -105,7 +138,20 @@ impl BuilderExt for Builder {
...
+
+        // The `_USBPM_CLIENT_CONFIG_EXTRA_INFO` struct only has members when
+        // _KERNEL_MODE flag is defined. We need to mark this type as opaque to avoid
+        // generating an empty struct, since  they are not currently supported by
+        // bindgen: https://github.com/rust-lang/rust-bindgen/issues/1683
+        if let DriverConfig::Umdf(_) = config.driver_config {
+            builder = builder
+                .blocklist_item("_USBPM_CLIENT_CONFIG_EXTRA_INFO")
+                .blocklist_item("USBPM_CLIENT_CONFIG_EXTRA_INFO")
+                .opaque_type("PUSBPM_CLIENT_CONFIG_EXTRA_INFO");
```
wmmc88: Just for clarity, what does `PUSBPM_CLIENT_CONFIG_EXTRA_INFO` get generated as in umdf now?

wmmc88 (approved):
LGTM!

crates/wdk-build/src/bindgen.rs
leon-xd: Based on our conversation yesterday I switched the opaque type to the second struct.

hamzamust (approved):

wmmc88 (dismissed):

wmmc88 (approved):

hamzamust (approved):

Issue #331, opened 2025-04-04
cargo clippy failure for crate level UMDF config


leon-xd:
Seems like running `cargo build --all-features` does not play nicely with the UMDF config uncommented in the root `cargo.toml`. Somehow, the `DRIVER_OBJECT` is not able to be detected in root. After investigation, this type is generated in the bindings.

A clean `cargo build` builds perfectly fine.

```
error[E0432]: unresolved import `crate::DRIVER_OBJECT`
  --> crates\wdk-sys\src\test_stubs.rs:18:13
   |
18 | use crate::{DRIVER_OBJECT, NTSTATUS, PCUNICODE_STRING};
   |             ^^^^^^^^^^^^^
   |             |
   |             no `DRIVER_OBJECT` in the root
   |             help: a similar name exists in the module: `PDRIVER_OBJECT`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `wdk-sys` (lib) due to 1 previous error
```

wmmc88:
Ah. `DRIVER_OBJECT` is only available in KM

PR #332, opened 2025-04-04
fix: passing cache tests when WDK config is enabled

leon-xd:
### Overview
Currently, our automated test matrix is not testing bindings when the WDK configuration is enabled at the project level. Locally, @wmmc88 discovered that the following `wdk-macros` tests failed when enabling the KMDF configuration:
- `tests::get_wdf_function_info_map::valid_input_no_cache`
- `tests::get_wdf_function_info_map::valid_input_cache_exists`
- `tests::inputs::generate_derived_ast_fragments::valid_input`
- `tests::inputs::generate_derived_ast_fragments::valid_input_with_no_arguments`.

These all test the feature I developed in #295 .

 ### Root cause
These tests failed with configuration turned on because of the `wdk` crate. We [conditionally compile](https://github.com/microsoft/windows-drivers-rs/blob/0ae8259826784334cf197bc9355b2ef011b70ec1/crates/wdk/src/lib.rs#L37) the abstractions contained in the `wdf` mod based on whether KMDF/UMDF is enabled in our config. Within these abstractions, we use the `call_unsafe_wdf_function_binding` macro several times to call the bindings to the WDK functions.

Generating and saving the cache is triggered by the `call_unsafe_wdf_function_binding` macro. I used the [`scratch` crate](https://docs.rs/scratch/latest/scratch/) in order to determine the location of our saved cache -- by creating a scratch directory called [`wdk_macros_ast_fragments`](https://github.com/microsoft/windows-drivers-rs/blob/0ae8259826784334cf197bc9355b2ef011b70ec1/crates/wdk-macros/src/lib.rs#L409), we can write, save, and read the cache in a location that is accessible throughout the build process.

However, turning on a WDK configuration conflicted with assumptions made when writing tests for the cache generation. When all root WDK configurations were disabled, there was no call to the `call_unsafe_wdf_function_binding` macro, since the methods that used this macro in the `wdk` crate weren't compiled. Therefore, tests were written expecting a completely clean `scratch` directory.  This worked fine until this configuration was enabled, as the `call_unsafe_wdf_function_binding` macro was called, and the cache was generated *before tests ran*. This triggered [an assertion](https://github.com/microsoft/windows-drivers-rs/blob/0ae8259826784334cf197bc9355b2ef011b70ec1/crates/wdk-macros/src/lib.rs#L948) which checked for a completely clean environment, causing the failure of each of the tests.

### Solution
In order to fix the root cause of this issue, I clean out the environment before each test is ran. However, more issues arose thinking about this more.

This problem broke a few assumptions I made when initially writing the tests, leading me to rewrite a portion of both the test and cache generation code. My old implementation included a `test.lock` that tests relied on for synchronous access to the `scratch` directory, as acquiring the lock used in the actual code resulted in a double acquire, and hence a deadlock. However, because now I know other crates can mutate the `scratch` directory, having separate locks for tests and implementation is not a clear enough guarantee to prevent a data race (even though Cargo runs build & test separately, this didn't feel like a strong enough invariant to prevent).

This lead me to conditionally compile two different versions of the function `get_wdf_function_info_map`. The `cfg(test)` version assumes an lock has already been obtained, and the `cfg(not(test)` version uses either a shared or exclusive lock depending on whether it needs to read or write to the cache. Now, the tests obtain the same lock as the actual code, ensuring that no data race exists between the `wdk` crate and the `wdk-macros` tests.

### Other issues
`cargo clippy --all-features` also had not been run with the configuration turned on, and many warnings prevented its completion when run locally. While this does not explicitly stop any of the pipeline checks we have today, when we eventually turn this on in our pipelines, these clippy warnings will have to be dealt with. Any `#[allow]` is a result of me silencing these warnings pre-emptively.

### Update 4-10
Instead of conditionally compiling two different versions of `get_wdf_function_info_map`, I instead conditionally compile two different locations for the `scratch` crate. This allows us to keep most of the old logic.

I also raised #331 as a result of tinkering with running `cargo build/clippy --all-features` with the UMDF config enabled.

### DIFF
--- a/crates/wdk-macros/src/lib.rs
+++ b/crates/wdk-macros/src/lib.rs
@@ -450,8 +450,7 @@ fn get_wdf_function_info_map(
                 types_path,
                 cached_function_info_map_path.as_path(),
                 span,
-            );
-            let function_info_map = function_info_map?;
+            )?;
             return Ok(function_info_map);
         }
     }
--- a/crates/wdk-macros/src/lib.rs
+++ b/crates/wdk-macros/src/lib.rs
@@ -411,13 +411,10 @@ fn get_wdf_function_info_map(
 
     let get_function_info_map_in_exclusively_locked_dir = || {
         if cached_function_info_map_path.exists() {
-            let function_info_map = read_wdf_function_info_file_cache(
-                &cached_function_info_map_path,
-                span,
-            )?;
+            let function_info_map =
+                read_wdf_function_info_file_cache(&cached_function_info_map_path, span)?;
             Ok::<_, syn::Error>(function_info_map)
-        }
-        else {
+        } else {
             let function_info_map = create_wdf_function_info_file_cache(
                 types_path,
                 &cached_function_info_map_path,
@@ -427,31 +424,26 @@ fn get_wdf_function_info_map(
         }
     };
 
-    // Tests need to obtain an exclusive lock on the scratch directory to create different environments. We conditionally compile here in order to avoid obtaining the lock again. 
+    // Tests need to obtain an exclusive lock on the scratch directory to create
+    // different environments. We conditionally compile here in order to avoid
+    // obtaining the lock again.
     #[cfg(test)]
-    return {
-        let function_info_map = get_function_info_map_in_exclusively_locked_dir();
-        function_info_map
-    };
+    {
+        get_function_info_map_in_exclusively_locked_dir()
+    }
 
-    // In non-test environments, if the cache exists we want to obtain a shared lock on the scratch directory to allow multiple threads to read the cache at once.
-    // Otherwise, we want to obtain an exclusive lock on the scratch directory to create the cache.
+    // In non-test environments, if the cache exists we want to obtain a shared lock
+    // on the scratch directory to allow multiple threads to read the cache at once.
+    // Otherwise, we want to obtain an exclusive lock on the scratch directory to
+    // create the cache.
     #[cfg(not(test))]
     if cached_function_info_map_path.exists() {
-        let function_info_map = with_shared_file_lock( 
-            &scratch_dir,
-            span,
-            || {
-                let read_map = read_wdf_function_info_file_cache(
-                    &cached_function_info_map_path,
-                    span
-                )?;
-                Ok(read_map)
-            }
-        )?;
+        let function_info_map = with_shared_file_lock(&scratch_dir, span, || {
+            let read_map = read_wdf_function_info_file_cache(&cached_function_info_map_path, span)?;
+            Ok(read_map)
+        })?;
         Ok(function_info_map)
-    }
-    else {
+    } else {
         let function_info_map = with_exclusive_file_lock(
             &scratch_dir,
             span,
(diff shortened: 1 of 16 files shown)


crates/wdk-macros/src/lib.rs
```diff
@@ -933,26 +995,31 @@ mod tests {
         LazyLock::new(|| scratch::path(concat!(env!("CARGO_CRATE_NAME"), "_ast_fragments")));
     const CACHE_FILE_NAME: &str = "cached_function_info_map.json";
 
-    fn with_file_lock<F>(f: F)
+    fn clean_cache_test_env(file_path: &PathBuf) {
+        if file_path.exists() {
+            std::fs::remove_file(file_path).unwrap();
+        }
+
+        pretty_assert_eq!(
```
wmmc88: If remove_file failed, wouldn't it just panic the test (i.e. this assert can never fire)?

crates/wdk-macros/src/lib.rs
```diff
@@ -933,26 +995,31 @@ mod tests {
...
+        let cached_function_info_map_path = SCRATCH_DIR.join(CACHE_FILE_NAME);
 
+        clean_cache_test_env(&cached_function_info_map_path);
         f();
-
-        if cached_function_info_map_path.exists() {
-            std::fs::remove_file(&cached_function_info_map_path).unwrap();
-        }
-        pretty_assert_eq!(cached_function_info_map_path.exists(), false);
+        clean_cache_test_env(&cached_function_info_map_path);
```
wmmc88: Don't you not need this 2nd clean now that every test cleans before execution?

wmmc88: oh or is it because you don't want the cache generated for tests to be re-used for actual execution. If that's the case please add inline comments

crates/wdk-macros/src/lib.rs
```diff
@@ -407,38 +407,100 @@ fn get_wdf_function_info_map(
...
     let scratch_dir = scratch::path(concat!(env!("CARGO_CRATE_NAME"), "_ast_fragments"));
-    let flock = std::fs::File::create(scratch_dir.join(".lock"))
-        .to_syn_result(span, "unable to create file")?;
-
     let cached_function_info_map_path = scratch_dir.join("cached_function_info_map.json");
 
-    if !cached_function_info_map_path.exists() {
-        FileExt::lock_exclusive(&flock).to_syn_result(span, "unable to obtain file lock")?;
-
-        // Before this thread acquires the lock, it's possible that a concurrent thread
```
wmmc88, line 418: is info from this comment lost now?

crates/wdk-macros/src/lib.rs
```diff
@@ -407,38 +407,100 @@ fn get_wdf_function_info_map(
...
-        // already created the cache. If so, this thread skips cache generation.
-        if !cached_function_info_map_path.exists() {
-            let generated_map = create_wdf_function_info_file_cache(
+    let get_function_info_map_in_exclusively_locked_dir = || {
```
wmmc88: why use a closure here? its clearer just to use a function + allows you to have doc comments

crates/wdk-macros/src/lib.rs
```diff
@@ -407,38 +407,100 @@ fn get_wdf_function_info_map(
...
-            return Ok(generated_map);
+            Ok(function_info_map)
         }
-        FileExt::unlock(&flock).to_syn_result(span, "unable to unlock file lock")?;
+    };
+
+    // Tests need to obtain an exclusive lock on the scratch directory to create
+    // different environments. We conditionally compile here in order to avoid
+    // obtaining the lock again.
+    #[cfg(test)]
```
wmmc88: If you're going to do this type of conditional compilation, using `cfg_if` makes it a lot clearer that only one of these two branches will ever get executed

wmmc88 (changes requested):
In the new logic, you are conditionally changing the locking behavior of a single lock, depending on if the code executing was compiled for test or normally. I think this results in fairly messy code. It is not immediately obvious that what's going on here is:
1. If cache doesnt exist and not in test, the lock is exclusively held (preventing tests from running)
2. If the cache does exist and not in test, the lock is non-exclusively held (preventing tests from running but allowing normal proc-macro execution
3. If the code is in test, regardless of if the cache exists or not, the lock is exclusively held and normal proc-macro execution is blocked

I think the reason the code feels unclear is largely because of how much the logic diverges between test-path and normal path, and I think we want to minimize that as much as possible. One way to do this is to revert the logic to how it was originally, but just have an extra test-specific lock. The normal path can acquire this "test" lock non-exclusively and the normal path does nothing.

crates/wdk-macros/src/lib.rs
leon-xd: Yes. I suppose I also view asserts as communicating invariants on logical state to the developer. If this one is too redundant given the above logic, I could remove it. I'd argue that this also helps the developer understand that no messy business happened with the file remove/detection logic.

crates/wdk-macros/src/lib.rs
leon-xd: yes -- don't want this to be re-used for execution. will add this comment

crates/wdk-macros/src/lib.rs
leon-xd: brought this comment back

crates/wdk-macros/src/lib.rs
leon-xd: removed

crates/wdk-macros/src/lib.rs
leon-xd: removed conditional here -- should I do `cfg_if` in the new version too?

crates/wdk-macros/src/lib.rs
leon-xd: Since we are using a test scratch directory now I am removing this

crates/wdk-macros/src/lib.rs
```diff
@@ -406,39 +406,53 @@ fn get_wdf_function_info_map(
...
 
         // Before this thread acquires the lock, it's possible that a concurrent thread
         // already created the cache. If so, this thread skips cache generation.
         if !cached_function_info_map_path.exists() {
-            let generated_map = create_wdf_function_info_file_cache(
+            let function_info_map = create_wdf_function_info_file_cache(
                 types_path,
                 &cached_function_info_map_path,
                 span,
             )?;
```
gurry: If this calls fails, the `?` operator will cause an early return and `FileExt::unlock` below won't run.

Consider using the lock through an RAII pattern, which should eliminate all such problems.

crates/wdk-macros/src/lib.rs
```diff
@@ -406,39 +406,53 @@ fn get_wdf_function_info_map(
...
+    let function_info_map =
+        read_wdf_function_info_file_cache(&cached_function_info_map_path, span)?;
+    Ok(function_info_map)
+}
 
-    let generated_map_string = std::fs::read_to_string(&cached_function_info_map_path)
+/// Reads the cache of function information, then deserializes it into a
+/// `BTreeMap`.
+fn read_wdf_function_info_file_cache(
+    cached_function_info_map_path: &PathBuf,
```
gurry:
```suggestion
    cached_function_info_map_path: &Path,
```
`&Path` is more flexible

leon-xd: fixed

crates/wdk-macros/src/lib.rs
leon-xd: added

crates/wdk-macros/src/lib.rs
leon-xd: i added this

crates/wdk-macros/src/lib.rs
```diff
@@ -406,42 +425,61 @@ fn get_wdf_function_info_map(
...
+        } else {
+            let scratch_dir = scratch::path(concat!(env!("CARGO_CRATE_NAME"), "_ast_fragments"));
+        }
+    }
 
     let cached_function_info_map_path = scratch_dir.join("cached_function_info_map.json");
 
     if !cached_function_info_map_path.exists() {
-        FileExt::lock_exclusive(&flock).to_syn_result(span, "unable to obtain file lock")?;
+        let flock = std::fs::File::create(scratch_dir.join(".lock"))
```
wmmc88, line 439: does moving this file creation down make it possible that you end up with the `.lock` not created and something trying to lock it (Ex. test)?

wmmc88: or is there a reason you moved it down?

crates/wdk-macros/src/lib.rs
wmmc88: technically the flock gets released implicitly (by the os) when this proc-macro process terminates(ex. from this error), but the new raii-style guard wrapper is more elegant anyways

crates/wdk-macros/src/lib.rs
leon-xd: Not anymore. This lock could've always been here. My initial implementation for this fix required a shared lock for reading, which isn't necessary anymore since we are using separate `scratch` directories. Reading the cache is in scope outside of this if statement, and because of that removal I can move this down to inside the if statement.

wmmc88 (approved):

gurry (approved):

PR #353, opened 2025-05-13
Implement aligned WDK allocator

### LINKED ISSUE #303: Ensure correct alignment in custom allocators
The [custom allocator ](https://github.com/microsoft/windows-drivers-rs/blob/f9752bde538d53f7bbd13538b27b92bd3626f4bd/crates/wdk-alloc/src/lib.rs#L57) in `wdk_alloc` is based on on `ExAllocPool2` which uses [a 16 byte alignment on 64 bit systems if the allocation size is < PAGE_SIZE and PAGE_SIZE otherwise](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-exallocatepool2#:~:text=If%E2%80%AFNumberOfBytes%E2%80%AFis,64%2Dbit%20systems.). We need to validate that this does not violate alignment requirements of Rust types.

Primitive Rust types should be fine. The biggest alignment requirement is 16 bytes for `u128` and `i128` which does not conflict with `ExAllocPool2`'s behaviour. 

We just need to check for composite types. As per the Rust reference Rust composite types with standard repr will have an [alignment _at least_ as big the maximum alignment of its fields](https://doc.rust-lang.org/reference/type-layout.html#the-rust-representation:~:text=The%20alignment%20of%20the%20type%20is%20at%20least%20the%20maximum%20alignment%20of%20its%20fields.). Types with `#[repr(align)]` can have arbitrary alignment requirements. The allocator should work correctly in all these cases.

Zero-Tang:
Now the WDK allocator can allocate aligned memory according to the layout argument.

Note that this allocator won't waste additional space for alignment if the alignment is normal (e.g.: less than or equal to 16-byte boundary on 64-bit platform). In average scenario, this is better than [`_aligned_malloc`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-malloc?view=msvc-170) and [`_aligned_free`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-free?view=msvc-170) in msvcrt, in that I guess we rarely need big alignments in drivers for some stuff in the pool. Still, some performance-critical codes might require cache-aligned (e.g.: 64-byte in x86) objects, which is usually bigger than `ExAllocatePool2`'s alignment.

Close #303.

### DIFF
--- a/crates/wdk-alloc/src/lib.rs
+++ b/crates/wdk-alloc/src/lib.rs
@@ -90,12 +90,15 @@ mod kernel_mode {
                         //   | ......... | p | ..................... | ...... |
                         //                   ^                       ^
                         //             aligned start            aligned end
-                        // "p" is where we store the original address returned by `ExAllocatePool2`.
-                        // We will also pass "p" to ExFreePoolWithTag on `dealloc` method.
                         // "aligned start" is the address we will return in `alloc` method.
+                        // The gap between the "aligned start" and the original address has a
+                        // guaranteed minimum size of two pointers, so it's safe to store this pointer.
+                        // "p" is where we store the original address returned by `ExAllocatePool2`.
+                        // We will also pass "p" to `ExFreePoolWithTag` on `dealloc` method.
+                        // This is also how `_aligned_free` in msvcrt.dll receives the original pointer.
                         unsafe {
                             // Store the original pointer right before the pointer to return,
-                            // so that ExFreePoolWithTag can receive the correct pointer.
+                            // so that `ExFreePoolWithTag` can receive the correct pointer.
                             // This is also how `_aligned_malloc` is implemented in msvcrt.dll
                             q.sub(1).write(p);
                         }
@@ -128,14 +131,14 @@ mod kernel_mode {
                 //                   ^                       ^
                 //             aligned start            aligned end
                 // "p" is where we store the original address returned by `ExAllocatePool2`.
-                // So we will pass "p" to ExFreePoolWithTag on `dealloc` method.
+                // So we will pass "p" to `ExFreePoolWithTag` on `dealloc` method.
                 // "aligned start" is the address we received from `ptr` argument in `dealloc` method.
                 unsafe { q.sub(1).read() }
             } else {
                 // The alignment is normal, so the ptr is already the original pointer.
                 ptr.cast()
             };
-            // SAFETY: `ExFreePool` is safe to call from any `IRQL` <= `DISPATCH_LEVEL`
+            // SAFETY: `ExFreePoolWithTag` is safe to call from any `IRQL` <= `DISPATCH_LEVEL`
             // since its freeing memory allocated from `POOL_FLAG_NON_PAGED` in `alloc`
             unsafe {
                 ExFreePoolWithTag(p, RUST_TAG);
--- a/crates/wdk-alloc/src/lib.rs
+++ b/crates/wdk-alloc/src/lib.rs
@@ -31,7 +31,7 @@ mod kernel_mode {
     use core::alloc::{GlobalAlloc, Layout};
 
     use wdk_sys::{
-        ntddk::{ExAllocatePool2, ExFreePoolWithTag}, PAGE_SIZE, POOL_FLAG_NON_PAGED, PVOID, SIZE_T, ULONG
+        ntddk::{ExAllocatePool2, ExFreePoolWithTag}, MEMORY_ALLOCATION_ALIGNMENT, PAGE_SIZE, POOL_FLAG_NON_PAGED, PVOID, SIZE_T, ULONG
     };
 
     /// Allocator implementation to use with `#[global_allocator]` to allow use
@@ -45,16 +45,15 @@ mod kernel_mode {
     // The value of memory tags are stored in little-endian order, so it is
     // convenient to reverse the order for readability in tooling (ie. Windbg).
     const RUST_TAG: ULONG = u32::from_ne_bytes(*b"rust");
-    
-    // The minimum alignment of pointer returned by ExAllocatePool2.
-    const POOL_ALIGNMENT: usize = size_of::<*mut u8>() * 2;
 
     #[inline] fn require_realignment(layout: Layout) -> bool {
-        if layout.align() <= POOL_ALIGNMENT {
-            false
-        } else {
-            layout.align() > PAGE_SIZE as usize || layout.size() < PAGE_SIZE as usize
-        }
+        // `ExAllocatePool2` uses an alignment of `PAGE_SIZE` or minimum pool
+        // alignment depending on the requested size. See documentation:
+        // https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-exallocatepool2#remarks
+        let chosen_alignment = if layout.size() >= PAGE_SIZE as usize { PAGE_SIZE } else { MEMORY_ALLOCATION_ALIGNMENT } as usize;
+        // Realignment needed only if requested alignment cannot
+        // be satisfied by `ExAllocatePool2`'s chosen alignment
+        layout.align() > chosen_alignment
     }
 
     // SAFETY: This is safe because the Wdk allocator:
@@ -73,6 +72,7 @@ mod kernel_mode {
                         ExAllocatePool2(POOL_FLAG_NON_PAGED, size as SIZE_T, RUST_TAG)
                     };
                     if !p.is_null() {
+                        // Align the pointer up to the first address that meets alignment requirement.
                         let q = ((p as usize & mask) + layout.align()) as *mut PVOID;
                         // Store the original pointer right before the pointer to return,
                         // so that ExFreePoolWithTag can receive the correct pointer.
@@ -101,9 +101,9 @@ mod kernel_mode {
             // SAFETY: `ExFreePool` is safe to call from any `IRQL` <= `DISPATCH_LEVEL`
             // since its freeing memory allocated from `POOL_FLAG_NON_PAGED` in `alloc`
             let p = if require_realignment(layout) {
-                // The alignment is too large, the original pointer is stored right before
+                // The alignment is too large, so the original pointer is stored right before
                 // the ptr. This is also how `_aligned_free` is implemented in msvcrt.dll
-                let q: *mut *mut u8 = ptr.cast();
+                let q: *mut PVOID = ptr.cast();
                 unsafe {
                     q.sub(1).read()
                 }
@@ -112,7 +112,7 @@ mod kernel_mode {
                 ptr.cast()
             };
             unsafe {
-                ExFreePoolWithTag(p.cast(), RUST_TAG);
+                ExFreePoolWithTag(p, RUST_TAG);
             }
         }
     }
--- a/crates/wdk-alloc/src/lib.rs
+++ b/crates/wdk-alloc/src/lib.rs
@@ -28,15 +28,10 @@ pub use kernel_mode::*;
 #[cfg(any(driver_model__driver_type = "WDM", driver_model__driver_type = "KMDF"))]
 mod kernel_mode {
 
-    use core::{
-        alloc::{GlobalAlloc, Layout},
-        sync::atomic::{AtomicU32, Ordering},
-    };
+    use core::alloc::{GlobalAlloc, Layout};
 
     use wdk_sys::{
-        ntddk::{ExAllocatePool2, ExFreePoolWithTag},
-        POOL_FLAG_NON_PAGED,
-        SIZE_T,
+        ntddk::{ExAllocatePool2, ExFreePoolWithTag}, PAGE_SIZE, POOL_FLAG_NON_PAGED, PVOID, SIZE_T, ULONG
     };
 
     /// Allocator implementation to use with `#[global_allocator]` to allow use
@@ -47,14 +42,20 @@ mod kernel_mode {
     /// <= `DISPATCH_LEVEL`
     pub struct WdkAllocator;
 
-    /// The value of memory tags are stored in little-endian order, so it is
-    /// convenient to reverse the order for readability in tooling (ie. Windbg).
-    /// The default tag value is `rust`. You may use `store` method to mutate its value.
-    /// However, it's strongly recommended that you mutate the tag for only once.
-    pub static GLOBAL_POOL_TAG: AtomicU32 = AtomicU32::new(u32::from_ne_bytes(*b"rust"));
+    // The value of memory tags are stored in little-endian order, so it is
+    // convenient to reverse the order for readability in tooling (ie. Windbg).
+    const RUST_TAG: ULONG = u32::from_ne_bytes(*b"rust");
     
     // The minimum alignment of pointer returned by ExAllocatePool2.
     const POOL_ALIGNMENT: usize = size_of::<*mut u8>() * 2;
+
+    #[inline] fn require_realignment(layout: Layout) -> bool {
+        if layout.align() <= POOL_ALIGNMENT {
+            false
+        } else {
+            layout.align() > PAGE_SIZE as usize || layout.size() < PAGE_SIZE as usize
+        }
+    }
 
     // SAFETY: This is safe because the Wdk allocator:
     //         1. can never unwind since it can never panic
@@ -64,26 +65,30 @@ mod kernel_mode {
         unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
             let ptr =
                 // SAFETY: `ExAllocatePool2` is safe to call from any `IRQL` <= `DISPATCH_LEVEL` since its allocating from `POOL_FLAG_NON_PAGED`
-                unsafe {
-                    if layout.align() > POOL_ALIGNMENT {
-                        // If the requested alignment is too large, we'll have to waste some space as large as the alignment.
-                        let size = layout.size() + layout.align();
-                        let mask = !(layout.align() - 1);
-                        let p = ExAllocatePool2(POOL_FLAG_NON_PAGED, size as SIZE_T, GLOBAL_POOL_TAG.load(Ordering::Relaxed));
-                        if !p.is_null() {
-                            let q = ((p as usize & mask) + layout.align()) as *mut *mut u8;
-                            // Store the original pointer right before the pointer to return,
-                            // so that ExFreePoolWithTag can receive the correct pointer.
-                            // This is also how `_aligned_malloc` is implemented in msvcrt.dll
-                            q.sub(1).write(p.cast());
-                            q.cast()
-                        } else {
-                            p
+                if require_realignment(layout) {
+                    // If the requested alignment is too large, we'll have to waste some space as large as the alignment.
+                    let size = layout.size() + layout.align();
+                    let mask = !(layout.align() - 1);
+                    let p = unsafe {
+                        ExAllocatePool2(POOL_FLAG_NON_PAGED, size as SIZE_T, RUST_TAG)
+                    };
+                    if !p.is_null() {
+                        let q = ((p as usize & mask) + layout.align()) as *mut PVOID;
+                        // Store the original pointer right before the pointer to return,
+                        // so that ExFreePoolWithTag can receive the correct pointer.
+                        // This is also how `_aligned_malloc` is implemented in msvcrt.dll
+                        unsafe {
+                            q.sub(1).write(p);
                         }
+                        q.cast()
                     } else {
-                        // Because we know the alignment while in both alloc and dealloc,
-                        // we don't need to waste space if the default alignment is small enough.
-                        ExAllocatePool2(POOL_FLAG_NON_PAGED, layout.size() as SIZE_T, GLOBAL_POOL_TAG.load(Ordering::Relaxed))
+                        p
+                    }
+                } else {
+                    // Because we know the alignment while in both alloc and dealloc,
+                    // we don't need to waste space if the default alignment is small enough.
+                    unsafe {
+                        ExAllocatePool2(POOL_FLAG_NON_PAGED, layout.size() as SIZE_T, RUST_TAG)
                     }
                 };
             if ptr.is_null() {
@@ -95,17 +100,19 @@ mod kernel_mode {
         unsafe fn dealloc(&self, ptr: *mut u8, layout: Layout) {
             // SAFETY: `ExFreePool` is safe to call from any `IRQL` <= `DISPATCH_LEVEL`
             // since its freeing memory allocated from `POOL_FLAG_NON_PAGED` in `alloc`
-            unsafe {
-                let p = if layout.align() > POOL_ALIGNMENT {
-                    // The alignment is too large, the original pointer is stored right before the ptr.
-                    // This is also how `_aligned_free` is implemented in msvcrt.dll
-                    let q: *mut *mut u8 = ptr.cast();
+            let p = if require_realignment(layout) {
+                // The alignment is too large, the original pointer is stored right before
+                // the ptr. This is also how `_aligned_free` is implemented in msvcrt.dll
+                let q: *mut *mut u8 = ptr.cast();
+                unsafe {
                     q.sub(1).read()
-                } else {
-                    // The alignment is normal, so the ptr is already the original pointer.
-                    ptr.cast()
-                };
-                ExFreePoolWithTag(p.cast(), GLOBAL_POOL_TAG.load(Ordering::Relaxed));
+                }
+            } else {
+                // The alignment is normal, so the ptr is already the original pointer.
+                ptr.cast()
+            };
+            unsafe {
+                ExFreePoolWithTag(p.cast(), RUST_TAG);
             }
         }
     }
--- a/crates/wdk-alloc/src/lib.rs
+++ b/crates/wdk-alloc/src/lib.rs
@@ -25,19 +25,16 @@
 #[cfg(any(driver_model__driver_type = "WDM", driver_model__driver_type = "KMDF"))]
 pub use kernel_mode::*;
 
+// Note: to run tests, go to "tests/wdk-alloc-tests" of the root repository
+// and run `cargo test --lib` command.
 #[cfg(any(driver_model__driver_type = "WDM", driver_model__driver_type = "KMDF"))]
 mod kernel_mode {
 
     use core::alloc::{GlobalAlloc, Layout};
 
-    use wdk_sys::{
-        ntddk::{ExAllocatePool2, ExFreePoolWithTag},
-        PAGE_SIZE,
-        POOL_FLAG_NON_PAGED,
-        PVOID,
-        SIZE_T,
-        ULONG,
-    };
+    use wdk_sys::{PAGE_SIZE, POOL_FLAGS, POOL_FLAG_NON_PAGED, PVOID, SIZE_T, ULONG};
+    #[cfg(not(test))]
+    use wdk_sys::ntddk::{ExAllocatePool2, ExFreePoolWithTag};
 
     /// Allocator implementation to use with `#[global_allocator]` to allow use
     /// of [`core::alloc`].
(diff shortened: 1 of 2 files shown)


crates/wdk-alloc/src/lib.rs
```diff
@@ -45,33 +47,65 @@ mod kernel_mode {
...
     pub struct WdkAllocator;
 
-    // The value of memory tags are stored in little-endian order, so it is
-    // convenient to reverse the order for readability in tooling (ie. Windbg)
-    const RUST_TAG: ULONG = u32::from_ne_bytes(*b"rust");
+    /// The value of memory tags are stored in little-endian order, so it is
+    /// convenient to reverse the order for readability in tooling (ie. Windbg).
+    /// The default tag value is `rust`. You may use `store` method to mutate its value.
+    /// However, it's strongly recommended that you mutate the tag for only once.
+    pub static GLOBAL_POOL_TAG: AtomicU32 = AtomicU32::new(u32::from_ne_bytes(*b"rust"));
```
wmmc88: I don't think we want to allow changing this tag in this fashion. at least not now and not for global allocator. 

It's possible that we may want to allow this for when we look into implementing `allocator_api` allocators

crates/wdk-alloc/src/lib.rs
```diff
@@ -45,33 +47,65 @@ mod kernel_mode {
...
+    
+    // The minimum alignment of pointer returned by ExAllocatePool2.
+    const POOL_ALIGNMENT: usize = size_of::<*mut u8>() * 2;
```
wmmc88: Can you derive this from a constant in the WDK? Where is this minimum alignment documented?

crates/wdk-alloc/src/lib.rs
```diff
@@ -45,33 +47,65 @@ mod kernel_mode {
...
+                    if layout.align() > POOL_ALIGNMENT {
+                        // If the requested alignment is too large, we'll have to waste some space as large as the alignment.
+                        let size = layout.size() + layout.align();
+                        let mask = !(layout.align() - 1);
+                        let p = ExAllocatePool2(POOL_FLAG_NON_PAGED, size as SIZE_T, GLOBAL_POOL_TAG.load(Ordering::Relaxed));
+                        if !p.is_null() {
+                            let q = ((p as usize & mask) + layout.align()) as *mut *mut u8;
+                            // Store the original pointer right before the pointer to return,
+                            // so that ExFreePoolWithTag can receive the correct pointer.
+                            // This is also how `_aligned_malloc` is implemented in msvcrt.dll
```
wmmc88, line 110: Where is this documented?

crates/wdk-alloc/src/lib.rs
```diff
@@ -45,33 +47,65 @@ mod kernel_mode {
...
     //         1. can never unwind since it can never panic
     //         2. has implementations of alloc and dealloc that maintain layout
-    //            constraints (FIXME: Alignment of the layout is currenty not
-    //            supported)
+    //            constraints (including size and alignment)
     unsafe impl GlobalAlloc for WdkAllocator {
         unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
             let ptr =
                 // SAFETY: `ExAllocatePool2` is safe to call from any `IRQL` <= `DISPATCH_LEVEL` since its allocating from `POOL_FLAG_NON_PAGED`
                 unsafe {
```
wmmc88: restructure the logic so that you only have one unsafe op per unsafe block. this may also help with overuse of nesting conditionals

crates/wdk-alloc/src/lib.rs
wmmc88: Allowing this to be changed right now also can cause situations where dealloc is called with a different tag from when it was allocated

wmmc88 (changes requested):
Thanks for the PR @Zero-Tang!

How did you go about validating these changes? I think some concrete examples of the non-aligned case should be added the example drivers (ie. sample wdm & sample kmdf)

crates/wdk-alloc/src/lib.rs
```diff
@@ -45,33 +47,65 @@ mod kernel_mode {
...
-                    ExAllocatePool2(POOL_FLAG_NON_PAGED, layout.size() as SIZE_T, RUST_TAG)
+                    if layout.align() > POOL_ALIGNMENT {
```
wmmc88: i think you need to do some validation on the alignment value

crates/wdk-alloc/src/lib.rs
Zero-Tang: In [Remarks section of `ExAllocatePool2`](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-exallocatepool2#remarks), it says:

> If NumberOfBytes is PAGE_SIZE or greater, a page-aligned buffer is allocated. Memory allocations of PAGE_SIZE or less are allocated within a page and do not cross page boundaries. Memory allocations of less than PAGE_SIZE are not necessarily page-aligned but are aligned to 8-byte boundaries in 32-bit systems and to 16-byte boundaries in 64-bit systems.

We might further optimize the space by checking if the alignment is smaller than or equal to `PAGE_SIZE` and size is bigger than `PAGE_SIZE`.

However, documentation didn't say if there's such a constant in WDK, so I just defined that constant here.

crates/wdk-alloc/src/lib.rs
Zero-Tang: It is not documented, but `_aligned_malloc` is actually somewhat open-source. Check the `%ProgramFiles%\Windows Kits\10\Source\<Your SDK version>\ucrt\heap\align.cpp` file, and you can find its implementation.

crates/wdk-alloc/src/lib.rs
Zero-Tang: If you mean whether the `layout.align()` returns a number being power of 2, there is no need. [This method guarantees it would be power of 2](https://doc.rust-lang.org/stable/core/alloc/struct.Layout.html#method.align).

Zero-Tang:
Thanks for the feedback!
I did wrote an example, but in a local separate crate, because rust-analyzer marks most parts of wdk-alloc crate as inactive codes for some strange reasons. It won't be marked as inactive if I check it in a separate crate.
I will copy'n'paste my validation example here soon.

gurry:
> related to #303? @gurry could you also take a look at this and provide feedback :)

Thanks @wmmc88. I will have a look.

crates/wdk-alloc/src/lib.rs
Zero-Tang: What about we put the pool flags and the tags into the `WdkAllocator` struct? It won't have interior mutability but still grants developers to use other options.

crates/wdk-alloc/src/lib.rs
```diff
@@ -46,32 +43,76 @@ mod kernel_mode {
     pub struct WdkAllocator;
 
     // The value of memory tags are stored in little-endian order, so it is
-    // convenient to reverse the order for readability in tooling (ie. Windbg)
+    // convenient to reverse the order for readability in tooling (ie. Windbg).
     const RUST_TAG: ULONG = u32::from_ne_bytes(*b"rust");
+    
+    // The minimum alignment of pointer returned by ExAllocatePool2.
+    const POOL_ALIGNMENT: usize = size_of::<*mut u8>() * 2;
```
gurry:
```suggestion
    const MIN_POOL_ALIGNMENT: usize = size_of::<*mut u8>() * 2;
```

`MIN_` prefix makes the meaning clearer

crates/wdk-alloc/src/lib.rs
```diff
@@ -46,32 +43,76 @@ mod kernel_mode {
...
+
+    #[inline] fn require_realignment(layout: Layout) -> bool {
+        if layout.align() <= POOL_ALIGNMENT {
+            false
+        } else {
+            layout.align() > PAGE_SIZE as usize || layout.size() < PAGE_SIZE as usize
+        }
```
gurry:
```suggestion
        // `ExAllocatePool2` uses an alignment of `PAGE_SIZE` or minimum pool 
        // alignment depending on the requested size.
        // (See documentation: https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-exallocatepool2#remarks
        let chosen_alignment = if layout.size() >= PAGE_SIZE { PAGE_SIZE } else { MIN_POOL_ALIGNMENT };

        // Realignment needed only if requested alignment cannot
        // be satisfied by `ExAllocatePool2`'s chosen alignment
        layout.align() > chosen_alignment
```
This makes the logic easier to understand

crates/wdk-alloc/src/lib.rs
```diff
@@ -46,32 +43,76 @@ mod kernel_mode {
...
-                    ExAllocatePool2(POOL_FLAG_NON_PAGED, layout.size() as SIZE_T, RUST_TAG)
+                if require_realignment(layout) {
+                    // If the requested alignment is too large, we'll have to waste some space as large as the alignment.
+                    let size = layout.size() + layout.align();
+                    let mask = !(layout.align() - 1);
+                    let p = unsafe {
+                        ExAllocatePool2(POOL_FLAG_NON_PAGED, size as SIZE_T, RUST_TAG)
+                    };
+                    if !p.is_null() {
+                        let q = ((p as usize & mask) + layout.align()) as *mut PVOID;
```
gurry, line 96:
```suggestion
                        // Align the pointer up to the first address that meets alignment requirement
                        let q = ((p as usize & mask) + layout.align()) as *mut PVOID;
```

crates/wdk-alloc/src/lib.rs
```diff
@@ -46,32 +43,76 @@ mod kernel_mode {
...
         }
 
-        unsafe fn dealloc(&self, ptr: *mut u8, _layout: Layout) {
+        unsafe fn dealloc(&self, ptr: *mut u8, layout: Layout) {
             // SAFETY: `ExFreePool` is safe to call from any `IRQL` <= `DISPATCH_LEVEL`
             // since its freeing memory allocated from `POOL_FLAG_NON_PAGED` in `alloc`
+            let p = if require_realignment(layout) {
+                // The alignment is too large, the original pointer is stored right before
+                // the ptr. This is also how `_aligned_free` is implemented in msvcrt.dll
+                let q: *mut *mut u8 = ptr.cast();
```
gurry:
```suggestion
                let q: *mut PVOID = ptr.cast();
```

For consistency with the corresponding cast in `alloc()` (line 76)

crates/wdk-alloc/src/lib.rs
```diff
@@ -46,32 +43,76 @@ mod kernel_mode {
...
+                unsafe {
+                    q.sub(1).read()
+                }
+            } else {
+                // The alignment is normal, so the ptr is already the original pointer.
+                ptr.cast()
+            };
             unsafe {
-                ExFreePool(ptr.cast());
+                ExFreePoolWithTag(p.cast(), RUST_TAG);
```
gurry: Do we need this `cast()`? I'm assuming the type will already be `PVOID`.

examples/sample-wdm-driver/src/lib.rs
```diff
@@ -53,7 +55,22 @@ pub unsafe extern "system" fn driver_entry(
...
+        // Verify its alignment.
+        assert_eq!((&raw const *ah) as usize & (align_of::<AlignedHigher>() - 1), 0);
+        // Allocate a vector that occupies more than a page.
+        let vh=vec![AlignedHigher(1234), AlignedHigher(5678), AlignedHigher(9012), AlignedHigher(3456), AlignedHigher(7890)];
+        println!("vh is allocated at {:p}, value={vh:?}",vh.as_ptr());
+        // Verify their alignments.
+        for x in &vh {
+            assert_eq!((core::ptr::from_ref::<AlignedHigher>(x) as usize) & (align_of::<AlignedHigher>() - 1), 0);
+        }
+    }
```
gurry: @wmmc88 Is there a more preferred way to write tests?

How about we write unit tests in the `crates/wdk-alloc/src/lib.rs` file itself. We can mock APIs like `ExAllocatePool2` by putting them in global statics and then then in the test replacing them with mock implementations.

gurry (commented):
@Zero-Tang The overall logic looks good to me. Thanks. Added comments for some small improvements.

crates/wdk-alloc/src/lib.rs
```diff
@@ -46,32 +43,76 @@ mod kernel_mode {
...
+                let q: *mut *mut u8 = ptr.cast();
+                unsafe {
+                    q.sub(1).read()
+                }
+            } else {
+                // The alignment is normal, so the ptr is already the original pointer.
+                ptr.cast()
+            };
             unsafe {
-                ExFreePool(ptr.cast());
```
gurry, line 74: @wmmc88 What is the reason we weren't using `ExFreePoolWithTag` earlier?

crates/wdk-alloc/src/lib.rs
Zero-Tang: I just stumbled upon the [Remarks section of `HeapAlloc`](https://learn.microsoft.com/en-us/windows/win32/api/heapapi/nf-heapapi-heapalloc#remarks). There is a constant called `MEMORY_ALLOCATION_ALIGNMENT` and it exists in `wdk_sys` crate. Should we use this constant instead?

examples/sample-wdm-driver/src/lib.rs
Zero-Tang: I think we can mock `ExAllocatePool2` by calling `VirtualAlloc` on Windows or `mmap` on Linux, then add the returned pointer by 16 or 8. In this case, we can guarantee no coincidental big alignments returned by CRT's `malloc`.

crates/wdk-alloc/src/lib.rs
gurry: > I just stumbled upon the [Remarks section of `HeapAlloc`](https://learn.microsoft.com/en-us/windows/win32/api/heapapi/nf-heapapi-heapalloc#remarks). There is a constant called `MEMORY_ALLOCATION_ALIGNMENT` and it exists in `wdk_sys` crate. Should we use this constant instead?

`HeapAlloc` is unrelated to `ExAllocatePool2` so it would be wrong to use that constant.

examples/sample-wdm-driver/src/lib.rs
gurry: It may be easier for the mock `ExAllocatePool2` to use a global array as its source of allocations instead of `VirtualAlloc` or `malloc`.

crates/wdk-alloc/src/lib.rs
Zero-Tang: Despite them sharing the same description? Well, I suppose it is not wrong to expect such description may change in future.

crates/wdk-alloc/src/lib.rs
```diff
@@ -46,32 +43,79 @@ mod kernel_mode {
...
     // SAFETY: This is safe because the Wdk allocator:
     //         1. can never unwind since it can never panic
     //         2. has implementations of alloc and dealloc that maintain layout
-    //            constraints (FIXME: Alignment of the layout is currenty not
-    //            supported)
+    //            constraints (including size and alignment)
     unsafe impl GlobalAlloc for WdkAllocator {
         unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
             let ptr =
                 // SAFETY: `ExAllocatePool2` is safe to call from any `IRQL` <= `DISPATCH_LEVEL` since its allocating from `POOL_FLAG_NON_PAGED`
```
gurry: Move this safety comment next to the `ExAllocatePool2` call. Since you have more than one such call now, please add it to other calls as well.

crates/wdk-alloc/src/lib.rs
```diff
@@ -46,32 +43,79 @@ mod kernel_mode {
...
+                    let p = unsafe {
+                        ExAllocatePool2(POOL_FLAG_NON_PAGED, size as SIZE_T, RUST_TAG)
+                    };
+                    if !p.is_null() {
+                        // Align the pointer up to the first address that meets alignment requirement.
+                        let q = ((p as usize & mask) + layout.align()) as *mut PVOID;
+                        // Store the original pointer right before the pointer to return,
+                        // so that ExFreePoolWithTag can receive the correct pointer.
+                        // This is also how `_aligned_malloc` is implemented in msvcrt.dll
+                        unsafe {
```
gurry, line 107: Add safety comment

crates/wdk-alloc/src/lib.rs
```diff
@@ -46,32 +43,79 @@ mod kernel_mode {
...
+                            q.sub(1).write(p);
+                        }
+                        q.cast()
+                    } else {
+                        p
+                    }
+                } else {
+                    // Because we know the alignment while in both alloc and dealloc,
+                    // we don't need to waste space if the default alignment is small enough.
+                    unsafe {
```
gurry, line 120: Add safety comment

crates/wdk-alloc/src/lib.rs
```diff
@@ -46,32 +43,79 @@ mod kernel_mode {
...
             if ptr.is_null() {
                 return core::ptr::null_mut();
             }
             ptr.cast()
         }
 
-        unsafe fn dealloc(&self, ptr: *mut u8, _layout: Layout) {
+        unsafe fn dealloc(&self, ptr: *mut u8, layout: Layout) {
             // SAFETY: `ExFreePool` is safe to call from any `IRQL` <= `DISPATCH_LEVEL`
             // since its freeing memory allocated from `POOL_FLAG_NON_PAGED` in `alloc`
```
gurry: Move this safey comment next to the call to `ExFreePoolWithTag` and update it appropriately.

crates/wdk-alloc/src/lib.rs
```diff
@@ -46,32 +43,79 @@ mod kernel_mode {
...
 
     // The value of memory tags are stored in little-endian order, so it is
-    // convenient to reverse the order for readability in tooling (ie. Windbg)
+    // convenient to reverse the order for readability in tooling (ie. Windbg).
     const RUST_TAG: ULONG = u32::from_ne_bytes(*b"rust");
+    
+    // The minimum alignment of pointer returned by ExAllocatePool2.
+    const MIN_POOL_ALIGNMENT: usize = size_of::<*mut u8>() * 2;
+
+    #[inline] fn require_realignment(layout: Layout) -> bool {
```
gurry: The formatting of `#[inline]` is wrong. It should be above the function. Please run `cargo +nightly fmt --all`. See formatting guidelines here: https://github.com/microsoft/windows-drivers-rs/blob/main/CONTRIBUTING.md#quality

examples/sample-wdm-driver/src/lib.rs
```diff
@@ -12,17 +12,19 @@ extern crate alloc;
...
 use wdk_alloc::WdkAllocator;
 use wdk_sys::{ntddk::DbgPrint, DRIVER_OBJECT, NTSTATUS, PCUNICODE_STRING, STATUS_SUCCESS};
 
-#[cfg(not(test))]
 #[global_allocator]
 static GLOBAL_ALLOCATOR: WdkAllocator = WdkAllocator;
 
+// Example of using custom-aligned allocator
+#[derive(Debug)]
+#[repr(C,align(1024))] struct AlignedHigher(u32);
```
gurry: Wrong formatting. Please run `cargo +nightly fmt --all`

gurry: Nit: consider renaming to `PageAligned` as it is clearer than `AlignedHigher`

gurry (commented):
Thanks for making the requested changes @Zero-Tang. Here are a few more suggestions.

crates/wdk-alloc/src/lib.rs
wmmc88: There might be a constant available. worth looking in the c headers to see if one is available

crates/wdk-alloc/src/lib.rs
wmmc88: I don't believe there was one

examples/sample-wdm-driver/src/lib.rs
wmmc88: It would be preferable if there are unit-tests with mocked wdk api calls in the `wdk-alloc`. 

Ideally, integration tests for the allocator would be facilitated in `wdk-alloc/examples` folder, but i don't think either our `cargo-make` scripts, nor the new `cargo-wdk` tool support packaging drivers from `examples` right now

crates/wdk-alloc/src/lib.rs
wmmc88: I'd consider addition of a `with_tag` constructor for `WdkAllocator` in a separate PR, but I would need some concrete examples of how this would be useful in a `global_allocator` context since all allocations will be tied to the same tag

crates/wdk-alloc/src/lib.rs
wmmc88: ah right. `Layout` guarantees non-zero power of 2 alignment. The masking trick to figure out the aligned address explicitly requires power of 2 and non-zero align to work, so it'd be worth writing a comment there clarifying that those invariants are upheld by `Layout`

crates/wdk-alloc/src/lib.rs
```diff
@@ -101,19 +119,24 @@ mod kernel_mode {
...
+                //             aligned start            aligned end
+                // "p" is where we store the original address returned by `ExAllocatePool2`.
+                // So we will pass "p" to ExFreePoolWithTag on `dealloc` method.
+                // "aligned start" is the address we received from `ptr` argument in `dealloc` method.
+                unsafe { q.sub(1).read() }
             } else {
                 // The alignment is normal, so the ptr is already the original pointer.
                 ptr.cast()
             };
+            // SAFETY: `ExFreePool` is safe to call from any `IRQL` <= `DISPATCH_LEVEL`
```
gurry:
```suggestion
            // SAFETY: `ExFreePoolWithTag` is safe to call from any `IRQL` <= `DISPATCH_LEVEL`
```

crates/wdk-alloc/src/lib.rs
```diff
@@ -101,19 +119,24 @@ mod kernel_mode {
...
                 let q: *mut PVOID = ptr.cast();
-                unsafe {
-                    q.sub(1).read()
-                }
+                // SAFETY: We stored the original pointer right before the aligned object. See the following graph:
+                //   | ......... | p | ..................... | ...... |
+                //                   ^                       ^
+                //             aligned start            aligned end
+                // "p" is where we store the original address returned by `ExAllocatePool2`.
+                // So we will pass "p" to ExFreePoolWithTag on `dealloc` method.
```
gurry:
```suggestion
                // So we will pass "p" to `ExFreePoolWithTag` on `dealloc` method.
```

crates/wdk-alloc/src/lib.rs
```diff
@@ -45,52 +50,65 @@ mod kernel_mode {
...
                         let q = ((p as usize & mask) + layout.align()) as *mut PVOID;
-                        // Store the original pointer right before the pointer to return,
-                        // so that ExFreePoolWithTag can receive the correct pointer.
-                        // This is also how `_aligned_malloc` is implemented in msvcrt.dll
+                        // SAFETY: There are sufficient space to store a pointer. See the following graph:
+                        //   | ......... | p | ..................... | ...... |
+                        //                   ^                       ^
+                        //             aligned start            aligned end
+                        // "p" is where we store the original address returned by `ExAllocatePool2`.
+                        // We will also pass "p" to ExFreePoolWithTag on `dealloc` method.
```
gurry:
```suggestion
                        // We will also pass "p" to `ExFreePoolWithTag` on `dealloc` method.
```

gurry:
```suggestion
                        // We will also pass "p" to ExFreePoolWithTag on `dealloc` method.
                        // This is also how `_aligned_free` is implemented in msvcrt.dll
```
I feel we should not lose the comment about `_aligned_free` as it gives useful context.

gurry (commented):
A few comments on comments.

Zero-Tang:
Just FYI that I edited the first comment of this PR thread so that this PR will close #303 on merge.

crates/wdk-alloc/src/lib.rs
Zero-Tang: I'd say the morale of doing so is to help developers debug memory leaks among multiple rust drivers. We should expect developers may use [`Box::leak`](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.leak) or [`Box::into_raw`](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.into_raw) method to extend its lifetime (to probably `'static`) but forgot to use [`Box::from_raw`](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.from_raw) in order to let RAII drop it. If all rust drivers use the same tag, it may confuse the developers if other drivers are actually leaking when they use [!poolused](https://learn.microsoft.com/en-us/windows-hardware/drivers/debuggercmds/-poolused) command in WinDbg.

In any case, I agree that let's do this in another PR.

gurry:
> Just FYI that I edited the first comment of this PR thread so that this PR will close #303 on merge.

Thanks. Yes, #303 should be closed. 

I am done with my review. Will wait for inputs from @wmmc88.

examples/sample-wdm-driver/src/lib.rs
gurry: > It would be preferable if there are unit-tests with mocked wdk api calls in the wdk-alloc.

In that case @Zero-Tang can you please see if you can write some key unit tests?

Zero-Tang: Emmmmmm, I'm not so sure how to do this.

gurry: An approach like this can work:

1. Declare a static pointing either to the real `ExAllocatePool2` or a mock depending on whether we're running tests or not:

```rust
type ExAllocatePool2Fn = unsafe extern "system" fn(pool_flags: u32, number_of_bytes: usize, tag: u32) -> *mut c_void;

#[cfg(not(test))]
pub static mut EX_ALLOCATE_POOL2_FN: ExAllocatePool2Fn = ExAllocatePool2;
#[cfg(test)]
pub static mut EX_ALLOCATE_POOL2_FN: ExAllocatePool2Fn = mock_ex_allocate_pool2;
```

2. Declare a static buffer that `mock_ex_allocate_pool2` will "allocate" from:

```rust
static ALLOC_BUFFER:  [u8; 5120] = [0; 5120]; // 5 Pages worth of buffer to allocate from
static mut FREE_INDEX: usize = 0; // Index to keep track of available memory in buffer
```
3. Implement `mock_ex_allocate_pool2` using this buffer as the source of allocations.

4. In `WdkAllcator` call `EX_ALLOCATE_POOL2_FN` instead of calling `ExAllocatePoo2` directly.

Now you should be ready to write your tests.

Zero-Tang: I kinda give up on this in that I couldn't make the test link to `wdk_alloc::kernel_mode` module due to the `#[cfg(any(driver_model__driver_type = "WDM", driver_model__driver_type = "KMDF"))]` attribute.
I tried `cargo test --config driver_model__driver_type=\"WDM\"` but it doesn't work.

gurry: @wmmc88 Can you please help with this?

wmmc88: > I kinda give up on this in that I couldn't make the test link to wdk_alloc::kernel_mode module due to the #[cfg(any(driver_model__driver_type = "WDM", driver_model__driver_type = "KMDF"))] attribute.
I tried cargo test --config driver_model__driver_type=\"WDM\" but it doesn't work.

I don't think this will work as currently structured. I am working on some tooling to enable better unit testing of code that relies on the metadata section. unfortunately, since this is such a high impact change, I think we'll wait for that tooling to finish before merging this (with tests).

Zero-Tang:
It's been a year since I opened this PR. How's the progress?

PR #356, opened 2025-05-14
Add hidport.h to the hid feature for umdf setup

### LINKED ISSUE #355: HID hidport.h headers missing for UMDF build
According to the examples for HID miniport drivers. It should be possible to implement HID miniport drivers with UMDF builds.

Example here: https://github.com/microsoft/Windows-driver-samples/tree/main/hid/vhidmini2

This requires types from hidport.h that has been marked as only available in kernel mode drivers here: https://github.com/microsoft/windows-drivers-rs/blob/b82b4278f7085cfb0165d7cc6ffd6e0ccaad41ff/crates/wdk-build/src/lib.rs#L753

That limitation should likely be lifted for generation of bindgen headers.

elupus:
Ensure hidport.h is added to Umdf builds.

Fixes #355

### DIFF
--- a/crates/wdk-build/src/lib.rs
+++ b/crates/wdk-build/src/lib.rs
@@ -755,6 +755,21 @@ impl Config {
             headers.extend(["hidpddi.h", "hidport.h", "kbdmou.h", "ntdd8042.h"]);
         }
 
+        if matches!(self.driver_config, DriverConfig::Umdf(_)) {
+            /* the hidport.h file is expected for Umdf, but located in km folder */
+            let include_directory = self.wdk_content_root.join("Include");
+            let sdk_version = utils::get_latest_windows_sdk_version(include_directory.as_path()).expect("sdk version");
+            let windows_sdk_include_path = include_directory.join(sdk_version);
+            let hidports_path = windows_sdk_include_path
+                .join("km")
+                .join("hidport.h")
+                .canonicalize().unwrap()
+                .strip_extended_length_path_prefix().unwrap()
+                .to_string_lossy().to_string().replace("\\", "/");
+
+            headers.push(Box::leak(hidports_path.into_boxed_str()));
+        }
+
         if matches!(self.driver_config, DriverConfig::Kmdf(_)) {
             headers.extend(["HidSpiCx/1.0/hidspicx.h"]);
         }


crates/wdk-build/src/lib.rs
```diff
@@ -755,6 +755,21 @@ impl Config {
...
+            let sdk_version = utils::get_latest_windows_sdk_version(include_directory.as_path()).expect("sdk version");
+            let windows_sdk_include_path = include_directory.join(sdk_version);
+            let hidports_path = windows_sdk_include_path
+                .join("km")
+                .join("hidport.h")
+                .canonicalize().unwrap()
+                .strip_extended_length_path_prefix().unwrap()
+                .to_string_lossy().to_string().replace("\\", "/");
+
+            headers.push(Box::leak(hidports_path.into_boxed_str()));
```
elupus: This is rather ugly. But my rust-fy is oh so limited. These functions likely should return some Cow string or just plain String instead of some static reference, but that feel above my skill lever right now.

crates/wdk-build/src/lib.rs
```diff
@@ -755,6 +755,21 @@ impl Config {
             headers.extend(["hidpddi.h", "hidport.h", "kbdmou.h", "ntdd8042.h"]);
         }
 
+        if matches!(self.driver_config, DriverConfig::Umdf(_)) {
```
wmmc88: Since we now cover all configs, we might as well combine this and the previous matches into a match

something like:

```Rust
match self.driver_config {
    DriverConfig::Wdm | DriverConfig::Kmdf(_) => {
        headers.extend(["hidpddi.h", "hidport.h", "kbdmou.h", "ntdd8042.h"]);
        if let DriverConfig::Kmdf(_) = self.driver_config {
            headers.extend(["HidSpiCx/1.0/hidspicx.h"]);
        }
    }
    DriverConfig::UMDF(_) => {
         // stuff here
    }
}
```

crates/wdk-build/src/lib.rs
wmmc88: I agree that this is ugly. we should not leak memory here. 

Here's a simple example of how Cow would work here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=201ce300262dabda495bc0b36b2bc3e0

You can change the signature of this private function to return Cow<'static, str>

crates/wdk-build/src/lib.rs
```diff
@@ -755,6 +755,21 @@ impl Config {
...
+            /* the hidport.h file is expected for Umdf, but located in km folder */
+            let include_directory = self.wdk_content_root.join("Include");
+            let sdk_version = utils::get_latest_windows_sdk_version(include_directory.as_path()).expect("sdk version");
+            let windows_sdk_include_path = include_directory.join(sdk_version);
+            let hidports_path = windows_sdk_include_path
+                .join("km")
+                .join("hidport.h")
+                .canonicalize().unwrap()
+                .strip_extended_length_path_prefix().unwrap()
```
wmmc88: I know I use `strip_extended_length_path_prefix` elsewhere, but https://doc.rust-lang.org/std/path/fn.absolute.html is sufficient now since its stabilized

crates/wdk-build/src/lib.rs
```diff
@@ -755,6 +755,21 @@ impl Config {
             headers.extend(["hidpddi.h", "hidport.h", "kbdmou.h", "ntdd8042.h"]);
         }
 
+        if matches!(self.driver_config, DriverConfig::Umdf(_)) {
+            /* the hidport.h file is expected for Umdf, but located in km folder */
+            let include_directory = self.wdk_content_root.join("Include");
```
wmmc88: Not something that needs to be addressed in this pr, but a lot of this logic is duplicated in a couple different places now. I think it might be good to have the `wdk_build::Config` constructor create all the "useful paths" and store them in the config struct. Possibly better is to provide public convenience methods to generate those paths. 

For useful paths, I'm thinking things that correspond with the entries defined in `WDK.props`:
image

wmmc88: If you want to help address this in this pr, i'd just narrow to only a `kernel_mode_include_path` that corresponds with `KM_IncludePath`

crates/wdk-build/src/lib.rs
elupus: Wasn't soo simple due to side effects of the match in headers(). Can you help me out here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=0e50c887aa9858043dbfbdb196c354d0 

I can't understand how it can be this hard to get this match statement to end up with compatible types. All the into, into_iters, to_owned is breaking my mind at the moment.

elupus:

crates/wdk-build/src/lib.rs
elupus: Ok. I think i got some solution to this.

Ps. Rust strings are really hard to work with even for somebody with lots of c and c++ background. Feels like so much premature optimization. But maybe I'll learn at some point :).

crates/wdk-build/src/lib.rs
elupus: I've added a helper for this value

elupus:
I think this should be alright now?

elupus:
I know i shouldn't have have accepted copilots suggestion.. Should be fixed now i hope.

elupus:
Ping

PR #360, opened 2025-05-21
fix(docs): added default target to  `metadata.docs.rs` in package manifests

mkatychev:
current [docs.rs](https://docs.rs/crate/wdk-sys/0.4.0/builds/2059469) builds are failing because the package is using linux as a build target by default:
https://docs.rs/about/metadata

Added explicit `default-target` to all published crates

### DIFF
--- a/crates/wdk-alloc/Cargo.toml
+++ b/crates/wdk-alloc/Cargo.toml
@@ -22,3 +22,6 @@ wdk-sys = { workspace = true, features = ["test-stubs"] }
 
 [lints]
 workspace = true
+
+[package.metadata.docs.rs]
+default-target = "x86_64-pc-windows-msvc"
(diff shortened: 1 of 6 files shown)


mkatychev:

crates/wdk-alloc/Cargo.toml
```diff
@@ -22,3 +22,6 @@ wdk-sys = { workspace = true, features = ["test-stubs"] }
 
 [lints]
 workspace = true
+
+[package.metadata.docs.rs]
+default-target = "x86_64-pc-windows-msvc"
```
wmmc88, line 27: can you add a `targets` section two with the arm64 variant

wmmc88 (changes requested):
Thanks for looking into this. Documentation will still be fairly limited since everything will be cfg'ed out. I also don't believe the docs rs containers contain the WDK: https://github.com/rust-lang/docs.rs/issues/2227

Have you been able to test locally via https://docs.rs/about/builds#testing-builds-locally?

mkatychev:
> Thanks for looking into this. Documentation will still be fairly limited since everything will be cfg'ed out. I also don't believe the docs rs containers contain the WDK: [rust-lang/docs.rs#2227](https://github.com/rust-lang/docs.rs/issues/2227)
> 
> Have you been able to test locally via https://docs.rs/about/builds#testing-builds-locally?

So that means that the documentation will be different per machine/WDK version? I'm curious if there's a way to "vendor" the generated code somehow.

PR #406, opened 2025-06-24
feat(ci): install and use `nuget` packages in CI workflows

svasista-ms:
Rework of https://github.com/microsoft/windows-drivers-rs/pull/236

Enables the CI workflows to use NuGet WDK/SDK packages instead of WinGet WDK installation. This allows us to specify multiple WDK versions in the workflow's matrix.

NOTE: The latest WDK version is installed as a NuGet package. Older versions of WDK (before v10.0.26100) still rely on WinGet. On a local computer eWDK environement can be used.

Closes https://github.com/microsoft/windows-drivers-rs/issues/214 partially. NI WDK and LLVM still use WinGet at the moment.

Summary of changes:
1. Added `actions/install-wdk/action.yml` that takes version, source, host and target arch as inputs and install the specific WDK.
2. Modified workflows that have WDK installation as a step to use the new action.
3. Added logic in `wdk-build` to check for the following env vars to set up the WDK version, and the library, include and binary paths required for the build process:
- `WDKContentRoot` - the root directory of the WDK content from which other paths are determined based on the target the driver is being built for.
- `Version_Number` - the Windows SDK Version (10.0.XXXXX[.YYYY])
- `WDKBinRoot` - path to the `bin` directory of the WDK that contains
- `WDKToolRoot` - path to `tools` directory of the Kit
- `WindowsSdkBinPath` - path to the `bin` directory of the SDK 
When `nuget` based WDK is used, the above env vars must be set to the respective paths for the build to succeed. 
The functions in `lib.rs`, `cargo-make.rs` and `utils.rs` were updated accordingly to use these env vars if set or else fallback to default logic (derive everything from `WDKContentRoot`).


.github/actions/install-wdk/action.yml
```diff
@@ -0,0 +1,167 @@
...
+                    Write-Host "Found $($availableVersions.Count) versions using Find-Package"
+                } catch {
+                    Write-Host "Find-Package failed, trying alternative approach..."
+                }
+                
+                # Fallback: try NuGet API directly
+                if ($availableVersions.Count -eq 0) {
+                    Write-Host "Trying NuGet API directly..."
+                    try {
+                        $nugetApiUrl = "https://api.nuget.org/v3-flatcontainer/microsoft.windows.wdk.x64/index.json"
```
krishnakumar4a4: I feel like one of the ways to fetch QFE should be enough. This makes the script simpler. There should be no reason Find-Package would succeed and not the Nuget API and vice versa. 

Personally, I prefer to keep the API approach as the API contracts are less likely to change without changing URL itself.

.github/actions/install-wdk/action.yml
```diff
@@ -0,0 +1,167 @@
...
+                $matchingVersions = $availableVersions | Where-Object { $_.StartsWith("$baseVersion.") }
+                if ($matchingVersions.Count -eq 0) {
+                    Write-Warning "No QFE versions found for base version $baseVersion, trying without QFE..."
+                    $version = $inputVersion
+                } else {
+                    # Sort versions and get the latest
+                    $version = $matchingVersions | Sort-Object { [System.Version]$_ } | Select-Object -Last 1
+                    Write-Host "Found latest QFE version: $version"
+                }
+            } catch {
```
krishnakumar4a4: Shouldn't we exit in this catch clause?

.github/actions/install-wdk/action.yml
```diff
@@ -0,0 +1,167 @@
...
+              Write-Error "Failed to install $_ version $version"
+              exit 1
+          }
+          if (-not (Test-Path $packages)) {
+            Write-Error "$packages path could not be found. Please check the WDK version and source."
+            exit 1
+          }
+        
+          Write-Host "Setting WDKContentRoot environment variable..."
+          Write-Output "WDKContentRoot=$packages\Microsoft.Windows.WDK.$targetArch.$version\c\" >> $env:GITHUB_ENV
```
krishnakumar4a4, line 118: Can you help me understand why only `WDKContentRoot` is based on targetArch?

.github/workflows/docs.yaml
krishnakumar4a4, line 1: Do we need to run docs, lint, codeql workflows with older WDK? 
Is there any direct/indirect relationship of WDK versions with these? 
I understand build, test workflows may be an exception.

.github/actions/install-wdk/action.yml
```diff
@@ -0,0 +1,167 @@
...
+
+          Write-Host "Setting Version_Number environment variable..."
+          Write-Output "Version_Number=$sdkVersion.0" >> $env:GITHUB_ENV
```
krishnakumar4a4, line 121: Stick to PacalCase'ing for env vars.

.github/actions/install-wdk/action.yml
svasista-ms: Noted, thank you.

.github/actions/install-wdk/action.yml
svasista-ms: `WDKContentRoot` is used to derive the `Lib` and `Include` paths that are used to generate the bindings. So, this should match the target arch. But the `bin` and `tools` folders that contain the executables used in the "packaging" phase it has to match the host arch.

.github/actions/install-wdk/action.yml
svasista-ms: This is an EWDK env var, which we are piggybacking off of. So, I did not change the case. Using eWDK vars allow us to support eWDK based builds as well.

.github/workflows/docs.yaml
svasista-ms: Yes, I think we should use older WDK at least for codeql and lint workflows, but not sure about the docs workflow. 
`codeql` builds the `examples` with `--all-features` flag, so different kits have different versions of the headers available.

gurry (dismissed):

crates/wdk-build/src/lib.rs
```diff
@@ -1444,8 +1444,6 @@ mod tests {
     use std::assert_matches::assert_matches;
     use std::{collections::HashMap, ffi::OsStr, sync::Mutex};
 
-    use assert_fs::TempDir;
-
     use super::*;
```
wmmc88, line 1447: why the removal of the tests here @svasista-ms ?

gurry: They were failing and it was bit of a hassle to fix them e.g. having to set environment variables etc. Since this code is only temporary and likely to be superseded by the system dependency install feature, we felt it's not worth the effort and we can come back to it later. 

We still have the unit tests in place for the core algorithm for finding the UCX folder: https://github.com/microsoft/windows-drivers-rs/blob/e6ea26d932e691457ff70ca652c662a467871684/crates/wdk-build/src/utils.rs#L811 So there's an adequate level of validation still.

svasista-ms: Thank you for explaining @gurry. The test was failing when `Version_Number` env var is set to `10.0.26100.0` because the version in the test is hardcoded to `10.0.22621.0`. Instead of using env vars in unit tests we decided to remove it for now and add it back once sys dep feature's design is complete. As Gurinder has mentioned, the `find_max_version_in_directory` fn is tested. Along with this, the `codeql` workflow tests building with `all-features`. Unit tests can be added when `Config` struct is expanded/modified for the sys dep feature.

.github/actions/install-wdk/action.yml
krishnakumar4a4: Got it, this helps my understanding.

krishnakumar4a4 (dismissed):

crates/wdk-build/src/cargo_make.rs
```diff
@@ -514,65 +511,72 @@ pub fn validate_command_line_args() -> impl IntoIterator<Item = String> {
...
-            .expect("arch_specific_wdk_tool_root should only contain valid UTF8"),
-    );
+        .strip_extended_length_path_prefix()?
+        .to_str()
+        .expect("WDK tool path should be valid UTF-8")
+        .to_string();
+    prepend_to_semicolon_delimited_env_var(PATH_ENV_VAR, host_windows_sdk_version_tool_path);
 
-    Ok([PATH_ENV_VAR].map(std::string::ToString::to_string))
+    Ok([PATH_ENV_VAR].map(string::ToString::to_string))
```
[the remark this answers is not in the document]
gurry: Better to import `std::string::ToString` and then use `ToString::to_string` here instead of what Copilot is suggesting.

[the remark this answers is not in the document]
svasista-ms: Oops! I already made the change :( 
Shall I revert it?

crates/wdk-build/src/cargo_make.rs
```diff
@@ -564,7 +563,7 @@ pub fn setup_path() -> Result<impl IntoIterator<Item = String>, ConfigError> {
         .to_string();
     prepend_to_semicolon_delimited_env_var(PATH_ENV_VAR, host_windows_sdk_version_tool_path);
 
-    Ok([PATH_ENV_VAR].map(string::ToString::to_string))
+    Ok([PATH_ENV_VAR].map(std::string::ToString::to_string))
```
gurry: @svasista-ms This is inconstent with line 496. We should use `ToString::to_string` at both places.

PR #411, opened 2025-07-02
fix: use latest version of ucx in the WDKContent as default

svasista-ms:
When tested with WDK Version 22621, the following bug was found - 

`--- stderr
  D:\a\windows-drivers-rs\windows-drivers-rs\crates\wdk-sys\types-input.h:53:10: fatal error: 'ucx/1.6/ucxclass.h' file not found`

https://github.com/microsoft/windows-drivers-rs/actions/runs/15851446771/job/44685702556#step:9:1631

Instead of using version `1.6`, the latest version of `ucx` available in the `$WDKContentRoot\Include\10.0.26100.0\km\ucx` directory is used.

image

Summary of changes,

1. `ucx_header_path` function in `impl Config` that determines the latest version of UCX header in the directory.
2. `utils::find_max_version_in_directory` function to find the max version in a directory that contains folders of the form `x.y` where x and y are valid `u32`s.
3. Extracted `windows_sdk_library_path` construction into to `arch_sdk_library_path` for reusability.
4. Added basic tests

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -60,19 +60,19 @@ jobs:
 
     - name: Install WDK (${{ matrix.wdk }})
       run: |
-        # Delete the Ge WDK (10.0.26100) Include and Lib folders if present on the system.
-        # Ni WDK (10.0.22621) will be the latest available Kit on the system.
-        # FIXME: Remove once issues with Ge WDK are resolved.
-        $wdkGePaths = @(
+        # Delete the latest WDK (10.0.26100) Include and Lib folders if present on the system.
+        # Older WDK (10.0.22621) will be the latest available Kit on the system once installed.
+        # FIXME: Remove once issues with using latest WDK are resolved.
+        $latestWdkPaths = @(
           "C:\Program Files (x86)\Windows Kits\10\Include\10.0.26100.0",
           "C:\Program Files (x86)\Windows Kits\10\Lib\10.0.26100.0"
         )
-        foreach ($path in $wdkGePaths) {
+        foreach ($path in $latestWdkPaths) {
           if (Test-Path $path) {
             Write-Host "Deleting folder: $path"
             Remove-Item -Path $path -Recurse -Force -ErrorAction SilentlyContinue
           } else {
-            Write-Host "Ge WDK folder not found, skipping deletion: $path"
+            Write-Host "WDK folder not found, skipping deletion: $path"
           }
         }
         
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -60,12 +60,35 @@ jobs:
 
     - name: Install WDK (${{ matrix.wdk }})
       run: |
-        if ((Get-WinGetPackage -Id ${{ matrix.wdk }} -Source winget -MatchOption Equals).Id -eq '${{ matrix.wdk }}') {
-          Write-Host "${{ matrix.wdk }} is already installed. Attempting to update..."
-          Update-WinGetPackage -Id ${{ matrix.wdk }} -Source winget -MatchOption Equals -Mode Silent -Force
-        } else {
-          Write-Host "Installing ${{ matrix.wdk }}..."
-          Install-WinGetPackage -Id ${{ matrix.wdk }} -Source winget -MatchOption Equals -Mode Silent -Force
+        # Delete the Ge WDK (10.0.26100) Include and Lib folders if present on the system.
+        # Ni WDK (10.0.22621) will be the latest available Kit on the system.
+        # FIXME: Remove once issues with Ge WDK are resolved.
+        $wdkGePaths = @(
+          "C:\Program Files (x86)\Windows Kits\10\Include\10.0.26100.0",
+          "C:\Program Files (x86)\Windows Kits\10\Lib\10.0.26100.0"
+        )
+        foreach ($path in $wdkGePaths) {
+          if (Test-Path $path) {
+            Write-Host "Deleting folder: $path"
+            Remove-Item -Path $path -Recurse -Force -ErrorAction SilentlyContinue
+          } else {
+            Write-Host "Ge WDK folder not found, skipping deletion: $path"
+          }
+        }
+        
+        # Install or update the target WDK
+        try {
+          $targetWdk = Get-WinGetPackage -Id ${{ matrix.wdk }} -Source winget -MatchOption Equals -ErrorAction SilentlyContinue
+          if ($targetWdk) {
+            Write-Host "${{ matrix.wdk }} is already installed. Attempting to update..."
+            Update-WinGetPackage -Id ${{ matrix.wdk }} -Source winget -MatchOption Equals -Mode Silent -Force
+          } else {
+            Write-Host "Installing ${{ matrix.wdk }}..."
+            Install-WinGetPackage -Id ${{ matrix.wdk }} -Source winget -MatchOption Equals -Mode Silent -Force
+          }
+        } catch {
+          Write-Host "Error installing ${{ matrix.wdk }}: $($_.Exception.Message)"
+          exit 1
         }
 
     - name: Install Rust Toolchain (${{ matrix.rust_toolchain }})
--- a/crates/wdk-build/src/lib.rs
+++ b/crates/wdk-build/src/lib.rs
@@ -11,7 +11,7 @@
 //! models (WDM, KMDF, UMDF).
 
 #![cfg_attr(nightly_toolchain, feature(assert_matches))]
-use std::{fmt, fmt::Write, str::FromStr};
+use std::{fmt, str::FromStr};
 
 pub use bindgen::BuilderExt;
 use metadata::TryFromCargoMetadataError;
@@ -243,15 +243,6 @@ rustflags = [\"-C\", \"target-feature=+crt-static\"]
     /// [`metadata::Wdk`]
     #[error(transparent)]
     SerdeError(#[from] metadata::Error),
-
-    /// Error returned when constructing header contents for `bindgen` fails in
-    /// `bindgen_header_contents` function.
-    #[error("failed to construct header contents")]
-    BindgenHeaderContents(
-        /// [`std::fmt::Error`] that caused formatting to fail
-        #[source]
-        std::fmt::Error,
-    ),
 
     /// Error returned when the UCX header file is not found
     #[error("failed to find ucx header file")]
@@ -956,20 +947,18 @@ impl Config {
     ///
     /// # Errors
     /// [`ConfigError`] - if the headers for a [`ApiSubset`] could not be
-    /// determined, or if formatting the header contents fails
+    /// determined
     pub fn bindgen_header_contents(
         &self,
         api_subsets: impl IntoIterator<Item = ApiSubset>,
     ) -> Result<String, ConfigError> {
-        api_subsets
+        Ok(api_subsets
             .into_iter()
-            .try_fold(String::new(), |mut acc, api_subset| {
-                for header in self.headers(api_subset)? {
-                    writeln!(acc, "#include \"{header}\"")
-                        .map_err(ConfigError::BindgenHeaderContents)?;
-                }
-                Ok(acc)
-            })
+            .map(|api_subset| self.headers(api_subset))
+            .collect::<Result<Vec<_>, _>>()?
+            .into_iter()
+            .flat_map(|iter| iter.map(|header| format!("#include \"{header}\"\n")))
+            .collect())
     }
 
     /// Configure a Cargo build of a library that depends on the WDK. This
--- a/crates/wdk-build/src/lib.rs
+++ b/crates/wdk-build/src/lib.rs
@@ -1165,7 +1165,7 @@ impl Config {
                     directory: ucx_header_root_dir.to_string_lossy().into(),
                 }
             })?;
-        let path = format!("ucx/{}.{}/ucxclass.h", max_version.major, max_version.minor);
+        let path = format!("ucx/{}.{}/ucxclass.h", max_version.0, max_version.1);
         Ok(Box::leak(path.into_boxed_str()))
     }
 }
--- a/crates/wdk-build/src/lib.rs
+++ b/crates/wdk-build/src/lib.rs
@@ -1160,7 +1160,7 @@ impl Config {
             utils::get_latest_windows_sdk_version(&self.wdk_content_root.join("Lib"))?;
         let ucx_header_root_dir = self.sdk_library_path(sdk_version)?.join("ucx");
         let max_version =
-            utils::find_max_version_in_directory(&ucx_header_root_dir).ok_or_else(|| {
+            utils::find_max_version_in_directory(&ucx_header_root_dir).map_err(|_| {
                 ConfigError::DirectoryNotFound {
                     directory: ucx_header_root_dir.to_string_lossy().into(),
                 }
--- a/crates/wdk-build/src/lib.rs
+++ b/crates/wdk-build/src/lib.rs
@@ -519,29 +519,13 @@ impl Config {
     /// exist.
     pub fn library_paths(&self) -> Result<impl Iterator<Item = PathBuf>, ConfigError> {
         let mut library_paths = vec![];
-
         let library_directory = self.wdk_content_root.join("Lib");
 
         // Add windows sdk library paths
         // Based off of logic from WindowsDriver.KernelMode.props &
         // WindowsDriver.UserMode.props in NI(22H2) WDK
         let sdk_version = utils::get_latest_windows_sdk_version(library_directory.as_path())?;
-        let windows_sdk_library_path =
-            library_directory
-                .join(sdk_version)
-                .join(match self.driver_config {
-                    DriverConfig::Wdm | DriverConfig::Kmdf(_) => {
-                        format!("km/{}", self.cpu_architecture.as_windows_str(),)
-                    }
-                    DriverConfig::Umdf(_) => {
-                        format!("um/{}", self.cpu_architecture.as_windows_str(),)
-                    }
-                });
-        if !windows_sdk_library_path.is_dir() {
-            return Err(ConfigError::DirectoryNotFound {
-                directory: windows_sdk_library_path.to_string_lossy().into(),
-            });
-        }
+        let windows_sdk_library_path = self.arch_sdk_library_path(sdk_version)?;
         library_paths.push(
             windows_sdk_library_path
                 .canonicalize()?
@@ -869,11 +853,18 @@ impl Config {
         }
 
         if matches!(self.driver_config, DriverConfig::Kmdf(_)) {
+            let ucx_header_path = self
+                .latest_ucx_header_path()
+                .map_err(|e| {
+                    tracing::error!("Failed to get latest UCX header: {e}");
+                    e
+                })
+                .expect("Failed to get UCX header. Please verify WDK installation");
             headers.extend([
                 "ucm/1.0/UcmCx.h",
                 "UcmTcpci/1.0/UcmTcpciCx.h",
                 "UcmUcsi/1.0/UcmucsiCx.h",
-                "ucx/1.6/ucxclass.h",
+                ucx_header_path,
                 "ude/1.1/UdeCx.h",
                 "ufx/1.1/ufxbase.h",
                 "ufxproprietarycharger.h",
@@ -1119,6 +1110,70 @@ impl Config {
             .expect("CARGO_CFG_TARGET_FEATURE should be set by Cargo");
 
         enabled_cpu_target_features.contains(STATICALLY_LINKED_C_RUNTIME_FEATURE_NAME)
+    }
+
+    /// Constructs the architecture-specific Windows SDK library path using the
+    /// provided SDK Version.
+    ///
+    /// Builds the library path following the Windows SDK convention:
+    /// `{library_directory}/{sdk_version}/{km|um}/{architecture}/`
+    ///
+    /// # Arguments
+    ///
+    /// * `sdk_version` - Windows SDK version string (e.g., "10.0.22621.0")
+    ///
+    /// # Returns
+    ///
+    /// The constructed library path if it exists, otherwise
+    /// `ConfigError::DirectoryNotFound`.
+    ///
+    /// # Examples
+    ///
+    /// KMDF/AMD64: `C:\...\Lib\10.0.22621.0\km\x64`
+    /// UMDF/ARM64: `C:\...\Lib\10.0.22621.0\um\arm64`
+    fn arch_sdk_library_path(&self, sdk_version: String) -> Result<PathBuf, ConfigError> {
+        let windows_sdk_library_path =
+            self.wdk_content_root
+                .join("Lib")
+                .join(sdk_version)
+                .join(match self.driver_config {
+                    DriverConfig::Wdm | DriverConfig::Kmdf(_) => {
+                        format!("km/{}", self.cpu_architecture.as_windows_str(),)
+                    }
+                    DriverConfig::Umdf(_) => {
+                        format!("um/{}", self.cpu_architecture.as_windows_str(),)
+                    }
+                });
+        if !windows_sdk_library_path.is_dir() {
+            return Err(ConfigError::DirectoryNotFound {
+                directory: windows_sdk_library_path.to_string_lossy().into(),
+            });
+        }
+        Ok(windows_sdk_library_path)
+    }
+
+    fn latest_ucx_header_path(&self) -> Result<&'static str, ConfigError> {
+        let sdk_version =
+            utils::get_latest_windows_sdk_version(&self.wdk_content_root.join("Lib"))?;
+        let ucx_header_root_dir = self
+            .arch_sdk_library_path(sdk_version)?
+            .join("ucx")
+            .canonicalize()?
+            .strip_extended_length_path_prefix()?;
+
+        // Find the latest version of UCX headers in the directory
+        let max_version =
+            utils::find_max_version_in_directory(&ucx_header_root_dir).ok_or_else(|| {
+                ConfigError::DirectoryNotFound {
+                    directory: format!(
+                        "No UCX version directories found in {}",
+                        ucx_header_root_dir.display()
+                    ),
+                }
+            })?;
+
+        let path = format!("ucx/{}.{}/ucxclass.h", max_version.0, max_version.1);
+        Ok(Box::leak(path.into_boxed_str()))
     }
 }
 
--- a/crates/wdk-build/src/lib.rs
+++ b/crates/wdk-build/src/lib.rs
@@ -1165,10 +1165,7 @@ impl Config {
         let max_version =
             utils::find_max_version_in_directory(&ucx_header_root_dir).ok_or_else(|| {
                 ConfigError::DirectoryNotFound {
-                    directory: format!(
-                        "No UCX version directories found in {}",
-                        ucx_header_root_dir.display()
-                    ),
+                    directory: ucx_header_root_dir.to_string_lossy().into(),
                 }
             })?;
 
--- a/crates/wdk-build/src/lib.rs
+++ b/crates/wdk-build/src/lib.rs
@@ -1179,7 +1179,6 @@ impl Config {
 
     /// Returns the path to the latest available UCX header file present in the
     /// Lib folder of the WDK content root
-    // TODO: SDK and UCX version should be configurable
     fn ucx_header(&self) -> Result<String, ConfigError> {
         let sdk_version =
             utils::get_latest_windows_sdk_version(&self.wdk_content_root.join("Lib"))?;
@@ -1710,8 +1709,17 @@ mod tests {
             // WDM: kernel-mode USB headers but no WDF or UCX
             let required_headers = ["usbbusif.h", "usbdlib.h", "usbfnattach.h", "usbfnioctl.h"];
 
-            let excluded_headers = ["wdfusb.h"];
-            let excluded_header_paths = ["ucx/", "ucm/"];
+            let excluded_headers = [
+                "wdfusb.h",
+                "ucm/1.0/UcmCx.h",
+                "UcmTcpci/1.0/UcmTcpciCx.h",
+                "UcmUcsi/1.0/UcmucsiCx.h",
+                "ude/1.1/UdeCx.h",
+                "ufx/1.1/ufxbase.h",
+                "ufxproprietarycharger.h",
+                "urs/1.0/UrsCx.h",
+                "ucx/1.2/ucxclass.h",
+            ];
 
             for header in required_headers {
                 assert!(
@@ -1724,13 +1732,6 @@ mod tests {
                 assert!(
                     !header_contents.contains(&format!("#include \"{header}\"")),
                     "{header} should not be included"
-                );
-            }
-
-            for path in excluded_header_paths {
-                assert!(
-                    !header_contents.contains(path),
-                    "{path} headers should not be included"
                 );
             }
         }
@@ -1775,7 +1776,16 @@ mod tests {
                 ["usbbusif.h", "usbdlib.h", "usbfnattach.h", "usbfnioctl.h"];
 
             let required_headers = ["wdfusb.h"];
-            let excluded_header_paths = ["ucx/", "ucm/", "ude/"];
+            let excluded_kmdf_headers = [
+                "ucm/1.0/UcmCx.h",
+                "UcmTcpci/1.0/UcmTcpciCx.h",
+                "UcmUcsi/1.0/UcmucsiCx.h",
+                "ude/1.1/UdeCx.h",
+                "ufx/1.1/ufxbase.h",
+                "ufxproprietarycharger.h",
+                "urs/1.0/UrsCx.h",
+                "ucx/1.2/ucxclass.h",
+            ];
 
             for header in excluded_kernel_headers {
                 assert!(
@@ -1791,25 +1801,21 @@ mod tests {
                 );
             }
 
-            for path in excluded_header_paths {
+            for header in excluded_kmdf_headers {
                 assert!(
-                    !header_contents.contains(path),
-                    "{path} headers should not be included"
+                    !header_contents.contains(&format!("#include \"{header}\"")),
+                    "{header} should not be included"
                 );
             }
         }
 
         /// Tests USB header generation for a specific driver configuration
         fn test_usb_headers_with(driver_config: &DriverConfig, cpu_architecture: CpuArchitecture) {
-            // Create a temporary directory structure that mimics a WDK installation
+            // Set up mock WDK directory structure
             let temp_dir = TempDir::new().unwrap();
             let wdk_content_root = temp_dir.path().to_path_buf();
-
-            // Set up mock WDK directory structure
             let lib_dir = wdk_content_root.join("Lib").join("10.0.22621.0");
             let include_dir = wdk_content_root.join("Include").join("10.0.22621.0");
-
-            // Create the necessary directories based on driver config
             match &driver_config {
                 DriverConfig::Wdm | DriverConfig::Kmdf(_) => {
                     std::fs::create_dir_all(
@@ -1841,10 +1847,8 @@ mod tests {
             assert!(result.is_ok(), "Expected USB test to succeed");
             let header_contents = result.unwrap();
 
-            // All driver configs should include common USB headers
             assert_common_usb_headers_present(&header_contents);
 
-            // Verify driver-specific logic branches
             match &driver_config {
                 DriverConfig::Wdm => assert_wdm_usb_headers(&header_contents),
                 DriverConfig::Kmdf(_) => assert_kmdf_usb_headers(&header_contents),
@@ -1854,7 +1858,7 @@ mod tests {
 
         #[test]
         fn usb_success() {
-            // Test cases covering usb_headers() logic branches:
+            // Test cases covering branches in usb_headers():
             // 1. WDM - Common headers + kernel-mode headers, no WDF/UCX
             // 2. KMDF - All headers including WDF, UCX, and KMDF-specific USB
             // 3. UMDF - Common headers + WDF, no kernel-mode or UCX
--- a/crates/wdk-build/src/lib.rs
+++ b/crates/wdk-build/src/lib.rs
@@ -853,8 +853,8 @@ impl Config {
         }
 
         if matches!(self.driver_config, DriverConfig::Kmdf(_)) {
-            let ucx_header_path = self
-                .latest_ucx_header_path()
+            let latest_ucx_header_path = self
+                .ucx_header_path()
                 .map_err(|e| {
                     tracing::error!("Failed to get latest UCX header: {e}");
                     e
@@ -864,7 +864,7 @@ impl Config {
                 "ucm/1.0/UcmCx.h",
                 "UcmTcpci/1.0/UcmTcpciCx.h",
                 "UcmUcsi/1.0/UcmucsiCx.h",
-                ucx_header_path,
+                latest_ucx_header_path,
                 "ude/1.1/UdeCx.h",
                 "ufx/1.1/ufxbase.h",
                 "ufxproprietarycharger.h",
@@ -1113,7 +1113,7 @@ impl Config {
     }
 
     /// Constructs the architecture-specific Windows SDK library path using the
-    /// provided SDK Version.
+    /// provided SDK Version and the driver configuration.
     ///
     /// Builds the library path following the Windows SDK convention:
     /// `{library_directory}/{sdk_version}/{km|um}/{architecture}/`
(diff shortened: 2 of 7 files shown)


crates/wdk-build/src/lib.rs
```diff
@@ -869,11 +853,18 @@ impl Config {
         }
 
         if matches!(self.driver_config, DriverConfig::Kmdf(_)) {
+            let ucx_header_path = self
+                .latest_ucx_header_path()
+                .map_err(|e| {
+                    tracing::error!("Failed to get latest UCX header: {e}");
+                    e
+                })
+                .expect("Failed to get UCX header. Please verify WDK installation");
```
krishnakumar4a4: You should use `?` here instead of using `.expect` so that the ConfigError is propagated up. The ConfigError already has enough details about ucx directory being not found.

crates/wdk-build/src/lib.rs
```diff
@@ -1120,6 +1111,70 @@ impl Config {
...
+            .arch_sdk_library_path(sdk_version)?
+            .join("ucx")
+            .canonicalize()?
+            .strip_extended_length_path_prefix()?;
+
+        // Find the latest version of UCX headers in the directory
+        let max_version =
+            utils::find_max_version_in_directory(&ucx_header_root_dir).ok_or_else(|| {
+                ConfigError::DirectoryNotFound {
+                    directory: format!(
```
krishnakumar4a4: As per the name of this field and per existing convention on how it is set in other places, this should just be set to `ucx_header_root_dir.display()` without any extra string formatting.

crates/wdk-build/src/lib.rs
svasista-ms: The `?` operator can only be used in a method that returns `Result` or `Option`. `usb_headers` returns a `Vec<&'static str>`, so I could not use `?` here.

crates/wdk-build/src/utils.rs
```diff
@@ -366,6 +366,43 @@ fn read_registry_key_string_value(
...
+/// This function will return `Some((1, 6))`
+pub fn find_max_version_in_directory<P: AsRef<Path>>(directory_path: P) -> Option<(u32, u32)> {
+    std::fs::read_dir(directory_path.as_ref())
+        .ok()?
+        .flatten()
+        .filter_map(|entry| {
+            let dir_name = entry.file_name();
+            let dir_name = dir_name.to_str()?;
+            let (major_str, minor_str) = dir_name.split_once('.')?;
+            let major = major_str.parse::<u32>().ok()?;
```
wmmc88: I think we already pull in the semver crate. Would it be better to just parse it as a https://docs.rs/semver/latest/semver/struct.Version.html?

crates/wdk-build/src/lib.rs
```diff
@@ -869,11 +853,18 @@ impl Config {
...
         if matches!(self.driver_config, DriverConfig::Kmdf(_)) {
+            let ucx_header_path = self
+                .latest_ucx_header_path()
+                .map_err(|e| {
+                    tracing::error!("Failed to get latest UCX header: {e}");
+                    e
+                })
+                .expect("Failed to get UCX header. Please verify WDK installation");
             headers.extend([
                 "ucm/1.0/UcmCx.h",
```
wmmc88: Not to be addressed in this pr, but we should track making all of these use the latest in the short term, and be metadata configurable in the medium term

crates/wdk-build/src/utils.rs
```diff
@@ -366,6 +366,43 @@ fn read_registry_key_string_value(
...
+/// # Examples
+/// Given a directory structure:
+/// ```text
+/// /some/path/
+/// ├── 1.4/
+/// ├── 1.5/
+/// └── 1.6/
+/// ```
+/// This function will return `Some((1, 6))`
+pub fn find_max_version_in_directory<P: AsRef<Path>>(directory_path: P) -> Option<(u32, u32)> {
```
wmmc88: let's default to keeping functions private

crates/wdk-build/src/lib.rs
```diff
@@ -1120,6 +1111,67 @@ impl Config {
...
+            return Err(ConfigError::DirectoryNotFound {
+                directory: windows_sdk_library_path.to_string_lossy().into(),
+            });
+        }
+        Ok(windows_sdk_library_path)
+    }
+
+    fn latest_ucx_header_path(&self) -> Result<&'static str, ConfigError> {
+        let sdk_version =
+            utils::get_latest_windows_sdk_version(&self.wdk_content_root.join("Lib"))?;
```
wmmc88, line 1184: nit: if there's a lot of these base paths to construct, we can maybe improve this to construct it once in `new` and reuse it as a struct member

crates/wdk-build/src/lib.rs
svasista-ms: Sure, we discussed this internally as well and decided to address them in subsequent PRs.
Created #412

crates/wdk-build/src/utils.rs
svasista-ms: This function is used in lib.rs. So, had to make it `pub`. It can be `pub(crate)` as it's not being used outside `wdk-build`

crates/wdk-build/src/utils.rs
svasista-ms: Parsing `x.y` to a `semver::Version` may fail, unless we add a dummy patch version.
From the docs:
> 1.0 — too few numeric components. A SemVer version must have exactly three. If you are looking at something that has fewer than three numbers in it, it’s possible it is a VersionReq instead (with an implicit default ^ comparison operator).

crates/wdk-build/src/lib.rs
```diff
@@ -1120,6 +1111,67 @@ impl Config {
...
+    /// # Returns
+    ///
+    /// The constructed library path if it exists, otherwise
+    /// `ConfigError::DirectoryNotFound`.
+    ///
+    /// # Examples
+    ///
+    /// KMDF/AMD64: `C:\...\Lib\10.0.22621.0\km\x64`
+    /// UMDF/ARM64: `C:\...\Lib\10.0.22621.0\um\arm64`
+    fn arch_sdk_library_path(&self, sdk_version: String) -> Result<PathBuf, ConfigError> {
```
gurry: Nit: `arch_` prefix seems unnecessary. How about calling it just `sdk_library_path`?

svasista-ms: This method actually constructs the path based on the `arch` detected by the build script (`wdk-build`). I added `arch` to denote the same. `sdk_library_path` would be good if path is until `C:\...\Lib\10.0.22621.0\`. Another name I had in mind was `arch_specific_sdk_lib_path`. I have created #413 to improve the way `wdk-build` manages paths. We can revisit the names as part of this issue. Please let me know if this works for you.

gurry: The fact that the paths are architecture-specific is implicit and self-evident. I don't think it needs to be included in the name. `sdk_library_path` is clear enough.

wmmc88:
> The "use the UCX header from the latest version" logic implemented here is only a temporary fix right? I assume we're going to make it smart enough later to pick the right header version according to the chosen kit version, especially when we have the relevant enhancements for this in cargo-wdk. Is that understanding correct?

It's tracked in https://github.com/microsoft/windows-drivers-rs/issues/412. The current thinking is default it to latest, but configurable in metadata in the future

crates/wdk-build/src/utils.rs
```diff
@@ -515,4 +552,24 @@ mod tests {
             )
         );
     }
+
+    #[test]
+    fn test_find_max_version_in_directory() {
+        use assert_fs::prelude::*;
```
wmmc88: nit: you can just import the prelude at the `tests` module level

crates/wdk-build/src/utils.rs
wmmc88: yes, pub(crate) is fine. I just want to limit what is exposed outside of the crate in case. Once its exposed, we want to limit the number of breaking changes to it as it causes semver major bump

crates/wdk-build/src/utils.rs
```diff
@@ -515,4 +552,24 @@ mod tests {
...
+        let temp_dir = assert_fs::TempDir::new().unwrap();
+        temp_dir.child("1.2").create_dir_all().unwrap();
+        temp_dir.child("1.10").create_dir_all().unwrap();
+        temp_dir.child("2.0").create_dir_all().unwrap();
+        temp_dir.child("not_a_version").create_dir_all().unwrap();
+        assert_eq!(find_max_version_in_directory(temp_dir.path()), Some((2, 0)));
```
wmmc88: you should add more test cases here. the one here only tests major version parsing working correctly

crates/wdk-build/src/utils.rs
```diff
@@ -515,4 +552,24 @@ mod tests {
             )
         );
     }
+
```
wmmc88, line 810: add test for failed parsing of u32

crates/wdk-build/src/lib.rs
```diff
@@ -1120,6 +1111,67 @@ impl Config {
...
+    }
+
+    fn latest_ucx_header_path(&self) -> Result<&'static str, ConfigError> {
+        let sdk_version =
+            utils::get_latest_windows_sdk_version(&self.wdk_content_root.join("Lib"))?;
+        let ucx_header_root_dir = self
+            .arch_sdk_library_path(sdk_version)?
+            .join("ucx")
+            .canonicalize()?
+            .strip_extended_length_path_prefix()?;
```
wmmc88: `strip_extended_length_path_prefix` is no longer needed since 1.79 and shouldnt be used in any new code. We should just use https://doc.rust-lang.org/std/path/fn.absolute.html now

crates/wdk-build/src/utils.rs
wmmc88: We should probably use a newtype here instead of a tuple for better type safety

crates/wdk-build/src/lib.rs
krishnakumar4a4: I see—I overlooked the function definition here. Yes, `usb_headers` currently doesn't return a Result.
However, I believe we should update these methods (usb_headers, base_headers, wdf_headers, etc.) to return a Result instead of panicking internally with `.expect`.
We should propagate the error up the chain (`main → bindgen_headers → headers → usb_headers`) and handle it appropriately at the top level.

gurry:
> It's tracked in #412. The current thinking is default it to latest, but configurable in metadata in the future

Got it. Thanks.

crates/wdk-build/src/lib.rs
svasista-ms: Created #413

crates/wdk-build/src/utils.rs
svasista-ms: I added more test cases for this function and the new type `BasicVersion`. Please take a look. Thanks

crates/wdk-build/src/utils.rs
```diff
@@ -26,6 +26,30 @@ use windows::{
 
 use crate::{ConfigError, CpuArchitecture};
 
+/// Type for versions of the format "major.minor"
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct BasicVersion {
```
gurry: Consider renaming it to `TwoPartVersion`. The word "Basic"  in `BasicVersion` does not convey any meaning.

crates/wdk-build/src/utils.rs
```diff
@@ -26,6 +26,30 @@ use windows::{
...
+    pub major: u32,
+    /// The minor version number
+    pub minor: u32,
+}
+
+impl BasicVersion {
+    /// Create a new BasicVersion from major and minor components
+    pub fn new(major: u32, minor: u32) -> Self {
+        Self { major, minor }
+    }
```
gurry: This constructor is not adding any value. We can remove it.

crates/wdk-build/src/utils.rs
```diff
@@ -26,6 +26,30 @@ use windows::{
...
 use crate::{ConfigError, CpuArchitecture};
 
+/// Type for versions of the format "major.minor"
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct BasicVersion {
+    /// The major version number
+    pub major: u32,
+    /// The minor version number
+    pub minor: u32,
+}
```
gurry: If this type is used only in this module remove `pub` from it and its fields

crates/wdk-build/src/utils.rs
```diff
@@ -374,39 +398,154 @@ fn read_registry_key_string_value(
...
+
+        #[test]
+        fn test_new() {
+            let version = BasicVersion::new(1, 2);
+            assert_eq!(version.major, 1);
+            assert_eq!(version.minor, 2);
+        }
+
+        #[test]
+        fn test_parse_valid_versions() {
```
gurry: No need to have the prefix `test_` in every test name. It does add anything that we don't already know.

crates/wdk-build/src/utils.rs
```diff
@@ -374,39 +398,154 @@ fn read_registry_key_string_value(
...
+            // Negative numbers
+            assert_eq!(BasicVersion::parse("-1.2"), None);
+            assert_eq!(BasicVersion::parse("1.-2"), None);
+            assert_eq!(BasicVersion::parse("-1.-2"), None);
+
+            // Whitespace
+            assert_eq!(BasicVersion::parse(" 1.2"), None);
+            assert_eq!(BasicVersion::parse("1.2 "), None);
+            assert_eq!(BasicVersion::parse("1 .2"), None);
+            assert_eq!(BasicVersion::parse("1. 2"), None);
```
gurry: Add a test case for an invalid input which contain only alphabets e.g. "MyFolder".

crates/wdk-build/src/utils.rs
```diff
@@ -26,6 +26,30 @@ use windows::{
 
 use crate::{ConfigError, CpuArchitecture};
 
+/// Type for versions of the format "major.minor"
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
```
gurry: I don't think this type needs to implement `Clone` and `Copy`

crates/wdk-build/src/utils.rs
```diff
@@ -374,39 +398,154 @@ fn read_registry_key_string_value(
...
+            assert!(debug_str.contains("major: 1"));
+            assert!(debug_str.contains("minor: 2"));
+        }
+
+        #[test]
+        fn test_max_selection() {
+            let versions = vec![
+                BasicVersion::new(1, 2),
+                BasicVersion::new(1, 10),
+                BasicVersion::new(2, 0),
```
gurry: Also add a value in it with major version 2 e.g. 2.1

crates/wdk-build/src/utils.rs
```diff
@@ -374,39 +398,154 @@ fn read_registry_key_string_value(
...
+            assert_eq!(v2, v3);
+        }
+
+        #[test]
+        fn test_debug_formatting() {
+            let version = BasicVersion::new(1, 2);
+            let debug_str = format!("{:?}", version);
+            assert!(debug_str.contains("BasicVersion"));
+            assert!(debug_str.contains("major: 1"));
+            assert!(debug_str.contains("minor: 2"));
```
gurry:
```suggestion
            assert_eq!(debug_str, "BasicVersion { major: 1, minor: 2 }");
```

Instead of the `contains` checks, better to just do a single exact check against the full debug string

crates/wdk-build/src/utils.rs
gurry: May be add a test case for overflow as well e.g. for major or minor strings having values bigger than `u32::MAX`

crates/wdk-build/src/utils.rs
```diff
@@ -26,6 +26,30 @@ use windows::{
 
 use crate::{ConfigError, CpuArchitecture};
 
+/// Type for versions of the format "major.minor"
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct BasicVersion {
+    /// The major version number
+    pub major: u32,
+    /// The minor version number
+    pub minor: u32,
```
gurry: Consider using a tuple struct instead of named fields .e.g. `BasicVersion(u32, u32)`

gurry (approved):

.github/workflows/codeql.yml
```diff
@@ -60,12 +60,35 @@ jobs:
...
-          Write-Host "Installing ${{ matrix.wdk }}..."
-          Install-WinGetPackage -Id ${{ matrix.wdk }} -Source winget -MatchOption Equals -Mode Silent -Force
+        # Delete the Ge WDK (10.0.26100) Include and Lib folders if present on the system.
+        # Ni WDK (10.0.22621) will be the latest available Kit on the system.
+        # FIXME: Remove once issues with Ge WDK are resolved.
+        $wdkGePaths = @(
+          "C:\Program Files (x86)\Windows Kits\10\Include\10.0.26100.0",
+          "C:\Program Files (x86)\Windows Kits\10\Lib\10.0.26100.0"
+        )
+        foreach ($path in $wdkGePaths) {
```
krishnakumar4a4: Just wanted to confirm -- uninstall-wingetpackage did not work for us here?

crates/wdk-build/src/utils.rs
```diff
@@ -26,6 +26,21 @@ use windows::{
...
 use crate::{ConfigError, CpuArchitecture};
 
+/// Type for versions of the format "major.minor"
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) struct TwoPartVersion(pub(crate) u32, pub(crate) u32);
+
+impl TwoPartVersion {
+    /// Parse a version in the "x.y" format
+    #[must_use]
+    pub fn parse(s: &str) -> Option<Self> {
```
krishnakumar4a4: 1. We should return a Result for this.
2. Looking at naming conventions. [parse](https://doc.rust-lang.org/std/string/struct.String.html#method.parse) is a method on str (we can call it source type) and convert to another target type `T`. Whereas [from_str](https://docs.rs/serde_json/latest/serde_json/fn.from_str.html) is a method on target type `T` itself and takes source type as argument. I think our conversion similar to the latter case, and it would be better to call this method `from_str`

krishnakumar4a4 (changes requested):

.github/workflows/codeql.yml
svasista-ms: Yes @krishnakumar4a4, `winget list` listed out both the versions 

image

But `Get-WinGetPackage -Id "WDK" -MatchOption Contains` does not show `10.0.26100`. Probably some issue with the installation.

image

crates/wdk-build/src/utils.rs
svasista-ms: Thank you for reviewing @krishnakumar4a4. I implemented the `FromStr` trait for `TwoPartVersion` which made the code more readable and improved error handling for `TwoPartVersion` type and `find_max_version_in_directory` function.

gurry:
We may have gone a bit too far by implementing `FromStr` and introducing new error types for `TwoPartVersion`. But anyway, let's get the PR in so that the pipeline is green.

gurry (approved):

svasista-ms:
> We may have gone a bit too far by implementing `FromStr` and introducing new error types for `TwoPartVersion` . But anyway, let's get the PR in so that the pipeline is green.

I think implementing `FromStr` instead of `parse` was required. It is more idiomatic compared to a custom `parse` method.

gurry:
> I think implementing `FromStr` instead of `parse` was required. It is more idiomatic compared to a custom `parse` method.

Yeah, it's idiomatic but not required because `TwoPartVersion` is not a general type that would be used widely.

svasista-ms:
> > I think implementing `FromStr` instead of `parse` was required. It is more idiomatic compared to a custom `parse` method.
> 
> Yeah, it's idiomatic but not required because `TwoPartVersion` is not a general type that would be used widely.

Yes, not very widely used. But this type will be handy when we implement same logic for other versioned headers in the Kit.

.github/workflows/codeql.yml
```diff
@@ -60,12 +60,35 @@ jobs:
...
       run: |
-        if ((Get-WinGetPackage -Id ${{ matrix.wdk }} -Source winget -MatchOption Equals).Id -eq '${{ matrix.wdk }}') {
-          Write-Host "${{ matrix.wdk }} is already installed. Attempting to update..."
-          Update-WinGetPackage -Id ${{ matrix.wdk }} -Source winget -MatchOption Equals -Mode Silent -Force
-        } else {
-          Write-Host "Installing ${{ matrix.wdk }}..."
-          Install-WinGetPackage -Id ${{ matrix.wdk }} -Source winget -MatchOption Equals -Mode Silent -Force
+        # Delete the Ge WDK (10.0.26100) Include and Lib folders if present on the system.
+        # Ni WDK (10.0.22621) will be the latest available Kit on the system.
+        # FIXME: Remove once issues with Ge WDK are resolved.
```
wmmc88: Create an issue tracking this and assign to yourself (or link relevant issues here)

.github/workflows/codeql.yml
```diff
@@ -60,12 +60,35 @@ jobs:
 
     - name: Install WDK (${{ matrix.wdk }})
       run: |
-        if ((Get-WinGetPackage -Id ${{ matrix.wdk }} -Source winget -MatchOption Equals).Id -eq '${{ matrix.wdk }}') {
-          Write-Host "${{ matrix.wdk }} is already installed. Attempting to update..."
-          Update-WinGetPackage -Id ${{ matrix.wdk }} -Source winget -MatchOption Equals -Mode Silent -Force
-        } else {
-          Write-Host "Installing ${{ matrix.wdk }}..."
-          Install-WinGetPackage -Id ${{ matrix.wdk }} -Source winget -MatchOption Equals -Mode Silent -Force
+        # Delete the Ge WDK (10.0.26100) Include and Lib folders if present on the system.
```
wmmc88: Just to be clear, you're goal is actually to delete any pre-installed WDKs that come with the runner image, right?

crates/wdk-build/src/lib.rs
```diff
@@ -1120,6 +1111,63 @@ impl Config {
 
         enabled_cpu_target_features.contains(STATICALLY_LINKED_C_RUNTIME_FEATURE_NAME)
     }
+
+    /// Constructs the architecture-specific Windows SDK library path using the
```
wmmc88, line 1140: Thank you for these good rustdoc comments

crates/wdk-build/src/lib.rs
```diff
@@ -1120,6 +1111,63 @@ impl Config {
...
+            return Err(ConfigError::DirectoryNotFound {
+                directory: windows_sdk_library_path.to_string_lossy().into(),
+            });
+        }
+        Ok(windows_sdk_library_path)
+    }
+
+    /// Returns the path to the latest available UCX header file present in the
+    /// Lib folder of the WDK content root
+    // TODO: SDK and UCX version should be configurable
```
wmmc88: Is there an issue for this?

crates/wdk-build/src/lib.rs
```diff
@@ -1120,6 +1111,63 @@ impl Config {
...
+            utils::get_latest_windows_sdk_version(&self.wdk_content_root.join("Lib"))?;
+        let ucx_header_root_dir = self.sdk_library_path(sdk_version)?.join("ucx");
+        let max_version =
+            utils::find_max_version_in_directory(&ucx_header_root_dir).map_err(|_| {
+                ConfigError::DirectoryNotFound {
+                    directory: ucx_header_root_dir.to_string_lossy().into(),
+                }
+            })?;
+        let path = format!("ucx/{}.{}/ucxclass.h", max_version.0, max_version.1);
+        Ok(Box::leak(path.into_boxed_str()))
```
wmmc88: this is a memory leak. i think you just return an owned `PathBuf` here

crates/wdk-build/src/utils.rs
```diff
@@ -81,6 +82,42 @@ where
...
+pub(crate) enum TwoPartVersionError {
+    #[error("Invalid version: {0}. Expected format is 'major.minor'")]
+    InvalidFormat(String),
+    #[error("Error parsing {0} version to 'u32'. Version string: {1}")]
+    ParseError(String, String),
+}
+
+/// Type for versions of the format "major.minor"
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) struct TwoPartVersion(pub(crate) u32, pub(crate) u32);
```
wmmc88, line 96: nit: IMO its always clearer to used named args (i.e major and minor) instead of tuple structs

wmmc88 (changes requested):
Actually, the Box leak should definitely be fixed before merge. Probably can just make `usb_headers` fn return a `Cow<'static, str>`

.github/workflows/codeql.yml
svasista-ms: Yes, that is correct. The `Include` and `Lib` folders of the pre-installed WDK (10.0.26100) are manually deleted. This will ensure that the version installed in the workflow, i.e. 10.0.22621 is chosen for the build.

crates/wdk-build/src/utils.rs
gurry: I am not sure about that. Tuple structs are a good fit for data that is inherently tuple like such as version in this case.

crates/wdk-build/src/lib.rs
gurry: Whoops

gurry:
> Yes, not very widely used. But this type will be handy when we implement same logic for other versioned headers in the Kit.

Remember [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

.github/workflows/codeql.yml
svasista-ms: In #406, install WDK step is moved to a separate action. Since `codeql.yml` will depend on this action, the workaround will be cleaned up. Created #424 to track the other known issue (`bindgen` failure due to `redefinition of enumerator` error) with WDK v10.0.26100. Once this issue is closed and #406 is merged, this workaround will be removed.

crates/wdk-build/src/lib.rs
svasista-ms: We already have #412 that is tracking "making versioned headers (metadata) configurable". The SDK version is being handled as part of the system dependency work.

crates/wdk-build/src/lib.rs
```diff
@@ -920,20 +947,27 @@ impl Config {
     /// Returns a [`String`] containing the contents of a header file designed
     /// for [`bindgen`](https://docs.rs/bindgen) to process
     ///
-    /// The contents contain `#include`'ed headers based off the [`ApiSubset`]
-    /// and [`Config`], as well as any additional definitions required for the
-    /// headers to be processed successfully
+    /// The contents contain `#include`'ed headers based on the provided
+    /// [`ApiSubset`]s and the [`Config`].
```
gurry: I feel the older comment was better and clearer

crates/wdk-build/src/lib.rs
```diff
@@ -920,20 +947,27 @@ impl Config {
...
-    ) -> String {
+    ) -> Result<String, ConfigError> {
         api_subsets
             .into_iter()
-            .flat_map(|api_subset| {
-                self.headers(api_subset)
-                    .map(|header| format!("#include \"{header}\"\n"))
+            .map(|api_subset| self.headers(api_subset))
+            .collect::<Result<Vec<_>, _>>()?
+            .into_iter()
```
gurry, line 959: Do we need these `collect` and `into_iter` calls?

crates/wdk-build/src/lib.rs
```diff
@@ -1611,7 +1643,7 @@ mod tests {
             });
 
             assert_eq!(
-                config.bindgen_header_contents([ApiSubset::Base]),
+                config.bindgen_header_contents([ApiSubset::Base]).unwrap(),
```
gurry, line 1636: Instead of using `unwrap` better to put an `Ok()` around the string in right hand side of this assert.

crates/wdk-build/src/lib.rs
```diff
@@ -1631,7 +1663,9 @@ mod tests {
             });
 
             assert_eq!(
-                config.bindgen_header_contents([ApiSubset::Base, ApiSubset::Wdf]),
+                config
+                    .bindgen_header_contents([ApiSubset::Base, ApiSubset::Wdf])
+                    .unwrap(),
```
gurry, line 1658: Same thing here. Use `Ok()` around the RHS instead of the `unwrap`. Applies to all other similar places

crates/wdk-build/src/lib.rs
```diff
@@ -723,24 +720,31 @@ impl Config {
         .map(std::string::ToString::to_string)
     }
 
-    /// Returns a [`String`] iterator over all the headers for a given
-    /// [`ApiSubset`]
+    /// Returns an iterator over the headers as [`String`]s  based on the
+    /// [`Config`] for a given [`ApiSubset`].
```
gurry: The older comment was better I feel

crates/wdk-build/src/lib.rs
```diff
@@ -1604,12 +1686,246 @@ mod tests {
...
+                "Expected KMDF USB test to fail with nonexistent WDK path"
+            );
+
+            #[cfg(nightly_toolchain)]
+            assert_matches!(
+                result.unwrap_err(),
+                ConfigError::IoError(_),
+                "Expected IoError due to nonexistent WDK path during UCX header resolution"
+            );
+        }
```
gurry: I haven't gone through these new tests as I am not fully familiar with the domain logic. Leaving it to @wmmc88 to check them.

crates/wdk-build/src/lib.rs
wmmc88: ok. with all these TODO's/workarounds, make sure you link a relevant bug if you're leaving the TODO in the repo

crates/wdk-build/src/lib.rs
```diff
@@ -712,19 +725,29 @@ impl Config {
...
             ApiSubset::Gpio => self.gpio_headers(),
             ApiSubset::Hid => self.hid_headers(),
             ApiSubset::ParallelPorts => self.parallel_ports_headers(),
             ApiSubset::Spb => self.spb_headers(),
             ApiSubset::Storage => self.storage_headers(),
-            ApiSubset::Usb => self.usb_headers(),
-        }
-        .into_iter()
-        .map(str::to_string)
+            ApiSubset::Usb => return self.usb_headers().map(std::iter::IntoIterator::into_iter),
```
wmmc88, line 735: its awkward to specialcase this one branch to early return. should make them all uniform (which is why I suggests just standardising on CoW: https://github.com/microsoft/windows-drivers-rs/pull/411#pullrequestreview-3022563060

crates/wdk-build/src/lib.rs
```diff
@@ -922,18 +952,24 @@ impl Config {
...
+    ) -> Result<String, ConfigError> {
         api_subsets
             .into_iter()
-            .flat_map(|api_subset| {
-                self.headers(api_subset)
-                    .map(|header| format!("#include \"{header}\"\n"))
+            .try_fold(String::new(), |mut acc, api_subset| {
+                for header in self.headers(api_subset)? {
+                    writeln!(acc, "#include \"{header}\"")
+                        .map_err(ConfigError::BindgenHeaderContents)?;
```
wmmc88: `ConfigError::BindgenHeaderContents` seems to only exist because of how this code is structured here (writeln). This string concatenation should be infallible, so this error shouldn't exist

crates/wdk-build/src/lib.rs
```diff
@@ -922,18 +952,24 @@ impl Config {
...
-    ) -> String {
+    ) -> Result<String, ConfigError> {
         api_subsets
             .into_iter()
-            .flat_map(|api_subset| {
-                self.headers(api_subset)
-                    .map(|header| format!("#include \"{header}\"\n"))
+            .try_fold(String::new(), |mut acc, api_subset| {
+                for header in self.headers(api_subset)? {
+                    writeln!(acc, "#include \"{header}\"")
```
wmmc88: this is missing the newlines compared to the old logic

crates/wdk-build/src/lib.rs
```diff
@@ -922,18 +952,24 @@ impl Config {
...
         &self,
         api_subsets: impl IntoIterator<Item = ApiSubset>,
-    ) -> String {
+    ) -> Result<String, ConfigError> {
         api_subsets
             .into_iter()
-            .flat_map(|api_subset| {
-                self.headers(api_subset)
-                    .map(|header| format!("#include \"{header}\"\n"))
+            .try_fold(String::new(), |mut acc, api_subset| {
```
wmmc88: the logic should look much closer to what it used to be instead of using try_fold here. You should be able to turn `self.headers(api_subset).map(XYZ)` to map from an iterator of results of strings into another iterator of results of strings. 

The outer flatmap can now flatmap iterator of api_subset to iterator of results of strings (previously was iterator of api_subset to iterator of strings). the final collect can collect an iterator of results into a single result which you can `?` on. essentially this: https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect

crates/wdk-build/src/lib.rs
gurry: We had a discussion about it. Using a `Cow` meant the return type of functions `base_headers`, `wdf_headers`, `gpio_headers`, `hid_headers`, `spb_headers`, `storage_headers` and `usb_headers` will all have to change to `Vec<Cow<'_, str>>` and therefore every element we put in a `Vec` in each of them will have to be made into a `Cow`. Compared to that doing an early return in one function seemed less awkward. That is not to say it is not ugly in itself, but it seemed less ugly in comparison.

crates/wdk-build/src/lib.rs
gurry: This is much better.

crates/wdk-build/src/lib.rs
svasista-ms: Thank you @wmmc88, removed the error in the latest commit. Please review

crates/wdk-build/src/lib.rs
svasista-ms: Reverted back to `format!` with `\n` instead of `writeln!`

crates/wdk-build/src/lib.rs
svasista-ms: @wmmc88 , @gurry , modified this to map each `api_subset`, collect them to a `Result` of `Vec`s, `flat_map` to generate the `#include`s and then finally collect them to a string. Since the `bindgen_header_contents` function now returns a `Result`, I had to wrap this in `Ok`. Please review this logic. Thank you

gurry: LGTM

crates/wdk-build/src/lib.rs
wmmc88: I'm assuming that most of these will need to return non-owned strings possibly due to the same reason that usb_headers need to

crates/wdk-build/src/utils.rs
wmmc88: but then it becomes ambiguous as to what corresponds with what in the tuple

crates/wdk-build/src/lib.rs
```diff
@@ -243,6 +243,10 @@ rustflags = [\"-C\", \"target-feature=+crt-static\"]
     /// [`metadata::Wdk`]
     #[error(transparent)]
     SerdeError(#[from] metadata::Error),
+
+    /// Error returned when the UCX header file is not found
+    #[error("failed to find ucx header file")]
+    UCXHeaderNotFound(#[source] std::io::Error),
```
wmmc88: nit: Should this be more generic instead of just for UCX? extra metadata can be in the error (like api subset or the path that doesn't exist

wmmc88 (dismissed):

crates/wdk-build/src/utils.rs
gurry: The order of tuple fields corresponds to how you would write the version e.g. 

`TwoPartVersion(1, 5)` <-> `"1.5"`

so it is quite clear which field corresponds to what.

crates/wdk-build/src/lib.rs
gurry: Yes, all of them except `usb_headers` return `Vec<&str>`. `usb_headers` used to return non-owned strings but can't any more.

gurry (approved):

crates/wdk-build/src/lib.rs
```diff
@@ -1604,12 +1677,246 @@ mod tests {
...
+            }
+        }
+
+        /// Verifies WDM-specific USB header inclusions and exclusions
+        fn assert_wdm_usb_headers(header_contents: &str) {
+            // WDM: kernel-mode USB headers but no WDF or UCX
+            let required_headers = ["usbbusif.h", "usbdlib.h", "usbfnattach.h", "usbfnioctl.h"];
+
+            let excluded_headers = ["wdfusb.h"];
+            let excluded_header_paths = ["ucx/", "ucm/"];
```
krishnakumar4a4: This should be an exhaustive list and include all of the below
```
                    "ucm/1.0/UcmCx.h",
                    "UcmTcpci/1.0/UcmTcpciCx.h",
                    "UcmUcsi/1.0/UcmucsiCx.h",
                    "ude/1.1/UdeCx.h",
                    "ufx/1.1/ufxbase.h",
                    "ufxproprietarycharger.h",
                    "urs/1.0/UrsCx.h",
```

crates/wdk-build/src/lib.rs
```diff
@@ -1604,12 +1677,246 @@ mod tests {
...
+        }
+
+        /// Verifies UMDF-specific USB header inclusions and exclusions
+        fn assert_umdf_usb_headers(header_contents: &str) {
+            // UMDF: common headers + WDF, but no kernel-mode or KMDF-specific
+            let excluded_kernel_headers =
+                ["usbbusif.h", "usbdlib.h", "usbfnattach.h", "usbfnioctl.h"];
+
+            let required_headers = ["wdfusb.h"];
+            let excluded_header_paths = ["ucx/", "ucm/", "ude/"];
```
krishnakumar4a4: Same as https://github.com/microsoft/windows-drivers-rs/pull/411/files#r2224200884

crates/wdk-build/src/lib.rs
gurry: Yes, There is no reason for this to be tied to UCX. 

@svasista-ms maybe we can rename it to `HeaderNotFound`, add a for header type and then interpolate that field in the error message e.g.:
```rust
#[error("failed to find {header_type} header file")]
HeaderNotFound(#[source] std::io::Error, header_type: String)
```

We can also add more context such as name of the header if that is relevant/possible.

crates/wdk-build/src/lib.rs
```diff
@@ -1604,12 +1677,246 @@ mod tests {
...
+                    !header_contents.contains(path),
+                    "{path} headers should not be included"
+                );
+            }
+        }
+
+        /// Tests USB header generation for a specific driver configuration
+        fn test_usb_headers_with(driver_config: &DriverConfig, cpu_architecture: CpuArchitecture) {
+            // Create a temporary directory structure that mimics a WDK installation
+            let temp_dir = TempDir::new().unwrap();
```
krishnakumar4a4, line 1815: I don't have a good feeling about unit tests having side effects like creating directories and interacting with the file system. I think it's best to move it out to `wdk-build/tests/` and treat this as an integration test where side effects are generally expected.

crates/wdk-build/src/utils.rs
```diff
@@ -515,4 +807,124 @@ mod tests {
             )
         );
     }
+
+    mod find_max_version_in_directory {
```
krishnakumar4a4, line 811: Similar case https://github.com/microsoft/windows-drivers-rs/pull/411/files#r2224217472

krishnakumar4a4 (changes requested):
I’ve added a few minor comments, primarily on the tests.

crates/wdk-build/src/lib.rs
```diff
@@ -245,8 +245,8 @@ rustflags = [\"-C\", \"target-feature=+crt-static\"]
     SerdeError(#[from] metadata::Error),
 
     /// Error returned when the UCX header file is not found
-    #[error("failed to find ucx header file")]
-    UCXHeaderNotFound(#[source] std::io::Error),
+    #[error("failed to find {0} header file: {1}")]
```
gurry: The `std::io::Error` field will show up as the underlying cause anyway due to the `#[source]` attribute. No need to include it in the error message.

gurry (approved):

crates/wdk-build/src/lib.rs
krishnakumar4a4: Since all fields of the Config struct cannot initialized from the integration test, we can defer this change for now.

crates/wdk-build/src/utils.rs
krishnakumar4a4: I am not sure if you have faced any issue in attempting to move out the integration tests, but we can do it together with other one https://github.com/microsoft/windows-drivers-rs/pull/411#discussion_r2224217472 since this is an high priority PR.

crates/wdk-build/src/lib.rs
svasista-ms: We may have to introduce setter methods or have a new/constructor method that takes parameters. This allows us to create mocks for the integration tests. As suggested, I will take this up in a separate PR.

crates/wdk-build/src/utils.rs
svasista-ms: Sure, thanks. I did not attempt to move these tests. We can do it along with the other tests.

Issue #348, opened 2025-05-06
[cargo-wdk] Ensure cargo-wdk respect other forms of target override


krishnakumar4a4:
[The first cut of cargo-wdk](https://github.com/microsoft/windows-drivers-rs/pull/306) only respects the `--target-arch` CLI argument to build the driver to specific CPU architecture (only Amd64 & Arm64 supported as of now).

If the driver developer attempts to override target through a `config.toml` or through an env variable, the behaviour is undetermined & unsupported as of now.

This issue is opened to track and address the same.

PR #420, opened 2025-07-09
ci: add ARM64 Windows runner support to workflows

krishnakumar4a4:
This PR includes the following changes:
- Added windows-11-arm64 runners alongside windows-latest for the build, test, lint, local-development-makefile, docs and codeQL workflows.
- Updated the winget-install action to use latest winget-cli build.
- Simplified the test and local-development workflows by removing `target_triple` from the matrix, as cross-compilation wouldn't work.


.github/workflows/lint.yaml
```diff
@@ -78,7 +92,7 @@ jobs:
           targets: ${{ matrix.target_triple.name }}
 
       - name: Run Cargo Clippy
-        run: cargo +${{ matrix.rust_toolchain }} clippy --locked --profile ${{ matrix.cargo_profile }} --target ${{ matrix.target_triple.name }} --all-targets --all-features
+        run: cargo +${{ matrix.rust_toolchain }} clippy --locked --profile ${{ matrix.cargo_profile }} --target ${{ matrix.target_triple.name }} --all-features
```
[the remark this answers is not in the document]
krishnakumar4a4: The `--all-targets` argument was redundant and has been removed, since `--target` already exists and all targets are covered as part of the matrix.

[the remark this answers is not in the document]
wmmc88: @krishnakumar4a4 that is not correct. The `--all-targets` refers to cargo targets, while `--target` refers to target architecture. The former lets clippy lint build scripts, tests, etc. It should not be removed

.github/actions/install-wdk/action.yml
```diff
@@ -26,6 +26,7 @@ runs:
         $targetArch = '${{ inputs.target }}'
         switch ($targetArch) {
             'amd64' { $targetArch = 'x64' }
+            'x64' { $targetArch = 'x64' }
```
wmmc88: Why the redundancy here? should be consistent across the repo

crates/cargo-wdk/tests/wdm-driver/src/lib.rs
```diff
@@ -16,10 +16,16 @@ use wdk_alloc::WdkAllocator;
 #[global_allocator]
 static GLOBAL_ALLOCATOR: WdkAllocator = WdkAllocator;
 
+extern crate alloc;
```
wmmc88: this seems very odd/code smell. why don't we see this when building sample-wdm-driver on arm64?

.github/workflows/lint.yaml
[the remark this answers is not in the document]
krishnakumar4a4: Okay I see, I went by the names as `--all-targets` arg seemed very similar to `--target`. Now I looked it up and understood the distinction. The `--target` arg takes target triple as value and cross compiles to it. Whereas the target in the `--all-targets` arg refers to build targets lib, bins, examples, tests, etc which is not the same as former.

I will revert this change and wherever I might have removed it. Thanks @wmmc88

.github/workflows/build.yaml
```diff
@@ -91,14 +105,20 @@ jobs:
         run: cargo make --cwd ./examples build +${{ matrix.rust_toolchain }} --locked --profile ${{ matrix.cargo_profile }} --target ${{ matrix.target_triple.name }}
 
       - name: Package Examples (via Cargo Make)
+        # Skip this step when WDK 10.0.22621 is used with ARM64 runner as it doesn't ship certmgr and makecert binaries
+        if: ${{ !(matrix.wdk.version == '10.0.22621' && matrix.runner.arch == 'arm64') }}
```
krishnakumar4a4: Note: I have noticed the package 10.0.22621 is missing certmgr and makecert  exes on arm64 machine when installed via winget channel. This `if` block is to skip the steps where these exes are needed.

.github/workflows/build.yaml
```diff
@@ -53,20 +61,26 @@ jobs:
 
       - name: Install LLVM ${{ matrix.llvm }}
         run: |
-          if ((Get-WinGetPackage -Id LLVM -Source winget -MatchOption Equals).InstalledVersion -eq '${{ matrix.llvm }}') {
-            Write-Host "LLVM ${{ matrix.llvm }} is already installed."
-          } else {
-            Write-Host "Installing LLVM ${{ matrix.llvm }}..."
-            Install-WinGetPackage -Id LLVM.LLVM -Version ${{ matrix.llvm }} -Source winget -MatchOption Equals -Mode Silent -Force
+          # Use native winget.exe instead of PowerShell module for ARM64 compatibility
```
krishnakumar4a4: Note: This new definition always installs LLVM, even if it is already present. In contrast, the older definition skips the installation if LLVM is already installed. I'm torn between adding that condition here or continuing to install it unconditionally, as parsing the winget output to check for an existing LLVM installation doesn't seem very reliable.

Update: Since LLVM version (17.0.6) being installed here always gets installed even with older and newer definition. The net effect of this change is going to be zero.

.github/workflows/lint.yaml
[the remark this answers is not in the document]
gurry: Naming was already hard. Rust made it 10x harder.

.github/actions/winget-install/action.yaml
```diff
@@ -12,31 +12,54 @@ runs:
       run: |
         $AuthenticatedHeaders = @{ "Authorization" = "Bearer ${{ inputs.GITHUB_TOKEN }}" }
 
-        $ReleaseInfo = Invoke-RestMethod -Headers $AuthenticatedHeaders 'https://api.github.com/repos/microsoft/winget-cli/releases/164835566'
+        # Detect runner architecture
+        $Architecture = if ($env:RUNNER_ARCH -eq "ARM64") { "arm64" } else { "x64" }
+        Write-Host "Detected architecture: $Architecture"
```
gurry:
```suggestion
        Write-Host "Runner architecture: $Architecture"
```

The word "Runner" makes it clear that it's the runner's architecture we are talking about. "Detected" can be removed as it is not adding any useful information.

.github/actions/winget-install/action.yaml
```diff
@@ -12,31 +12,54 @@ runs:
...
+
+        # winget-cli release v1.11.400
+        $ReleaseInfo = Invoke-RestMethod -Headers $AuthenticatedHeaders 'https://api.github.com/repos/microsoft/winget-cli/releases/226250169'
         $WingetDownloadUri = $ReleaseInfo.assets.browser_download_url | Where-Object { $_.EndsWith('.msixbundle') }
         $WingetLicenseDownloadUri = $ReleaseInfo.assets.browser_download_url | Where-Object { $_.EndsWith('License1.xml') }
+        $WingetDependenciesZipUri = $ReleaseInfo.assets.browser_download_url | Where-Object { $_.EndsWith('DesktopAppInstaller_Dependencies.zip') }
```
gurry:
```suggestion
        $WingetDependenciesDownloadUri = $ReleaseInfo.assets.browser_download_url | Where-Object { $_.EndsWith('DesktopAppInstaller_Dependencies.zip') }
```
For consistency as the other two variables also have the word "Download" in them

.github/actions/winget-install/action.yaml
```diff
@@ -12,31 +12,54 @@ runs:
...
         # Print to logs
         Write-Host "WingetDownloadUri=$WingetDownloadUri"
         Write-Host "WingetLicenseDownloadUri=$WingetLicenseDownloadUri"
+        Write-Host "WingetDependenciesZipUri=$WingetDependenciesZipUri"
 
         # Save output for next step
         Write-Output "WingetDownloadUri=$WingetDownloadUri" >> $env:GITHUB_ENV
         Write-Output "WingetLicenseDownloadUri=$WingetLicenseDownloadUri" >> $env:GITHUB_ENV
+        Write-Output "Architecture=$Architecture" >> $env:GITHUB_ENV
+        Write-Output "WingetDependenciesZipUri=$WingetDependenciesZipUri" >> $env:GITHUB_ENV
```
gurry:
```suggestion
        Write-Output "WingetDependenciesDownloadUri=$WingetDependenciesDownloadUri" >> $env:GITHUB_ENV
```

.github/actions/winget-install/action.yaml
```diff
@@ -12,31 +12,54 @@ runs:
...
 
     - name: Start Winget Installation for all Users
       shell: pwsh
       run: |
-        Add-AppxProvisionedPackage -Online -PackagePath $env:RUNNER_TEMP/winget-install/winget.msixbundle -LicensePath $env:RUNNER_TEMP/winget-install/license.xml -DependencyPackagePath $env:RUNNER_TEMP/winget-install/Microsoft.UI.Xaml.2.8.x64.appx, $env:RUNNER_TEMP/winget-install/Microsoft.VCLibs.x64.14.00.Desktop.appx
+        # Use architecture-specific dependency paths
+        if ($env:Architecture -eq "arm64") {
+          [string[]]$DependencyPaths = (Get-ChildItem -Path "$env:RUNNER_TEMP/winget-install/arm64" -Filter '*.appx' -File -Force).FullName
+          Write-Host "Found Dependency $($DependencyPaths[0])"
+          Write-Host "Found Dependency $($DependencyPaths[1])"
```
gurry: It's not clear what the first two dependencies in the array are and why they are important. Better to assign them to variables with meaningful names and use those variables instead.

.github/actions/winget-install/action.yaml
```diff
@@ -12,31 +12,54 @@ runs:
...
+          Add-AppxProvisionedPackage -Online -PackagePath $env:RUNNER_TEMP/winget-install/winget.msixbundle -LicensePath $env:RUNNER_TEMP/winget-install/license.xml -DependencyPackagePath "$($DependencyPaths[0])", "$($DependencyPaths[1])"
+        } else {
+          [string[]]$DependencyPaths = (Get-ChildItem -Path "$env:RUNNER_TEMP/winget-install/x64" -Filter '*.appx' -File -Force).FullName
+          Write-Host "Found Dependency $($DependencyPaths[0])"
+          Write-Host "Found Dependency $($DependencyPaths[1])"
+          Add-AppxProvisionedPackage -Online -PackagePath $env:RUNNER_TEMP/winget-install/winget.msixbundle -LicensePath $env:RUNNER_TEMP/winget-install/license.xml -DependencyPackagePath "$($DependencyPaths[0])", "$($DependencyPaths[1])"
+        }
```
gurry: Both the if and else blocks are identical except for the architecture part of the path in the first statement. We can completely remove if-else if we use the $env:Architecture in the path like so:
```
[string[]]$DependencyPaths = (Get-ChildItem -Path "$env:RUNNER_TEMP/winget-install/$env:Architecture" -Filter '*.appx' -File -Force).FullName
```

.github/workflows/build.yaml
```diff
@@ -15,9 +15,15 @@ env:
 jobs:
   build:
     name: Build
-    runs-on: windows-latest
     strategy:
+      fail-fast: false
```
gurry: This change is already coming in PR https://github.com/microsoft/windows-drivers-rs/pull/439. Consider removig it from here to avoid a merge conflict.

.github/workflows/build.yaml
gurry: If LLVM installation is fast and is not liable to fail due to the existing one, then forced installation is fine.

.github/workflows/build.yaml
```diff
@@ -53,20 +61,26 @@ jobs:
 
       - name: Install LLVM ${{ matrix.llvm }}
         run: |
-          if ((Get-WinGetPackage -Id LLVM -Source winget -MatchOption Equals).InstalledVersion -eq '${{ matrix.llvm }}') {
-            Write-Host "LLVM ${{ matrix.llvm }} is already installed."
-          } else {
-            Write-Host "Installing LLVM ${{ matrix.llvm }}..."
-            Install-WinGetPackage -Id LLVM.LLVM -Version ${{ matrix.llvm }} -Source winget -MatchOption Equals -Mode Silent -Force
+          # Use native winget.exe instead of PowerShell module for ARM64 compatibility
+          Write-Host "Checking for existing LLVM installation..."
```
gurry:
```suggestion
          Write-Host "Existing LLVM installation(s):"
```
If we are merely listing what's currently present and not really changing our behavior based on that, then saying "checking" is a bit wrong. The suggested text should reflect the intent better.

.github/workflows/build.yaml
gurry: Is it possible to centralize this predicate into a variable e.g. `$certToolsMissing` at the top somewhere and then use that variable instead of duplicating the predicate multiple times?

.github/workflows/codeql.yaml
```diff
@@ -55,20 +62,26 @@ jobs:
...
+        # Use native winget.exe instead of PowerShell module for ARM64 compatibility
+        Write-Host "Checking for existing LLVM installation..."
+        winget list --id LLVM.LLVM --exact --accept-source-agreements --disable-interactivity
+
+        Write-Host "Force Installing LLVM ${{ matrix.llvm }}..."
+        winget install --id LLVM.LLVM --version ${{ matrix.llvm }} --source winget --force --exact --silent --accept-source-agreements --accept-package-agreements --disable-interactivity
+
+        if ($LASTEXITCODE -ne 0) {
+          Write-Error "LLVM ${{ matrix.llvm }} force installation failed with exit code: $LASTEXITCODE"
+          exit 1
```
gurry: Since this code is duplicated in multiple workflows better to extract it into an action.

.github/workflows/lint.yaml
```diff
@@ -14,11 +14,17 @@ env:
 jobs:
   clippy:
     name: Clippy
-    runs-on: windows-latest
     permissions:
       checks: write
     strategy:
+      fail-fast: false
```
gurry: Consider removing it as it is already coming in PR #439

.github/workflows/local-development-makefile.yaml
```diff
@@ -28,9 +34,19 @@ jobs:
         target_triple:
           - name: x86_64-pc-windows-msvc
             arch: amd64
-          # - name: aarch64-pc-windows-msvc FIXME: validate on aarch64
-          #   arch: arm64
-
```
gurry: As per the PR description the target triple was removed because cross compilation doesn't make sense in the local development scenario. Trying to understand that better, why doesn't it make sense?

.github/workflows/test.yaml
```diff
@@ -10,9 +10,15 @@ name: Test
 jobs:
   test:
     name: Test
-    runs-on: windows-latest
     strategy:
+      fail-fast: false
```
gurry: Consider removing it

crates/cargo-wdk/templates/wdm/lib.rs.tmp
```diff
@@ -16,10 +16,14 @@ use wdk_alloc::WdkAllocator;
...
 
+extern crate alloc;
+use alloc::ffi::CString;
+
 #[unsafe(export_name = "DriverEntry")] // WDF expects a symbol with the name DriverEntry
 pub unsafe extern "system" fn driver_entry(
    _driver: PDRIVER_OBJECT,
    _registry_path: PCUNICODE_STRING,
 ) -> NTSTATUS {
+   let _string = CString::new("Hello World!\n").unwrap();
```
gurry:
```suggestion
   let _ = CString::new("Hello World!\n").unwrap();
```
Nit

crates/cargo-wdk/tests/wdm-driver/src/lib.rs
```diff
@@ -16,10 +16,16 @@ use wdk_alloc::WdkAllocator;
...
+use alloc::ffi::CString;
+use core::hint::black_box;
+
 #[unsafe(export_name = "DriverEntry")] // WDF expects a symbol with the name DriverEntry
 pub unsafe extern "system" fn driver_entry(
    _driver: PDRIVER_OBJECT,
    _registry_path: PCUNICODE_STRING,
 ) -> NTSTATUS {
+   // Prevent optimization: ensure allocator is linked (workaround for ARM64 linker issues)
+   black_box(CString::new("Hello World!\n").unwrap());
```
gurry: Why is `black_box` needed here while it isn't in the case of `crates/cargo-wdk/templates/wdm/lib.rs.tmp`?

krishnakumar4a4: This suggestion came from copilot for this file. I simply accepted it. I need to replicate the same in the template itself. 

I also wonder if there is a better way to do address this linker issue.

gurry: Also, wouldn't `String` do the trick instead of the somewhat exotic `CString`?

redsun82:
👋 @krishnakumar4a4 I've noticed this PR as we detected an increased number of CodeQL failures coming from here or your fork.

Unfortunately CodeQL doesn't support running under Windows ARM (see [system requirements](https://codeql.github.com/docs/codeql-overview/system-requirements/)). While analysing some languages might work, traced languages definitely don't.

This is prompting us to improve logging of this at the `codeql-action` level, so apologies if the error seemed more esoteric than it should.

Unfortunately, I think you'll need to drop extending the CodeQL workflow to ARM runners 🙇

.github/workflows/build.yaml
krishnakumar4a4: Force installation seems to be holding up, I could see `clang --version` is being printed as 17.0.6 for both x64 and Arm64 runners. 
This particular version needs to be force installed almost all the time as the default LLVM that exists on the runners would be different and higher than 17.0.6

krishnakumar4a4: On an average LLVM installation step is taking approximately one minute. Refer screenshot below

image

.github/workflows/build.yaml
krishnakumar4a4: Yes, one if the recommended ways to do so is to set an env variable `CERT_TOOLS_PRESENT` and use it in the steps. I made that change now.

.github/workflows/build.yaml
krishnakumar4a4: Agreed.
Right now, I need this to debug pipeline issues. Once all the workflows are green, I can consider removing this. 
Although I would like to keep it for one benefit, that if this PR gets merged first, then having `fail-fast: false` helps to identify any issues immediately after the merge.

.github/workflows/lint.yaml
krishnakumar4a4: https://github.com/microsoft/windows-drivers-rs/pull/420#discussion_r2257087537

.github/workflows/test.yaml
krishnakumar4a4: https://github.com/microsoft/windows-drivers-rs/pull/420#discussion_r2257087537

.github/workflows/local-development-makefile.yaml
krishnakumar4a4: I was a bit vague when I stated cross compilation doesn't make sense. Here is the correct explanation for local-development-makefile
1. The existing target_triple `x86_64-pc-windows-msvc` really has no bearing as this is the default target and it should come preinstalled with the rust-toolchain itself. The same thing should apply for even Arm64 runner, i.e `aarch64-pc-windows-msvc` should come pre-installed with rust toolchain.
2. Secondly, the local-development-workflow doesn't expose a straightforward to override target_triple today. It picks the default already.

Hence for the above reasons, I concluded we do not need explicit target_triple in the workflow. As rust toolchain installation step on each runner installs the default target and the local-development-makefile simply picks the default target which is matching our local development workflow.

For test workflow, cross compilation wouldn't make sense as unit tests are generally built as exes to the target architecture (assuming target is different from host), and those binaries can't be run on the host architecture as part of test run. However, there is a provision to make these unit tests run by passing a custom runner configuration in the config.toml matching the target architecture. I believe it is not necessary to explore the custom runner configuration for this PR scope and left it out.

.github/workflows/codeql.yaml
krishnakumar4a4: Make sense, I moved this to install-llvm action.

.github/workflows/build.yaml
krishnakumar4a4: made the change

.github/actions/winget-install/action.yaml
krishnakumar4a4: Makes sense. I made the change accordingly. 
Bunch of this code is generated by copilot.

.github/actions/winget-install/action.yaml
krishnakumar4a4: I created 2 new variables as suggested.

.github/actions/winget-install/action.yaml
krishnakumar4a4: Accepted the change

.github/actions/winget-install/action.yaml
krishnakumar4a4: Changed the variable to `WingetDependenciesZipDownloadUri`

.github/actions/winget-install/action.yaml
krishnakumar4a4: Changed the variable to WingetDependenciesZipDownloadUri

.github/actions/install-wdk/action.yml
krishnakumar4a4: I made architecture names consistent across all the workflows and removed this redundant case as well.

crates/cargo-wdk/tests/wdm-driver/src/lib.rs
krishnakumar4a4: I verified this directly on the arm64 runner and was able to reproduce the linker issue if the `driver_entry` routine is empty. 
Let me verify the same on a VM and see if it reproduces.

.github/workflows/build.yaml
gurry: Sure. In that case, you could just deal with it at the time of conflict resolution. As you see fit

.github/workflows/build.yaml
gurry: Umm... 1 minute per matrix cell per workflow might add up. Can we calculate the total time it might add? If that's too much we should avoid force install.

gurry: > Force installation seems to be holding up, I could see `clang --version` is being printed as 17.0.6 for both x64 and Arm64 runners. This particular version needs to be force installed almost all the time as the default LLVM that exists on the runners would be different and higher than 17.0.6

If this forces a force install on us how was the older logic which didn't use force install working? Were the conditions any different then?

.github/workflows/local-development-makefile.yaml
gurry: Okay. Maybe I'm misunderstanidng the puprpose of local development workflow. What is it for?

.github/workflows/build.yaml
krishnakumar4a4: The older definition skips installation if that particular version already exists, which was always false—Yes, it installed LLVM every time and took approximately one minute. Refer to the main pipeline here -> https://github.com/microsoft/windows-drivers-rs/actions/runs/16788106403/job/47543648050

With the current definition using winget.exe, we install LLVM unconditionally without checking for an existing matching version. This installation also takes approximately one minute, so there is no additional time added—the net time difference is zero.

By the way, if you're by any chance referring to the `--force` flag added to the winget invocation, it also exists in the older definition as `-Force`.

So effectively, I need to go back and update my note to say that this new definition (using winget.exe) has the exact same net effect as the older definition (which used Winget PowerShell modules), even though it doesn’t include a check for an existing LLVM installation. This should help avoid any confusion.

.github/workflows/local-development-makefile.yaml
krishnakumar4a4: We run the wdk-pre-commit-flow cargo-make task locally before pushing commits to WDR. Although it's not mentioned in the README, we use it for sanity checks as part of our development workflow.

Here is its definition, it does execute a bunch of tasks as mentioned in dependencies section

```yaml
[tasks.wdk-pre-commit-flow]
description = "Run pre-commit tasks and checks"
category = "Development"
workspace = false
dependencies = [
  "sort-deps",
  "format-flow",
  "format-toml-conditioned-flow",
  "build",
  "package-driver-workspace-flow",
  "test-flow",
  "clippy-flow",
  "audit-flow",
  "unused-dependencies-flow",
  "docs",
  "nightly-test-flow",
  "nightly-clippy-flow",
  "nightly-docs",
  "beta-test-flow",
  "beta-clippy-flow",
  "beta-docs",
]
```

crates/cargo-wdk/tests/wdm-driver/src/lib.rs
krishnakumar4a4: Looks like I overlooked driver name in your question. The sample-wdm-driver is the example you're referring to, which is different from wdm-driver itself.
The linker problem is only observed with drivers that have an empty DriverEntry routine. The current wdm-driver, used in the cargo-wdk integration test, is the problematic one. In contrast, sample-wdm-driver has a non-empty DriverEntry that includes slightly heavier operations, such as allocating a string and converting a UTF-16 string to a Rust string.

Having said that, I am trying to validate this issue on Arm64 VM right now.

.github/workflows/build.yaml
gurry: Oh, if there's no difference at all, then foce install is totally fine.

.github/workflows/local-development-makefile.yaml
gurry: If it's used only locally then using only the runner/host's architecture is sorta okay. But what would be better is if there was a way to override it if needed because then one can test for cross compilation as well if needed. If that's easy to achieve please do so, otherwise this is fine.

.github/workflows/build.yaml
krishnakumar4a4: Yes, there won't be any change in terms of pipeline run time.

.github/workflows/local-development-makefile.yaml
krishnakumar4a4: I can see from the wdk-build code that target arch is evaluated from the `CARGO_CFG_TARGET_ARCH` environment variable. Therefore, setting it explicitly or via config.toml might work for most of the tasks. However, I suspect we’ll run into issues for few tasks, given at least one known gap with cargo-wdk: https://github.com/microsoft/windows-drivers-rs/issues/348. 

Since this is nice to have for developers, I can open an issue for this and keep https://github.com/microsoft/windows-drivers-rs/issues/348 as dependency to it.

gurry: Sounds good

krishnakumar4a4: I created an issue for the enhancement https://github.com/microsoft/windows-drivers-rs/issues/449 and, hence resolving this comment.

crates/cargo-wdk/tests/wdm-driver/src/lib.rs
krishnakumar4a4: I no longer able to reproduce this linker issue both on Arm64 VM and as well as on the pipeline now. I have reverted the wdm related changes from the PR. It seemed to me like an intermittent issue that happened earlier and now it is not reproducible anymore.

There is one more difference, the state of the workflow when the linker issue happened is only working with older WDK 10.0.22621 installed via winget whereas the latest one is running 10.0.26100 installed via Nuget (10.0.22621 is excluded for arm since certmgr.exe and makecert.exe were missing). I will attempt to run the latest code with 10.0.22621 and see if it reproduces the linker issue.

Update: I re-enabled the older WDK 10.0.22621 on test pipeline and it doesn't reproduce the linker issue but instead it shows the known binaries missing issue. https://github.com/microsoft/windows-drivers-rs/actions/runs/16820885703/job/47647429155?pr=420

-----

Just for history, this is the link to the pipeline which has given me linker error https://github.com/microsoft/windows-drivers-rs/actions/runs/16211475105/job/45772319236?pr=420 and the commit corresponding to it was this https://github.com/microsoft/windows-drivers-rs/tree/74663bcd418846cb72b3dfb1e6bf0a6b6cfb9b9d

The latest pipeline which succeeded without WDM changes is this https://github.com/microsoft/windows-drivers-rs/actions/runs/16807610339/job/47604114956?pr=420 and the commit corresponding to it is this https://github.com/microsoft/windows-drivers-rs/tree/81323beaf84cae84598f368e30ed69b2440d7efa

crates/cargo-wdk/tests/wdm-driver/src/lib.rs
krishnakumar4a4: This change is no longer required and has been removed.
Refer https://github.com/microsoft/windows-drivers-rs/pull/420#discussion_r2261833703

.github/actions/winget-install/action.yaml
```diff
@@ -12,31 +12,52 @@ runs:
...
         # Save output for next step
         Write-Output "WingetDownloadUri=$WingetDownloadUri" >> $env:GITHUB_ENV
         Write-Output "WingetLicenseDownloadUri=$WingetLicenseDownloadUri" >> $env:GITHUB_ENV
+        Write-Output "Architecture=$Architecture" >> $env:GITHUB_ENV
+        Write-Output "WingetDependenciesZipDownloadUri=$WingetDependenciesZipDownloadUri" >> $env:GITHUB_ENV
 
     - name: Download Winget Assets and Dependencies
       shell: pwsh
       run: |
         New-Item -Type Directory $env:RUNNER_TEMP/winget-install
```
gurry: This value is repeated and concatenated with `$env:RUNNER_TEMP` 9-10 times. It would be better to do the concatenation once and put it in a variable called `InstallDir` or something and use that.

.github/actions/winget-install/action.yaml
gurry: Copilot is better at writing new code rather than modifying existing code. Do not trust it blindly.

gurry:
Rename the folder `winget-install` to `install-winget` to make it consistent with the names of other install actions.

.github/actions/winget-install/action.yaml
krishnakumar4a4: Makes sense, I have made the change now.

.github/actions/install-winget/action.yaml
```diff
@@ -32,37 +32,38 @@ runs:
         Write-Output "WingetLicenseDownloadUri=$WingetLicenseDownloadUri" >> $env:GITHUB_ENV
         Write-Output "Architecture=$Architecture" >> $env:GITHUB_ENV
         Write-Output "WingetDependenciesZipDownloadUri=$WingetDependenciesZipDownloadUri" >> $env:GITHUB_ENV
+        Write-Output "InstallWingetTempDir=$env:RUNNER_TEMP/install-winget" >> $env:GITHUB_ENV
 
     - name: Download Winget Assets and Dependencies
       shell: pwsh
       run: |
-        New-Item -Type Directory $env:RUNNER_TEMP/winget-install
+        New-Item -Type Directory $env:InstallWingetTempDir
```
gurry, line 43: In general you can shorten your variable names by omitting from them things that are apparent from the context. For example in this case we can avoid "Winget" in this name because the context tells us what we will install in this dir. 

No need to change in this PR, but something to keep in mind for later.

.github/actions/install-winget/action.yaml
```diff
@@ -0,0 +1,78 @@
+name: "Install Winget"
+description: "Install winget on windows runners since its not installed by default: https://github.com/actions/runner-images/issues/6472"
+inputs:
+  GITHUB_TOKEN:
+    description: "GitHub token to execute authenticated Github API requests (for higher rate limit)"
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - name: Get URIs for Winget v1.11.400 assets
```
wmmc88, line 10: Is there a reason you increased this? I previously pinned it because certain newer versions were problematic

.github/actions/install-llvm/action.yml
```diff
@@ -0,0 +1,27 @@
...
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install LLVM ${{ inputs.version }}
+      shell: pwsh
+      run: |
+        # Use native winget.exe instead of PowerShell module for ARM64 compatibility
+        Write-Host "Existing LLVM installation(s):..."
+        winget list --id LLVM.LLVM --exact --accept-source-agreements --disable-interactivity
```
wmmc88: This doesnt match the old logic. it used to check if the exact version was installed, and if not, it would install it

.github/workflows/build.yaml
```diff
@@ -41,32 +47,33 @@ jobs:
             arch: amd64
           - name: aarch64-pc-windows-msvc
             arch: arm64
+    
+    runs-on: ${{ matrix.runner.name }}
+
+    env:
+      # WDK 10.0.22621 installed via winget channel on ARM64 runner doesn't ship certmgr and makecert binaries
```
wmmc88: I think at this point we've validated that this is not true. it is that the SDK is not installed for that version

.github/workflows/local-development-makefile.yaml
```diff
@@ -25,57 +31,53 @@ jobs:
         llvm:
           - 17.0.6
 
-        target_triple:
-          - name: x86_64-pc-windows-msvc
-            arch: amd64
-          # - name: aarch64-pc-windows-msvc FIXME: validate on aarch64
-          #   arch: arm64
-          
+        # Exclude WDK 10.0.22621 with ARM64 runner as it doesn't ship certmgr and makecert binaries
```
wmmc88: ditto on the comment

.github/workflows/test.yaml
```diff
@@ -31,48 +37,48 @@ jobs:
           - dev
           - release
 
-        target_triple:
-          - name: x86_64-pc-windows-msvc
-            arch: amd64
-          # - aarch64-pc-windows-msvc FIXME: Add support for executing ARM64 tests. Maybe use target specific test runner in emulator: https://doc.rust-lang.org/cargo/reference/config.html#target
+        # Exclude WDK 10.0.22621 with ARM64 runner as it doesn't ship certmgr and makecert binaries
```
wmmc88: ditto

.github/workflows/test.yaml
```diff
@@ -31,48 +37,48 @@ jobs:
...
           toolchain: ${{ matrix.rust_toolchain }}
-          targets: ${{ matrix.target_triple.name }}
 
       - name: Install Cargo Expand
         uses: taiki-e/install-action@v2
         with:
           tool: cargo-expand@1.0.85
 
       - name: Run Cargo Test
-        run: cargo +${{ matrix.rust_toolchain }} test --locked --profile ${{ matrix.cargo_profile }} --target ${{ matrix.target_triple.name }} --all-features
```
wmmc88, line 83: With this, we are losing the check that tests cross compile successfully

.github/actions/install-winget/action.yaml
krishnakumar4a4: Updating to the latest version and modifying the installation process to use dependencies from the DesktopAppInstaller_Dependencies.zip file helped reduce flakiness in installing and using winget on the ARM64 runner.

Here are some earlier failed pipelines that used various approaches for installation.
1. Failure with older version on Arm64-> https://github.com/microsoft/windows-drivers-rs/actions/runs/16624331689/job/47036494999?pr=420
2. Intermittent failures with Install-Winget powershell script -> https://github.com/microsoft/windows-drivers-rs/actions/runs/16651987020/job/47126910692?pr=420
3. The recommended installation approach from the MSDN documentation works for ARM64 but fails intermittently on the x64 runner -> https://github.com/microsoft/windows-drivers-rs/actions/runs/16649226063/job/47117267716?pr=420

Let me know if anything else needs to be verified for using the latest winget version. I do not know all the reasons why the older version was pinned.

.github/actions/install-llvm/action.yml
krishnakumar4a4: Yes, I have mentioned the same in the resolved discussion here https://github.com/microsoft/windows-drivers-rs/pull/420#discussion_r2255792444

In short, we could modify this step to include a condition to skip installation if the matching LLVM version already installed. That would involve parsing winget output and determine whether if its already installed, which seemed flaky to me. For reference, the output of winget list (which can be used determine if a version is installed) looks like this:
```
PS D:\a\windows-drivers-rs\windows-drivers-rs> winget list --id LLVM.LLVM --exact --accept-source-agreements --disable-interactivity 2>$null                     
Name Id        Version Available Source
---------------------------------------
LLVM LLVM.LLVM 17.0.6  20.1.8    winget
```

Since we’ve pinned LLVM version 17.0.6, which is older than the latest version pre-installed on these runners, we end up installing 17.0.6 every time. Moreover, if we switch to the latest version and pin that, installation will likely be skipped until a newer version is published and the runner image is updated. So, in my view, we’re not losing much by installing LLVM every time, as it doesn’t add any extra runtime—except for not saving time during a brief window when the pinned version matches the pre-installed one.

.github/workflows/build.yaml
krishnakumar4a4: Yes, I can update the comment accordingly. 
Additionally, If we can get the SDK installation to work, then we can even get rid of this entirely.

.github/workflows/test.yaml
krishnakumar4a4: Please refer to my earlier discussion here https://github.com/microsoft/windows-drivers-rs/pull/420#discussion_r2257156850

Quoting the part which talks about the test workflow below
>For test workflow, cross compilation wouldn't make sense as unit tests are generally built as exes to the target architecture (assuming target is different from host), and those binaries can't be run on the host architecture as part of test run. However, there is a provision to make these unit tests run by passing a custom runner configuration in the config.toml matching the target architecture. I believe it is not necessary to explore the custom runner configuration for this PR scope and left it out.

Is there anything else we could do here?

.github/workflows/build.yaml
krishnakumar4a4: Updated the comment.

.github/workflows/local-development-makefile.yaml
krishnakumar4a4: Updated the comment.

.github/workflows/test.yaml
krishnakumar4a4: Updated the comment.

.github/actions/install-winget/action.yaml
```diff
@@ -0,0 +1,78 @@
...
+      shell: pwsh
+      run: |
+        $AuthenticatedHeaders = @{ "Authorization" = "Bearer ${{ inputs.GITHUB_TOKEN }}" }
+
+        # Detect runner architecture
+        $Architecture = if ($env:RUNNER_ARCH -eq "ARM64") { "arm64" } else { "x64" }
+        Write-Host "Runner architecture: $Architecture"
+
+        # winget-cli release v1.11.400
+        $ReleaseInfo = Invoke-RestMethod -Headers $AuthenticatedHeaders 'https://api.github.com/repos/microsoft/winget-cli/releases/226250169'
```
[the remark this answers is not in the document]
krishnakumar4a4: Accepted this change.

.github/workflows/build.yaml
krishnakumar4a4: Installing 10.0.22621 SDK with installer works now. We no longer need this condition and removed it in the latest commit.

.github/actions/install-llvm/action.yml
wmmc88: If you're going to force install every time, then remove the `winget list` line. the idea is to expand this matrix to include more than just one version of llvm. 

>That would involve parsing winget output and determine whether if its already installed, which seemed flaky to me

Yes this is what i used to do and was a little brittle. winget pwsh module does the same thing internally but with much more robustness... its unfortunate you coudlnt get it to work on arm

.github/workflows/test.yaml
wmmc88: i will sometime cross-compile tests and copy them onto different architectures to run them if i dont have the test computer setup for development. that scenario coverage is what we lose with this change. i think it might be fine to lose it in tradeoff for less build minutes-- just pointing out we are reducing what the matrix covers with this change

.github/actions/install-wdk/action.yml
```diff
@@ -22,16 +22,31 @@ runs:
...
         }
+
+        # Function to install or update a WinGet package
+        function Install-WinGetPackageIfNeeded($packageId, $packageName) {
+            if ((Get-WinGetPackage -Id $packageId -Source winget -MatchOption Equals -ErrorAction SilentlyContinue).Id -eq $packageId) {
+                Write-Host "$packageName ($packageId) is already installed. Attempting to update..."
+                Update-WinGetPackage -Id $packageId -Source winget -MatchOption Equals -Mode Silent -Force
+            } else {
+                Write-Host "Installing $packageName ($packageId)..."
+                Install-WinGetPackage -Id $packageId -Source winget -MatchOption Equals -Mode Silent -Force
```
wmmc88: Im a little perplexed here. above in install-llvm you use the winget-cli because the pwsh module is not working in arm, but here you are unconditionally using the poweshell module?

wmmc88: with all the winget powershell modules, you must inspect their output to determine if they succeeded or not. failing install is still considered a success so you need to explicitly check status or installer error code

.github/actions/install-wdk/action.yml
```diff
@@ -130,16 +145,33 @@ runs:
...
+
+            Write-Host "Installing Windows SDK 10.0.22621.5040 from Microsoft download link..."
+            $sdkInstallerUrl = "https://go.microsoft.com/fwlink/?linkid=2311806"
+            $installerPath = "$env:RUNNER_TEMP\winsdksetup.exe"
+
+            Write-Host "Downloading SDK installer from $sdkInstallerUrl..."
+            Invoke-WebRequest -Uri $sdkInstallerUrl -OutFile $installerPath -UseBasicParsing
+
+            Write-Host "Running SDK installer with all features..."
+            Start-Process -FilePath $installerPath -ArgumentList "/quiet", "/features", "+" -Wait -NoNewWindow
```
wmmc88: what does /features do? is it needed?

wmmc88 (commented):
im still seeing if there is something to be done about SDK. I've reached out to folks who own it and they're looking into it. rather than download the exe, i've also got winget to install it properly via --override

.github/workflows/test.yaml
krishnakumar4a4: 1. If I understand correctly, you're referring to using runner configuration for executing cross-compiled tests? I think we can create a separate issue for this if that makes sense.
2. I have one question. For example, if we cross-compile tests to AArch64 on an x64 machine, what additional benefit do we gain if we're already running the same tests on an ARM64 runner? What does cross-compiling the tests actually provide?
3. Just for clarity, the way I see it, the existing matrix has only covered host compilation so far. While it includes a placeholder for cross-compilation, but it was not implemented. So, we're not reducing the current coverage; instead, we're extending it to include test runs on ARM64 as well.

.github/actions/install-wdk/action.yml
krishnakumar4a4: You're right—and it's a bit awkward for me as well. I noticed this change recently; it sneaked into this PR when I rebased with the NuGet PR. I'm fairly certain this was failing earlier on the ARM64 runner. I wanted to verify this by reverting the LLVM installation mechanism and I forgot about it. Let me do it now.

Even the linker issue I mentioned [in this resolved conversation](https://github.com/microsoft/windows-drivers-rs/pull/420#discussion_r2261833703) is no longer reproducible.

.github/actions/install-wdk/action.yml
```diff
@@ -134,16 +145,33 @@ runs:
...
-          Write-Host "Using Winget to install WDK version $inputVersion..."
-          $wdk = "Microsoft.WindowsWDK.$inputVersion"
-          if ((Get-WinGetPackage -Id $wdk -Source winget -MatchOption Equals).Id -eq '$wdk') {
-            Write-Host "$wdk is already installed. Attempting to update..."
-            Update-WinGetPackage -Id $wdk -Source winget -MatchOption Equals -Mode Silent -Force
+          Write-Host "Using Winget to install WDK and SDK version $inputVersion..."
+
+          # Install Windows SDK first
+          if ($sdkVersion -eq "10.0.22621") {
+            # Installing the 10.0.22621 SDK via winget is not working, so it is being installed using the winsdksetup.exe installer
```
gurry: My guess is that the only reason installing the SDK via winget is "working" for versions other than 10.0.22621 is because it's actually already present on the runner and Winget doesn't really end up installing anything. If so, then this comment is a bit misleading.

.github/actions/install-wdk/action.yml
gurry: "/features +" means install all features i.e. all the MSIs that are part of the SDK.

@krishnakumar4a4 How long does the installation of the SDK take. If it takes a while then we might consider selectively installing the features we need rather than everything to speed things up.

krishnakumar4a4: Before this change, time taken by Install WDK step
1. On x64 to install 10.0.22621 WDK is ~1m 30s
2. On x64 to install 10.0.26100 WDK is ~2m 30s
3. On Arm64 to install 10.0.22621 WDK is ~2m
4. On Arm64 to install 10.0.26100 WDK is ~3m

After this change
1. On x64 to install 10.0.22621 SDK & WDK is ~1m 30s
2. On x64 to install 10.0.26100 SDK & WDK is ~3m 48s
3. On Arm64 to install 10.0.22621 SDK & WDK is ~2m
4. On Arm64 to install 10.0.26100 SDK & WDK is ~3m

.github/actions/install-wdk/action.yml
krishnakumar4a4: That’s exactly how it appears, but we can’t be certain as we don’t understand the generic error or its cause as of now.

gurry: The comment is saying winget doesn't work for the 22621 SDK which implies it works for other kits. But we don't know that because the other kit 26100 is already present and winget never attempts to install it. So the comment's claim is wrong or at least not based on evidence.

In any case, all of this is moot given that @wmmc88 has made Winget PS cmdlet great again through his experiements so we won't need this version-specific logic anymore.

.github/actions/install-llvm/action.yml
gurry: Hopefully we can use what we learnt from installing the SDK through the Winget cmdlet and use it to install LLVM too which should resolve this problem.

.github/actions/install-wdk/action.yml
```diff
@@ -130,7 +134,16 @@ runs:
           Write-Output "WindowsSdkBinPath=$packages\Microsoft.Windows.SDK.CPP.$version\c\bin" >> $env:GITHUB_ENV
         } else {
 
-          Write-Host "Using Winget to install WDK version $inputVersion..."
+          Write-Host "Using Winget to install WDK and SDK version $inputVersion..."
+
+          # Install Windows SDK 10.0.22621 explicitly when WDK 10.0.22621 is requested
```
gurry: Comment should explain _why_ we are installing this version explicilty (i.e. because it's not present on the runner)

.github/actions/install-llvm/action.yml
krishnakumar4a4: I have reverted LLVM installation to older approach which uses powershell modules as powershell modules seems to work fine on Arm64 runners.

.github/actions/install-wdk/action.yml
krishnakumar4a4: Using powershell modules for LLVM also works fine now. I suspect there is a change in the Arm64 runner which would have addressed both the powershell exception and [linker exception](https://github.com/microsoft/windows-drivers-rs/pull/420#discussion_r2261833703).

Here is the information of the older Arm64 runner where I have seen problems originally
```
Current runner version: '2.325.0'
Operating System
  Microsoft Windows 11
  10.0.22631
  Enterprise
Runner Image
  Image: windows-11
  Version: 0.0.3
  Included Software: https://github.com/actions/runner-images/blob/win11/0.0/images/windows/Windows11-Arm-Readme.md
  Image Release: https://github.com/actions/runner-images/releases/tag/win11F0.0
Runner Image Provisioner
  2.0.449.1
```

and the latest
```
Current runner version: '2.327.1'
Runner Image Provisioner
  Hosted Compute Agent
  Version: 20250711.363
  Commit: 6785254374ce925a23743850c1cb91912ce5c14c
  Build Date: 2025-07-11T20:04:25Z
VM Image
  - OS: Windows (Arm64)
  - Source: Partner
  - Name: Microsoft Windows 11 Desktop by Arm Limited
  - Version: 20250728.24.1
  - Included Software: https://github.com/actions/partner-runner-images/blob/main/images/arm-windows-11-image.md
```

krishnakumar4a4: I did study about it and understood that using automatic success variable `$?` gives proper success status for powershell modules/scripts. I now added a check using this automatic success variable to both wdk and LLVM installation. Please have a look at it.

.github/actions/install-wdk/action.yml
krishnakumar4a4: With the workaround discovered by @wmmc88, I reverted back to using powershell modules to install 22621 sdk and wdk.

.github/actions/install-wdk/action.yml
krishnakumar4a4: The logic has now been simplified to unconditionally install the SDK and WDK for a couple of reasons. First, the Winget installation is only used to install one older kit—22621—and it's unlikely we'll use Winget to install another older kit. Second, for all newer kits, we prefer the NuGet channel, which is also the direction we're heading with cargo-wdk.

.github/actions/install-wdk/action.yml
krishnakumar4a4: This comment is removed in the latest code. Closing this comment.

.github/actions/install-winget/action.yaml
krishnakumar4a4, line 1: Note: This file is not entirely new. The parent folder renamed from `winget-install` to `install-winget` to maintain consistency with other actions. Hence the inline history is lost.

.github/actions/install-wdk/action.yml
```diff
@@ -130,14 +134,24 @@ runs:
...
+          $sdk = "Microsoft.WindowsSDK.$sdkVersion"
           $wdk = "Microsoft.WindowsWDK.$inputVersion"
-          if ((Get-WinGetPackage -Id $wdk -Source winget -MatchOption Equals).Id -eq '$wdk') {
-            Write-Host "$wdk is already installed. Attempting to update..."
-            Update-WinGetPackage -Id $wdk -Source winget -MatchOption Equals -Mode Silent -Force
-          } else {
-            Write-Host "Installing $wdk..."
-            Install-WinGetPackage -Id $wdk -Source winget -MatchOption Equals -Mode Silent -Force
+          Write-Host "Installing $sdk..."
+          Install-WinGetPackage -Id $sdk -Source winget -MatchOption Equals -Mode Silent -Force -Override '/q /features +'
```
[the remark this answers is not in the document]
krishnakumar4a4: I do not see a trailing space.

.github/actions/install-wdk/action.yml
wmmc88: alright! sounds good. seems like the arm images are being updated fairly quickly rn since its still in beta

.github/workflows/test.yaml
wmmc88: 1. No. i would just cross compile the arm64 tests, and copy them over to an arm64 host manually. 
2. only small benefit is a check that the cross-compilation works
3. ah. ur right, it was commented out

anyways, i think this is fine

.github/actions/install-llvm/action.yml
```diff
@@ -0,0 +1,28 @@
...
+      shell: pwsh
+      run: |
+        if ((Get-WinGetPackage -Id LLVM -Source winget -MatchOption Equals).InstalledVersion -eq '${{ inputs.version }}') {
+          Write-Host "LLVM ${{ inputs.version }} is already installed."
+        } else {
+          Write-Host "Installing LLVM ${{ inputs.version }}..."
+          Install-WinGetPackage -Id LLVM.LLVM -Version ${{ inputs.version }} -Source winget -MatchOption Equals -Mode Silent -Force
+
+          # Check if WinGet installation succeeded
+          $installSuccess = $?
```
wmmc88, line 21: This is not sufficient. Winget may succeed but the installer fails:

```
PS C:\Users\melvinwang> install-WinGetPackage Microsoft.WindowsSDK.10.0.22621

Id                              Name                                                       Source InstallerErrorCode Status       RebootRequired ExtendedErrorCode
--                              ----                                                       ------ ------------------ ------       -------------- -----------------
Microsoft.WindowsSDK.10.0.22621 Windows Software Development Kit - Windows 10.0.22621.2428 winget 2147942487         InstallError False          System.Runtime.InteropServices.COMExcepti…

PS C:\Users\melvinwang> $?
True
```

wmmc88: The same check needs to be made for uninstal-wingetpackage

.github/actions/install-wdk/action.yml
wmmc88: See https://github.com/microsoft/windows-drivers-rs/pull/420/files#r2274086103 

Specifically for Winget Powershell modules, if the winget interface succeeds, then $? will show success. it doesn't speak to the status of the underlying installers that winget calls

.github/actions/install-llvm/action.yml
wmmc88: Actually i think its find to defer this to #456 since i've already done some work to improve the error handling there

wmmc88 (approved):

.github/actions/install-wdk/action.yml
krishnakumar4a4: Yes.

.github/actions/install-wdk/action.yml
```diff
@@ -136,21 +136,22 @@ runs:
...
-          $wdk = "Microsoft.WindowsWDK.$inputVersion"
-          if ((Get-WinGetPackage -Id $wdk -Source winget -MatchOption Equals).Id -eq '$wdk') {
-            Write-Host "$wdk is already installed. Attempting to update..."
-            Update-WinGetPackage -Id $wdk -Source winget -MatchOption Equals -Mode Silent -Force
-          } else {
-            Write-Host "Installing $wdk..."
-            Install-WinGetPackage -Id $wdk -Source winget -MatchOption Equals -Mode Silent -Force
+          Write-Host "Installing $wdk..."
+          Install-WinGetPackage -Id $wdk -Source winget -MatchOption Equals -Mode Silent -Force
+          $installSuccess = $?
```
gurry, line 151: Nit: you don't really need this `$installSuccess` variable because `$?` is a well known variable in PowerShell and people known what it means.

krishnakumar4a4: That's true, but the main reason for adding this variable is that we needed the status code in two places: once for the `if` condition and again to print it using the `Write-Error` log. Its possible to read `$?` twice without creating a new variable, but I'm being extra cautious in case someone adds a new line (like an extra log statement) between the `if` check and the `Write-Error` statement, as `$?` would then reflect the status of the new line instead of `Install-WinGetPackage`. This could be easy to overlook and difficult to catch.

gurry: Sure. That makes sense.

.github/actions/install-llvm/action.yml
krishnakumar4a4: Sure @wmmc88, I will defer changing it from the current PR. 
As you stated, the check I have added only slightly improves the situation (E.g: If an invalid version is given), but will not catch the underlying installer errors.

.github/actions/install-wdk/action.yml
krishnakumar4a4: Understood, deferring further changes as you suggested.
https://github.com/microsoft/windows-drivers-rs/pull/420#discussion_r2276576500

PR #438, opened 2025-08-04
Add stubs for __CxxFrameHandler4 and __GSHandlerCheck_EH4

ladipro:
Certain build environments may use version 4 of EH frame handlers and hit the same issue that prompted adding the `__CxxFrameHandler3` stub. This change is adding two more stubs, `__CxxFrameHandler4` and `__GSHandlerCheck_EH4`, and makes sure they are not folded (have distinct addresses in the resulting binary) by returning distinct values.

None of these functions will ever execute. They are a workaround for a compiler design issue which will take some time to resolve.

### DIFF
--- a/crates/wdk-sys/src/lib.rs
+++ b/crates/wdk-sys/src/lib.rs
@@ -123,17 +123,15 @@ pub static _fltused: () = ();
 // FIXME: Is there any way to avoid these stubs? See https://github.com/rust-lang/rust/issues/101134
 #[cfg(panic = "abort")]
 #[allow(missing_docs)]
-#[allow(clippy::missing_const_for_fn)] // const extern is not yet supported: https://github.com/rust-lang/rust/issues/64926
 #[no_mangle]
-pub extern "system" fn __CxxFrameHandler3() -> i32 {
+pub const extern "system" fn __CxxFrameHandler3() -> i32 {
     0
 }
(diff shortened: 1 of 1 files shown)


ladipro:
cc @gurry

crates/wdk-sys/src/lib.rs
```diff
@@ -129,6 +129,29 @@ pub extern "system" fn __CxxFrameHandler3() -> i32 {
     0
 }
 
+#[cfg(panic = "abort")]
+#[allow(missing_docs)]
+#[allow(clippy::missing_const_for_fn)] // const extern is not yet supported: https://github.com/rust-lang/rust/issues/64926
```
wmmc88: https://github.com/rust-lang/rust/issues/64926 has already been stabilized so all three of these should be able to be `pub const extern "system" fn`

wmmc88 (changes requested):
just one small nit on `const`ifying it since the feature has long since been stabilized. Otherwise LGTM. 

this is blocked on #406 going in first, but that should be resolved shortly

crates/wdk-sys/src/lib.rs
ladipro: Done, thank you.

gurry (approved):

PR #439, opened 2025-08-04
ci: add concurrency control to workflows

### LINKED ISSUE #363: [cargo-wdk] Cancel pipelines on new push
This issue is supposed to handle two things
1. Cancel the previously running pipeline on new push. This will save CI resources.
2. Introduce `continue-on-error=true` for the matrix strategy of test and build actions files. This will help catch errors specific to particular combination of matrix that is not compliant. Not having this today, will cancel all the pipelines if one of the pipeline in the matrix fails essentially hiding possible problems with other pipelines. Note: Point 1 is prerequisite for this change.

svasista-ms:
Closes #363 

- Added concurrency control to cancel in-progress jobs on the same branch
- Added `fail-fast: false` to allow all jobs to complete even when some fail.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,11 +1,11 @@
+name: Build
+
 on:
   push:
   pull_request:
   merge_group:
   schedule: # Trigger a job on default branch at 4AM PST everyday
     - cron: "0 11 * * *"
-
-name: Build
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -9,7 +9,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 env:
   RUSTFLAGS: >-
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -7,6 +7,10 @@ on:
 
 name: Build
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 env:
   RUSTFLAGS: >-
     -D warnings
@@ -17,6 +21,7 @@ jobs:
     name: Build
     runs-on: windows-latest
     strategy:
+      fail-fast: false # Allow all matrix variants to complete even if some fail
       matrix:
         wdk:
           - Microsoft.WindowsWDK.10.0.22621 # NI WDK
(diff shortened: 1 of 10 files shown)


.github/workflows/build.yaml
```diff
@@ -7,6 +7,10 @@ on:
 
 name: Build
 
+concurrency:
```
gurry, line 10: I think we should try to cancel all workflows on a new push not just build and test.

wmmc88: i agree it should be on **all** workflows

svasista-ms: Added concurrency control to all workflows. New pushes on the same branch will cancel all in-progress workflows.

svasista-ms:
> Could you trigger some pipelines to ensure this is working as expected?

I triggered 2 workflows with the following commits:
- Merge remote-tracking branch 'origin/main' into ci-concurrency-control (2afb3e8)
- test concurrency control (e1cc56a)

In-progress workflows were cancelled when a new commit was pushed. New pushes cancel workflows/jobs which are in both `Queued` and `In Progress` states. This ensures that all old workflows/jobs are cancelled upon a new event trigger.

SS before pushing `test concurrency control (e1cc56a)`:
before

After the push:
after

.github/workflows/build.yaml
```diff
@@ -1,11 +1,15 @@
...
   push:
   pull_request:
   merge_group:
   schedule: # Trigger a job on default branch at 4AM PST everyday
     - cron: "0 11 * * *"
 
-name: Build
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
wmmc88: https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency#example-only-cancel-in-progress-jobs-on-specific-branches

We probably don't want CI runs on main to be cancelled when we have many prs merging in quick succession

.github/workflows/code-formatting-check.yaml
```diff
@@ -1,11 +1,16 @@
 name: Code Formatting Check
```
wmmc88, line 1: All workflows that have cancel-in-progress enables should also have fail-fast disabled

wmmc88 (changes requested):

.github/workflows/code-formatting-check.yaml
svasista-ms: According to the [GitHub Actions documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstrategyfail-fast), `fail-fast` cannot be applied as it's a property of the strategy section. The independent jobs in `code-formatting-check` workflow already run to completion regardless of other job failures, which achieves the same goal.

.github/workflows/build.yaml
svasista-ms: Thanks for calling this out @wmmc88. Excluded the `main` from `cancel-in-progress`

krishnakumar4a4 (approved):

PR #450, opened 2025-08-09
Add WSK (via Network API subset)

GameGuyThrowaway:
This PR adds in bindings for the Winsock Kernel (WSK) Network Programming Interface. In adding it in, it seemed best fitting to add the Network API subset, though I haven't implemented the rest of its components yet.

One current issue is that it doesn't compile because of a mismatch on INET_PORT_RESERVATION's size.
```cpp
typedef struct {
#ifdef __cplusplus
    INET_PORT_RESERVATION Reservation;
    INET_PORT_RESERVATION_TOKEN Token;
#else
    INET_PORT_RESERVATION;
    INET_PORT_RESERVATION_TOKEN;
#endif
} INET_PORT_RESERVATION_INSTANCE, *PINET_PORT_RESERVATION_INSTANCE;
```
```rust
#[repr(C)]
#[derive(Debug, Default, Copy, Clone)]
pub struct INET_PORT_RESERVATION_INSTANCE {
    pub _address: u8,
}
#[allow(clippy::unnecessary_operation, clippy::identity_op)]
const _: () = {
    [
        "Size of INET_PORT_RESERVATION_INSTANCE",
    ][::core::mem::size_of::<INET_PORT_RESERVATION_INSTANCE>() - 16usize];
    [
        "Alignment of INET_PORT_RESERVATION_INSTANCE",
    ][::core::mem::align_of::<INET_PORT_RESERVATION_INSTANCE>() - 8usize];
};
```
I assume this is because of the preprocessor directive, but the bindings define the struct as being 1 byte in size, while expecting it to be 16 bytes in size (as the original is). I'm not super familiar with using bindgen, and if this is either a local issue on my machine, or an upstream bug, and how the maintainers here would prefer to override the definition (if possible).

Another potential issue is the fact that the rest of the Network API subset hasn't been included in this PR. The 80 other headers are out of scope for my usage, and I'm not extremely familiar with determining whether a header is for KMDF vs WDM vs UMDF. If someone else wants to pick this up, I'd appreciate it. Otherwise, if it's required in order to be merged, I can try to find some time to implement the rest.

### DIFF
--- a/crates/wdk-build/src/lib.rs
+++ b/crates/wdk-build/src/lib.rs
@@ -270,6 +270,10 @@ pub enum ApiSubset {
     Storage,
     /// API subset for USB (Universal Serial Bus) drivers: <https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/_usbref/>
     Usb,
+    /// API subset for Network drivers <https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/_netvista/>
+    ///
+    /// Currently only supports the WSK (Winsock Kernel) portion of the Network API subset.
+    Network,
 }
 
 impl Default for Config {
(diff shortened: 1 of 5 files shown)


GameGuyThrowaway:

PR #451, opened 2025-08-09
feat: enhance cargo metadata parsing to respect config.toml

wmmc88:
This pull request improves how the build system and the `wdk-build` crate handle Cargo workspace configuration, particularly with respect to custom `config.toml` files and patched dependencies. It also adds a new test workspace to verify these behaviors and updates the CI workflows to build and test this new setup.

**Build system and metadata resolution improvements:**

* Updated `wdk-build` to always run `cargo_metadata` from the top-level Cargo workspace directory, ensuring that custom `.cargo/config.toml` (including `[patch]` tables) are respected for dependency resolution. This includes adding debug-level tracing and more robust path handling in `bindgen.rs` and `lib.rs`. [[1]](diffhunk://#diff-6068e3322dcfcfd64e4b10bb9ddaa00820ea46e0c62b3ecfeacb5bb10af24e6aR11-R13) [[2]](diffhunk://#diff-6068e3322dcfcfd64e4b10bb9ddaa00820ea46e0c62b3ecfeacb5bb10af24e6aR246-R259) [[3]](diffhunk://#diff-43524f733cccc1a175c3bf83309866acb53b8cadb1628478ca1c87d31dda2b26R18) [[4]](diffhunk://#diff-43524f733cccc1a175c3bf83309866acb53b8cadb1628478ca1c87d31dda2b26R312-R330) [[5]](diffhunk://#diff-43524f733cccc1a175c3bf83309866acb53b8cadb1628478ca1c87d31dda2b26L323-R339)

**Continuous Integration (CI) enhancements:**

* Modified the GitHub Actions workflows to build and package the new test workspace and to run its tests using `cargo make`, ensuring the new configuration is tested in CI. [[1]](diffhunk://#diff-d0777657fa3fd81d23aaf7273e58aee453b04e67882517900c56daeef9b3e4c1R96-R101) [[2]](diffhunk://#diff-245392b692a50c38ecab4381b118862db514035c10983f3bd4f4b7f1f4be4692L77-R81)

**Testing for custom config.toml and patched dependencies:**

* Added a new test workspace `customized-config-toml-workspace` under `tests/`, which demonstrates and verifies that `[patch.crates-io]` entries in `.cargo/config.toml` are honored during build and metadata resolution. This includes a patched dependency, a test crate, and a build script that exercises the new logic. [[1]](diffhunk://#diff-f961035d34d867319408a7f32bbf78b28d2a2ad7285802f9f4feae2f347138cfR9) [[2]](diffhunk://#diff-d39ab5f93e105eb9ddc62301e8abaf817094770760874165b43d4b5751f982ffR1-R5) [[3]](diffhunk://#diff-8e784deb1526efe3f01b53c1a1525dc4782419517888d02c28eac4bf9d757c36R1-R3) [[4]](diffhunk://#diff-6b488af911d9780494f5ea4ffe29fa0ec9fcbe750dae69a8ef7e6f19b2dce4faR1-R14) [[5]](diffhunk://#diff-83bf29bef45bdc5b1a1f084c525d04bca15119c753818093afa6e919bf68e735R1-R6) [[6]](diffhunk://#diff-6526d86b07062ce2edc4a6383f3ae111979884c7748c0bcd1a13bc255e5d6baaR1-R5) [[7]](diffhunk://#diff-5378a2f57373ffb54465d68dd1bd7c834a159115cf2c6c31cc7b78cd7478de23R1-R4) [[8]](diffhunk://#diff-6b4540e162ba82dbbfafbd58480e75b531dbadffd2b8da3cb4a8ff91255fb759R1-R3)Improve handling of cargo metadata to respect `config.toml` configurations and add build and package steps for tests in GitHub workflows. Introduce a new library crate and a workspace to facilitate these changes.


crates/wdk-build/src/bindgen.rs
```diff
@@ -242,10 +243,20 @@ fn get_stable_rust_target() -> Result<bindgen::RustTarget, ConfigError> {
 // Returns `ConfigError::CargoMetadataPackageNotFound` if the `wdk-build`
 // package is not found, or `ConfigError::UnsupportedRustEdition` if the edition
 // is not supported.
+#[tracing::instrument(level = "debug")]
```
krishnakumar4a4, line 246: Shouldn't we take this opportunity to add instrumentation for other methods in this module like `get_rust-target`?

crates/wdk-build/src/lib.rs
```diff
@@ -308,10 +309,25 @@ impl Config {
     /// # Panics
     ///
     /// Panics if the resolved top-level Cargo manifest path is not valid UTF-8
+    #[tracing::instrument(level = "debug")]
```
krishnakumar4a4, line 363: how about adding instrumentation for all other the public functions in this module?

tests/mixed-package-kmdf-workspace/Cargo.toml
```diff
@@ -9,12 +9,12 @@ kmdf-version-major = 1
 target-kmdf-version-minor = 33
 
 [workspace.dependencies]
-wdk = { path = "../../crates/wdk", version = "0.3.1" }
-wdk-alloc = { path = "../../crates/wdk-alloc", version = "0.3.1" }
-wdk-build = { path = "../../crates/wdk-build", version = "0.4.0" }
-wdk-macros = { path = "../../crates/wdk-macros", version = "0.4.0" }
-wdk-panic = { path = "../../crates/wdk-panic", version = "0.3.1" }
-wdk-sys = { path = "../../crates/wdk-sys", version = "0.4.0" }
+wdk = { path = "../../crates/wdk" }
```
krishnakumar4a4: This is better now. Could you make this change in `tests/umdf-driver-workspace` and `examples/*` as well, for consistency?

I wonder why both version and path have existed for dependencies in some tests and examples until now.

If the default strategy was to build tests using the latest commit of the crates, then just having the path is clearer. Although including version is still valid and acts as an additional check, it’s redundant.

As discussed earlier, we could open a separate issue to set up a pipeline that runs the same tests and examples using the last published version of these crates.

gurry: Having both `path` and `version` ensures that when you build locally then you depend through the local path, but if you publish this crate then you depend through crates.io.

Since we never publish the test crates, `version` in this case was unnecessary.

tests/mixed-package-kmdf-workspace/crates/driver/Cargo.toml
```diff
@@ -9,6 +9,8 @@ publish = false
 
 [lib]
 crate-type = ["cdylib"]
+# Windows Server 2022 (used in CI) has issues running driver tests
```
gurry: Just curious, what were the issues?

gurry (approved):

tests/mixed-package-kmdf-workspace/Cargo.toml
wmmc88: @gurry is exactly right. ive also updated the pr for the rest of the tests folder

crates/wdk-build/src/lib.rs
wmmc88: I will address this in a separate pr for #454 since it involves some other non-related changes (ex. some of these args require impl fmt::Debug)

crates/wdk-build/src/bindgen.rs
wmmc88: https://github.com/microsoft/windows-drivers-rs/issues/454

tests/mixed-package-kmdf-workspace/crates/driver/Cargo.toml
wmmc88: The failure looks like this: https://github.com/microsoft/windows-drivers-rs/actions/runs/16865937018/job/47772837088#step:10:2027

happens regardless of if its in cargo make or just natively cargo test

gurry (dismissed):

tests/mixed-package-kmdf-workspace/Cargo.toml
krishnakumar4a4: Just reiterating a couple of things in case this missed your attention. 

1. We have version along with path in the `examples` as well. Shouldn't we make this change there as well?
2. Also, Shouldn't we open a separate issue to run specific (ones which simulate dev workflow and require backward compat) integration tests & examples against latest published versions of the crates as an additional pipeline?

wmmc88: 1. Yes I'll remove
2. I'm not sure what scenarios you suggest we test for here. These test should ways only run against the current state of the repo. Previously released versions have their own set of tests

wmmc88: To clarify, my response to 2 is wrt to top level tests folder. Cargo wdk is a different story since it's no longer tightly coupled and that requires its own testing strategy

gurry: > 1. We have version along with path in the `examples` as well. Shouldn't we make this change there as well?

Yes, I believe we should.

> 2. Also, Shouldn't we open a separate issue to run specific (ones which simulate dev workflow and require backward compat) integration tests & examples against latest published versions of the crates as an additional pipeline?

Crates once published don't change so there is no need to test them. However, if your goal is to test something else, like cargo-wdk, that _uses_ the published crates then such a pipline would certainly be useful.

krishnakumar4a4: Yes @gurry, I am referring to the following cases
1. Integration tests within the cargo-wdk crate, where we invoke cargo-wdk to build test driver setups: https://github.com/krishnakumar4a4/windows-drivers-rs/blob/68a72b2b1e3cc94eab20644ad28fa0b6839f1cf1/crates/cargo-wdk/tests/build_command_test.rs#L17
2. Examples and tests in the root folder, which are built using cargo-wdk as part of the existing workflow: https://github.com/krishnakumar4a4/windows-drivers-rs/blob/68a72b2b1e3cc94eab20644ad28fa0b6839f1cf1/.github/workflows/build.yaml#L106 & https://github.com/krishnakumar4a4/windows-drivers-rs/blob/68a72b2b1e3cc94eab20644ad28fa0b6839f1cf1/.github/workflows/build.yaml#L110

The existing pipeline only validates the latest cargo-wdk commit with all the tests and examples, which point to the latest crate revisions. However, there is no pipeline that checks the latest cargo-wdk against the same tests and examples using the last released versions of the crates.
I was referring to having a pipeline or setup for that, to ensure cargo-wdk builds remain backward compatible with tests and examples using released crate versions.

Just flagging the need for this pipeline—can open a separate issue if we’re on the same page.

krishnakumar4a4 (dismissed):

gurry (approved):

krishnakumar4a4 (approved):

wmmc88:
blocked on #501

crates/cargo-wdk/src/actions/build/build_task.rs
```diff
@@ -100,7 +102,10 @@ impl<'a> BuildTask<'a> {
             .iter()
             .map(std::string::String::as_str)
             .collect::<Vec<&str>>();
-        self.command_exec.run("cargo", &args, None)?;
+        // Run cargo build from the provided working directory so that workspace
+        // resolution and relative paths behave as expected for the selected package
```
wmmc88:
```suggestion
        // Run cargo build from the provided working directory so that config.toml
        // is respected
```

gurry (approved):

krishnakumar4a4 (approved):

PR #462, opened 2025-08-14
refactor: use `std::path::absolute` instead of canonicalize + strip_extended_path_prefix

### LINKED ISSUE #367: Replace uses of `strip_extended_path_prefix` with `path::absolute`
This module is meant for private utils contained for use within this crate. It looks all that's used in `PathExt` and `StripExtendedPathPrefixError`. I think both those things are obsolete now that `https://doc.rust-lang.org/std/path/fn.absolute.html` is available on stable. I originally wrote that code before it was stabilized. 

All uses of `PathExt` because of canonicalize should be changed to use `path::absolute` and other stdlib functions. The `utils` module should also be returned to being private.

This is a `wdk-build` release blocker.

_Originally posted by @wmmc88 in https://github.com/microsoft/windows-drivers-rs/pull/306#discussion_r2122199203_

svasista-ms:
Closes #367 and #479 

We previously combined `std::fs::canonicalize` with a custom strip_extended_path_prefix helper to normalize Windows paths and remove the extended (verbatim) prefix (\?). Issue #367 tracked replacing this logic with a standard library alternative, `std::path::absolute`.

Motivation
1. Avoid unnecessary filesystem resolution (canonicalize) that can fail or touch the disk when simple normalization is enough.
2. Rely on stable (std::path::absolute) for syntactic absolutization
3. Reduce maintenance of custom path utilities.

User Impact:
Slight behavior change in `cargo wdk new`: verbatim (extended) paths now return a clear error instead of being silently normalized. (Issue #481 was created to add support for verbatim, extended paths)

### DIFF
--- a/crates/cargo-wdk/src/actions/build/build_task.rs
+++ b/crates/cargo-wdk/src/actions/build/build_task.rs
@@ -135,6 +135,13 @@ mod tests {
         assert_eq!(build_task.profile, Some(&profile));
         assert_eq!(build_task.target_arch, target_arch);
         assert_eq!(build_task.manifest_path, working_dir.join("Cargo.toml"));
+        assert_eq!(
+            std::ptr::from_ref(build_task.command_exec),
+            &raw const command_exec,
+            "CommandExec instances are not the same"
+        );
+        // TODO: Add assert for verbosity_level once `clap-verbosity-flag` crate
+        // is updated to 3.0.4
     }
 
     #[test]
--- a/crates/cargo-wdk/src/actions/build/error.rs
+++ b/crates/cargo-wdk/src/actions/build/error.rs
@@ -11,8 +11,8 @@ use crate::providers::error::{CommandError, FileError};
 /// Errors for the build action layer
 #[derive(Error, Debug)]
 pub enum BuildActionError {
-    #[error(transparent)]
-    Io(#[from] io::Error),
+    #[error("Provided path is not absolute: {0}")]
+    NotAbsolute(PathBuf, #[source] io::Error),
     #[error(transparent)]
     WdkBuildConfig(#[from] wdk_build::ConfigError),
     #[error("Error Parsing Cargo.toml, not a valid rust project/workspace")]
--- a/crates/cargo-wdk/src/actions/build/mod.rs
+++ b/crates/cargo-wdk/src/actions/build/mod.rs
@@ -225,8 +225,14 @@ impl<'a> BuildAction<'a> {
         let cargo_metadata = &self.get_cargo_metadata(working_dir)?;
         let wdk_metadata = Wdk::try_from(cargo_metadata);
         let workspace_packages = cargo_metadata.workspace_packages();
-        let workspace_root = absolute(cargo_metadata.workspace_root.as_std_path())?;
-        let target_directory = absolute(cargo_metadata.target_directory.as_std_path())?;
+        let workspace_root =
+            absolute(cargo_metadata.workspace_root.as_std_path()).map_err(|e| {
+                BuildActionError::NotAbsolute(cargo_metadata.workspace_root.clone().into(), e)
+            })?;
+        let target_directory =
+            absolute(cargo_metadata.target_directory.as_std_path()).map_err(|e| {
+                BuildActionError::NotAbsolute(cargo_metadata.target_directory.clone().into(), e)
+            })?;
         if workspace_root.eq(&working_dir) {
             // If the working directory is root of a standalone project or a
             // workspace
@@ -242,7 +248,8 @@ impl<'a> BuildAction<'a> {
                     .expect("Unable to find package path from Cargo manifest path")
                     .into();
 
-                let package_root_path = absolute(package_root_path.as_path())?;
+                let package_root_path = absolute(package_root_path.as_path())
+                    .map_err(|e| BuildActionError::NotAbsolute(package_root_path.clone(), e))?;
                 debug!(
                     "Processing workspace member package: {}",
                     package_root_path.display()


crates/cargo-wdk/src/actions/build/build_task.rs
```diff
@@ -30,47 +29,35 @@ pub struct BuildTask<'a> {
...
-    /// * `command_exec` - The command execution provider
-    /// * `fs` - The file system provider
-    /// # Returns
-    /// * `Result<Self, BuildTaskError>` - A result containing the new instance
-    ///   of `BuildTask` or an error
-    /// # Errors
-    /// * `BuildTaskError::CanonicalizeManifestPath` - If there is an IO error
-    ///   while canonicalizing the working dir
-    /// * `BuildTaskError::EmptyManifestPath` - If the manifest path is empty
+    /// Factory method for `BuildTask`.
```
gurry: I feel this is an unnecessary change. This is not a factory method strictly speaking

crates/cargo-wdk/src/actions/build/mod.rs
```diff
@@ -57,31 +57,27 @@ pub struct BuildAction<'a> {
...
-    /// * `command_exec` - The command execution provider instance
-    /// * `fs` - The file system provider instance
-    /// * `metadata` - The metadata provider instance
-    /// # Returns
-    /// * `Result<Self>` - A result containing the new instance of `BuildAction`
-    ///   or an error
-    /// # Errors
-    /// * `BuildActionError::IoError` - If there is an IO error while
-    ///   canonicalizing the working dir
+    /// Factory method for `BuildAction`.
```
gurry: No need to change the description

crates/cargo-wdk/src/actions/build/build_task.rs
```diff
@@ -30,47 +29,35 @@ pub struct BuildTask<'a> {
...
+    ///
+    /// Arguments:
```
gurry:
```suggestion
    /// # Arguments
```

crates/cargo-wdk/src/actions/build/mod.rs
```diff
@@ -57,31 +57,27 @@ pub struct BuildAction<'a> {
...
+    ///
+    /// Args:
```
gurry:
```suggestion
    /// # Arguments
```

crates/cargo-wdk/src/actions/build/mod.rs
```diff
@@ -57,31 +57,27 @@ pub struct BuildAction<'a> {
...
+    /// * `params` – build inputs
+    /// * `wdk_build` – WDK helper
+    /// * `command_exec` – command runner
+    /// * `fs` – filesystem provider
+    /// * `metadata` – cargo metadata provider
```
gurry: There is no need to change these descriptions. The existing ones were perfectly fine, In fact they were better because they weren't this terse.

crates/cargo-wdk/src/actions/build/mod.rs
```diff
@@ -127,21 +123,22 @@ impl<'a> BuildAction<'a> {
             "Initialized build for project at: {}",
             self.working_dir.display()
         );
-        wdk_build::cargo_make::setup_path()?;
-        debug!("PATH env variable is set with WDK bin and tools paths");
         let build_number = self.wdk_build.detect_wdk_build_number()?;
         debug!("WDK build number: {}", build_number);
+        wdk_build::cargo_make::setup_path()?;
+        debug!("PATH env variable is set with WDK bin and tools paths");
+        let working_dir = absolute(&self.working_dir)?;
```
gurry: Better to call `absolute` in `new` before creating the object instead of here, just like we were doing with `canonicalize`. `absolute` is okay to call in a constructor because it does not touch the file system.

crates/cargo-wdk/src/actions/build/mod.rs
```diff
@@ -209,25 +206,23 @@ impl<'a> BuildAction<'a> {
...
         let cargo_metadata = &self.get_cargo_metadata(working_dir)?;
-        let target_directory = cargo_metadata.target_directory.as_std_path().to_path_buf();
         let wdk_metadata = Wdk::try_from(cargo_metadata);
         let workspace_packages = cargo_metadata.workspace_packages();
-        let workspace_root = self
-            .fs
-            .canonicalize_path(cargo_metadata.workspace_root.clone().as_std_path())?;
-        if workspace_root.eq(working_dir) {
+        let workspace_root = absolute(cargo_metadata.workspace_root.as_std_path())?;
+        let target_directory = absolute(cargo_metadata.target_directory.as_std_path())?;
```
gurry: We weren't calling `canonicalize` on `target_director` earlier. Will calling `absolute` now  break something?

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -72,7 +72,7 @@ pub struct PackageTask<'a> {
 }
 
 impl<'a> PackageTask<'a> {
-    /// Creates a new instance of `PackageTask`.
+    /// Factory method for `PackageTask`.
```
gurry: This is an unnecessary change

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -37,7 +37,7 @@ pub struct NewAction<'a> {
 }
 
 impl<'a> NewAction<'a> {
-    /// Creates a new instance of `NewAction`.
+    /// Factory method for `NewAction`.
```
gurry: Not needed

crates/cargo-wdk/src/cli.rs
```diff
@@ -135,6 +135,15 @@ impl Cli {
 
         match self.sub_cmd {
             Subcmd::New(cli_args) => {
+                // TODO: Support extended path as cargo supports it
+                if let Some(path) = &cli_args.path {
+                    if path.as_os_str().to_string_lossy().starts_with(r"\\?\") {
+                        return Err(anyhow::anyhow!(
+                            "Extended (verbatim prefixed) paths are currently not supported"
```
```suggestion
                            "Extended (verbatim) paths are currently not supported"
```

crates/cargo-wdk/src/cli.rs
```diff
@@ -350,4 +359,27 @@ mod tests {
...
+                path: Some(PathBuf::from(r"\\?\C:\some\path")),
+            }),
+            verbose: clap_verbosity_flag::Verbosity::default(),
+        };
+
+        let result = cli.run();
+        assert!(result.is_err());
+        assert_eq!(
+            result.err().unwrap().to_string(),
+            "Extended (verbatim prefixed) paths are currently not supported"
```
```suggestion
            "Extended (verbatim) paths are currently not supported"
```

gurry (approved):
Thanks for working on this @svasista-ms.

I want to suggest that it's better to avoid unrelated changes or those not directly related to the PR.

crates/cargo-wdk/src/actions/build/mod.rs
svasista-ms: It did not break anything. The path in `cargo_metadata.target_directory` seems to be absolute already. 
I checked the contents of `cargo_metadata` and it looks like they are absolute paths. 

`
.
"target_directory":"C:\\Users\\svasista\\Documents\\Rust\\windows-drivers-rs\\examples\\sample-kmdf-driver\\target",
.
"workspace_root":"C:\\Users\\svasista\\Documents\\Rust\\windows-drivers-rs\\examples\\sample-kmdf-driver"
.
`

These 2 calls to `absolute` (L223-224) may be redundant. Do you think I should remove these? (or replace it with a `debug_assert!`)

Regarding `workspace_root `, it had to be canonicalized for comparison with `working_dir`. Now they are both absolute.

crates/cargo-wdk/src/actions/build/build_task.rs
svasista-ms: Thank you for clarifying @gurry. I had a major misunderstanding regarding `factory` methods. Now it is clear. I'll revert these docs.

crates/wdk-build/src/lib.rs
```diff
@@ -19,18 +19,64 @@ use metadata::TryFromCargoMetadataError;
...
+    detect_wdk_content_root,
+    detect_windows_sdk_version,
+    get_wdk_version_number,
+    validate_wdk_version_format,
+};
 
 mod bindgen;
 
-use std::{env, path::PathBuf, sync::LazyLock};
+use std::{
```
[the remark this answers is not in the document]
svasista-ms: `use std::{fmt, str::FromStr};` present in L14

crates/cargo-wdk/src/actions/build/mod.rs
```diff
@@ -57,31 +57,28 @@ pub struct BuildAction<'a> {
...
     /// * `metadata` - The metadata provider instance
-    /// # Returns
-    /// * `Result<Self>` - A result containing the new instance of `BuildAction`
-    ///   or an error
-    /// # Errors
-    /// * `BuildActionError::IoError` - If there is an IO error while
-    ///   canonicalizing the working dir
+    ///
+    /// Returns:
+    /// * `Self` – a new instance of `BuildAction`.
```
gurry: The method returns `Result<Self>` so this needs to change. Will also need to bring back the `# Errors` section

crates/cargo-wdk/src/actions/build/mod.rs
gurry: Makes sense. Even if the `absolute` call on `target_directory` is redundant, let's keep it for consistency.

crates/wdk-build/src/lib.rs
```diff
@@ -19,18 +19,64 @@ use metadata::TryFromCargoMetadataError;
 pub mod cargo_make;
 pub mod metadata;
 
-pub mod utils;
+mod utils;
+pub use utils::{
+    detect_wdk_content_root,
+    detect_windows_sdk_version,
+    get_wdk_version_number,
+    validate_wdk_version_format,
```
wmmc88: Not to be addressed in this PR, but these functions were never meant to be exported either. They are meant to be internal implementation details so that we have flexibility on changing/improving them etc. It looks like they were made public to be accessed by cargo-wdk. 

the main `wdk_build` config struct already automatically derives the content root. public accessors should just be added for that and the other properties like SDK version, WDK version, etc. `validate_wdk_version_format` shouldn't even exist as it should probably just be a fromstr conversion on a NewType for WDK version

crates/wdk-build/src/lib.rs
```diff
@@ -1446,6 +1452,215 @@ mod tests {
 
     use super::*;
 
+    // Tests for TwoPartVersion and TwoPartVersionError moved from utils.rs now that
+    // the type is defined in lib.rs (public API).
```
wmmc88: Commentary on the diff shouldn't be left as comments in the code.. Copilot generated?

svasista-ms: Yes, removed. Ty!

crates/wdk-build/src/lib.rs
```diff
@@ -19,18 +19,25 @@ use metadata::TryFromCargoMetadataError;
...
+    detect_wdk_content_root,
+    detect_windows_sdk_version,
+    get_wdk_version_number,
+    validate_wdk_version_format,
+};
 
 mod bindgen;
 
-use std::{env, path::PathBuf, sync::LazyLock};
+use std::{
```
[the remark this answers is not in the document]
svasista-ms: `str::FromStr` was being imported on a different line. There were 2 places where imports from `std` lib were used. In the recent commit, I combined them so that all `std` lib imports are together. Not entirely sure why Copilot couldn't see the import. Maybe because of `#![cfg_attr(nightly_toolchain, feature(assert_matches))]` above the import statement?

crates/cargo-wdk/src/actions/build/mod.rs
svasista-ms: Thank you for catching this @gurry. Addressed this in the latest commit.

crates/cargo-wdk/src/actions/build/mod.rs
```diff
@@ -67,8 +67,13 @@ impl<'a> BuildAction<'a> {
...
     ///
-    /// Returns:
-    /// * `Self` – a new instance of `BuildAction`.
+    /// # Returns
+    /// * `Result<Self>` - A result containing either a new instance of
+    ///   `BuildAction` on success, or an `anyhow::Error`.
+    ///
+    /// # Errors
+    /// This function will return an error if the `absolute` path conversion
+    /// fails.
```
gurry:
```suggestion
    /// * [`anyhow::Error`] -  If `params.working_dir` is not a syntactically valid path, e.g. it is empty
```

gurry: The caller doesn't know that we're calling `absolute` internally. The comment should be in terms they understand.

gurry (commented):
Just a couple more comments. After that I'll approve

crates/cargo-wdk/src/actions/build/mod.rs
svasista-ms: @gurry , fixed the comment. Please review. Thank you!

crates/wdk-build/src/lib.rs
gurry: @svasista-ms please open an issue to track this.

crates/cargo-wdk/src/actions/build/mod.rs
gurry:
```text
 /// Returns an error if the provided working directory path cannot be
 resolved to an absolute path using `std::path::absolute`.
```
This description is still wrong. It's just a different phrasing of the the earlier version. The point is that we should not mention `absolute` in the comment becuase it's an internal detail. I'd recommend using the description I suggested above.

svasista-ms: Oops, totally missed your suggestion. Fixed it.

gurry (dismissed):

krishnakumar4a4 (approved):

crates/wdk-build/src/lib.rs
svasista-ms: Created #479

crates/cargo-wdk/src/actions/build/build_task.rs
```diff
@@ -30,47 +29,35 @@ pub struct BuildTask<'a> {
...
-        let manifest_path = match manifest_path.strip_extended_length_path_prefix() {
-            Ok(path) => path,
-            Err(StripExtendedPathPrefixError::NoExtendedPathPrefix) => manifest_path,
-            Err(StripExtendedPathPrefixError::EmptyPath) => {
-                return Err(BuildTaskError::EmptyManifestPath);
-            }
-        };
-        Ok(Self {
+    ) -> Self {
+        debug_assert!(working_dir.is_absolute(), "working_dir should be absolute");
```
wmmc88: why only assert in debug mode?

crates/cargo-wdk/src/actions/build/build_task.rs
```diff
@@ -30,47 +29,35 @@ pub struct BuildTask<'a> {
...
+        Self {
             package_name,
             profile,
             target_arch,
             verbosity_level,
-            manifest_path,
+            manifest_path: working_dir.join("Cargo.toml"),
             command_exec,
-        })
+        }
```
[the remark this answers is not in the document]
wmmc88: I 2nd this. what was the motivation to turn this into a non-result return?

crates/cargo-wdk/src/actions/build/error.rs
```diff
@@ -11,6 +11,8 @@ use crate::providers::error::{CommandError, FileError};
 /// Errors for the build action layer
 #[derive(Error, Debug)]
 pub enum BuildActionError {
+    #[error(transparent)]
+    Io(#[from] io::Error),
```
wmmc88: what sort of io errors does this return? is it solely from paths? you might want to consider https://github.com/microsoft/windows-drivers-rs/pull/480

wmmc88: not blocking pr. just something to consider to followup

wmmc88:
@svasista-ms  please use conventional commit as the pr title, and add a description (if you think it needs more beyond the title)

crates/cargo-wdk/src/cli.rs
```diff
@@ -154,7 +163,7 @@ impl Cli {
                         Self::detect_default_target_arch_using_rustc(&command_exec)?;
                     TargetArch::Default(detected_arch)
                 };
-                let build_action = BuildAction::new(
+                BuildAction::new(
```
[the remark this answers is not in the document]
wmmc88: eh. i disagree with copilot here. the new pattern is fine

crates/cargo-wdk/src/cli.rs
```diff
@@ -135,6 +135,20 @@ impl Cli {
 
         match self.sub_cmd {
             Subcmd::New(cli_args) => {
+                // TODO: Support extended path as cargo supports it
```
wmmc88, line 138: track in an issue

crates/cargo-wdk/src/cli.rs
```diff
@@ -135,6 +135,20 @@ impl Cli {
...
+                if let Some(path) = &cli_args.path {
+                    const EXTENDED_PATH_PREFIX: &str = r"\\?\";
+                    if path
+                        .as_os_str()
+                        .to_string_lossy()
+                        .starts_with(EXTENDED_PATH_PREFIX)
+                    {
+                        return Err(anyhow::anyhow!(
+                            "Extended (verbatim) paths are currently not supported"
```
wmmc88: may be worth saying that something like:

> Extended/Verbatim paths (ie. paths starting with '\\?\') are not currently supported

because not everyone knows what an extended or verbatim path is. also error should probably include the path

crates/cargo-wdk/src/actions/build/mod.rs
```diff
@@ -57,31 +57,33 @@ pub struct BuildAction<'a> {
...
     pub fn new(
         params: &BuildActionParams<'a>,
         wdk_build: &'a WdkBuild,
         command_exec: &'a CommandExec,
         fs: &'a Fs,
         metadata: &'a Metadata,
     ) -> Result<Self> {
-        // TODO: validate and init attrs here
-        let working_dir = fs.canonicalize_path(params.working_dir)?;
+        // TODO: validate params
```
wmmc88, line 84: track this in an issue assigned to somebody so that we don't lose track of these

wmmc88 (dismissed):
just a couple questions + fix your pr title

crates/cargo-wdk/src/actions/build/build_task.rs
svasista-ms: The `working_dir` passed here is converted to absolute in `BuildAction::new()`. AFAIU, there are no modifying calls in between. But during dev just to make sure nothing happened between creation of `BuildAction` and `BuildTask` I added a debug assert. Adding a regular check and returning an error would require the return type to be `Result`. Do you think it is a good idea to have this check? Also, what is your opinion on returning `Result` from constructor-type functions that do not have fallible calls in them?

crates/cargo-wdk/src/actions/build/build_task.rs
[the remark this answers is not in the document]
svasista-ms: Since the new `BuildTask::new()` method did not have any fallible calls, I changed it to a non-result return.

crates/cargo-wdk/src/actions/build/error.rs
svasista-ms: I think it is from paths, for ex empty path returns an `io::ErrorKind::InvalidInput`. Thanks for sharing #480 . I will take a look.

crates/cargo-wdk/src/cli.rs
svasista-ms: Opened #481 to track this

crates/cargo-wdk/src/actions/build/mod.rs
svasista-ms: Done. I opened #482 and assigned it to myself. `Clap` is already doing some validation, we need to close any gaps it may have.

crates/cargo-wdk/src/actions/build/build_task.rs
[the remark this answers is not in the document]
gurry: Not really a breaking change because this is not a public API

crates/cargo-wdk/src/actions/build/mod.rs
```diff
@@ -57,31 +57,27 @@ pub struct BuildAction<'a> {
...
         fs: &'a Fs,
         metadata: &'a Metadata,
-    ) -> Result<Self> {
-        // TODO: validate and init attrs here
-        let working_dir = fs.canonicalize_path(params.working_dir)?;
-        Ok(Self {
-            working_dir,
+    ) -> Self {
+        // TODO: validate params and return errors.
+        Self {
```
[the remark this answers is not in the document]
gurry: Not a breaking change since it's not a public API and is completely internal to the crate.

crates/cargo-wdk/src/actions/build/build_task.rs
wmmc88: typically, if a constructor is fallible due to **invalid input that the user provided,** then the constructor is conventionally named as `try_new`. malformed input in function arguments to a public function would be considered invalid user input imo.

For private functions, if there is an invariant that is guaranteed to always be true due to private implementation details, then I would just have a normal `assert`. panicking logic is usually fine for internal logic errors or "programmer errors". I would not limit it to only trigger in `dev` profile, because then you'll get weird behavior in `release` builds. 

Give this a read: https://doc.rust-lang.org/book/ch09-03-to-panic-or-not-to-panic.html

crates/cargo-wdk/src/actions/build/error.rs
wmmc88: If it is from paths, then definitely should track a followup to improve this. `std::io` errors from `fs` are very painful to debug without paths

wmmc88: It looks like there is already something providing metadata specifically for IO errors in cargo-wdk: https://github.com/microsoft/windows-drivers-rs/blob/fd8da612a50ad32e68c63702553c46f196977c1c/crates/cargo-wdk/src/providers/mod.rs#L52 

There should be followup to consolidate that with #480 and use it extensively in cargo-wdk. personally, I think the approach in #480 is more generically reusable

crates/cargo-wdk/src/actions/build/build_task.rs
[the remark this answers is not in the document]
wmmc88: > I 2nd this

To clarify, I meant I 2nd questioning why it was moved from result type return to non-result type. @gurry is right that since its a bin-only crate, its not actually a breaking change.

I'll resolve this thread since the Result type is being discussed in https://github.com/microsoft/windows-drivers-rs/pull/462/files/6fd2fc78c1ee5bc36de5cdc16cd7a01ce9df32b0#r2301663721

crates/cargo-wdk/src/actions/build/mod.rs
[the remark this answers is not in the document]
wmmc88: yes. not a breaking change, but if the idea is to validate params in a followup, should we just keep this as result?

crates/cargo-wdk/src/cli.rs
svasista-ms: Changed it. Thank you!

crates/cargo-wdk/src/actions/build/build_task.rs
svasista-ms: Thank you @wmmc88. The task layer operates on the guarantee that the input paths are absolute. I changed it to an `assert!` and added panic tests.

crates/cargo-wdk/src/actions/build/build_task.rs
```diff
@@ -111,3 +105,56 @@ impl<'a> BuildTask<'a> {
...
+
+#[cfg(test)]
+mod tests {
+    use wdk_build::CpuArchitecture;
+
+    use super::*;
+    use crate::actions::{Profile, TargetArch};
+
+    #[test]
+    fn new() {
```
gurry:
```suggestion
    fn new_succeeds_for_valid_args() {
```
A bit more descriptive title

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -493,3 +510,144 @@ impl<'a> PackageTask<'a> {
...
+#[cfg(test)]
+mod tests {
+    use std::path::PathBuf;
+
+    use wdk_build::{CpuArchitecture, KmdfConfig};
+
+    use super::*;
+
+    #[test]
+    fn new() {
```
gurry: These tests should not be hitting the file system. We should fix the `new` method so that it does perform any file operations.

crates/cargo-wdk/src/actions/build/error.rs
gurry: @svasista-ms We should open an issue for a follow up

crates/cargo-wdk/src/actions/build/package_task.rs
svasista-ms: Moved the file system call to `run()` method. Thank you.

crates/cargo-wdk/src/actions/build/build_task.rs
svasista-ms: Thanks for the suggestion. Changed it ✅

crates/cargo-wdk/src/actions/build/mod.rs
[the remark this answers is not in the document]
gurry: To someone reading the code now it would look odd to be returning `Result` for no reason so I feel we should not keep it. We could bring it back when we do add the validations.

gurry (dismissed):

crates/wdk-build/src/lib.rs
svasista-ms: @wmmc88, on closer inspection, we noticed that these functions are not directly used by `cargo-wdk`. When `cargo-wdk` was merged, [detect_wdk_build_number](https://github.com/microsoft/windows-drivers-rs/commit/5b9aa0a14e385daf5217c0f5232bcd4d3f4780a6#diff-43524f733cccc1a175c3bf83309866acb53b8cadb1628478ca1c87d31dda2b26R1350) function was added in `wdk-build/lib.rs` to avoid using fns in `utils.rs`. I removed these re-exports in the recent commit and I'm monitoring the pipelines for any failures related to this change (local test was green).

svasista-ms: There is one limitation in `wdk-build` that is currently preventing `cargo-wdk` from using it. IIUC, `Config` struct can be created only within a build script (due to [detect_cpu_architecture_in_build_script](https://github.com/microsoft/windows-drivers-rs/blob/bd60f3c40bff01c07066aa8f1f4a4ef884427e8e/crates/wdk-build/src/lib.rs#L283)). If we could provide a way to instantiate the same `Config` that was created during the build, `cargo-wdk` can use it for any downstream tasks. Until then can we continue to use `detect_wdk_build_number`?

crates/cargo-wdk/src/actions/build/build_task.rs
```diff
@@ -30,47 +29,42 @@ pub struct BuildTask<'a> {
...
-            Ok(path) => path,
-            Err(StripExtendedPathPrefixError::NoExtendedPathPrefix) => manifest_path,
-            Err(StripExtendedPathPrefixError::EmptyPath) => {
-                return Err(BuildTaskError::EmptyManifestPath);
-            }
-        };
-        Ok(Self {
+    ) -> Self {
+        assert!(
+            working_dir.is_absolute(),
```
krishnakumar4a4, line 56: We already have error propagation set up from `build_task` to `build_action` to `cli` to `main` module with ability to print nested errors.

This particular line is a fallible call and we should have simply leveraged the existing error propagation by introducing an error type for this case instead of panicking deep inside.

krishnakumar4a4: We have had a discussion internally and decided we will keep the assert statements.

crates/cargo-wdk/src/actions/build/build_task.rs
```diff
@@ -30,47 +29,42 @@ pub struct BuildTask<'a> {
...
+            "Working directory path must be absolute. Input path: {}",
+            working_dir.display()
+        );
```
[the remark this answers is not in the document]
gurry: No. The assert is fine.

crates/cargo-wdk/src/actions/build/build_task.rs
```diff
@@ -111,3 +105,56 @@ impl<'a> BuildTask<'a> {
...
+            Some(&profile),
+            target_arch,
+            verbosity_level,
+            &command_exec,
+        );
+
+        assert_eq!(build_task.package_name, package_name);
+        assert_eq!(build_task.profile, Some(&profile));
+        assert_eq!(build_task.target_arch, target_arch);
+        assert_eq!(build_task.manifest_path, working_dir.join("Cargo.toml"));
```
gurry, line 137: Verify `verbosity_level` and `command_exec` as well

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -73,24 +73,41 @@ pub struct PackageTask<'a> {
...
+        assert!(
+            params.working_dir.is_absolute(),
+            "Working directory path must be absolute. Input path: {}",
+            params.working_dir.display()
+        );
+        assert!(
+            params.target_dir.is_absolute(),
+            "Target directory path must be absolute. Input path: {}",
+            params.target_dir.display()
+        );
```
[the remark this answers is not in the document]
svasista-ms: Retaining `assert!` in the `new` function as it is in a private module not used outside of `BuildAction`. Panic here (early) is better than failing in the run method

crates/cargo-wdk/src/actions/build/error.rs
svasista-ms: @wmmc88, added a specific error type for `NotAbsolute` that captures the path along with the `IOError` for now. We discussed the changes in #480 and decided to revisit this and use it in `cargo-wdk` once it is merged.

crates/cargo-wdk/src/actions/build/build_task.rs
svasista-ms: Added assert for `command_exec`.
The current version of `clap_verbosity_level`(3.0.2) does not impl `PartialEq`/`Eq`. However, the latest version (3.0.4) has derived these impls. Added a TODO to update it once the workspace dependency is updated.

crates/cargo-wdk/src/actions/build/error.rs
wmmc88: Fine with that. Just create a followup issue to track considering migrating to fs errors with extra metadata

crates/wdk-build/src/lib.rs
```diff
@@ -26,19 +26,15 @@ pub mod cargo_make;
 pub mod metadata;
 
 mod utils;
-pub use utils::{
```
wmmc88: you can modify this description of this pr to close #479 now

gurry: Done

crates/cargo-wdk/src/actions/build/error.rs
gurry: Issue created: https://github.com/microsoft/windows-drivers-rs/issues/508

PR #476, opened 2025-08-20
Improve lint for mockall usage in cargo-wdk

krishnakumar4a4:
This pull request refactors how the `mockall` crate and related attributes are used in several provider modules within the `crates/cargo-wdk` package. The main goal is to limit the inclusion of mocking and dead code allowances to test configurations, improving code clarity and reducing unnecessary warnings in production builds. The most important changes are grouped below:

**Refactoring of Mocking and Dead Code Allowances:**

* Replaced unconditional uses of `#[automock]` with `#[cfg_attr(test, mockall::automock)]` on the `CommandExec`, `Fs`, `Metadata`, and `WdkBuild` structs, so mocking is only enabled during tests. [[1]](diffhunk://#diff-26f72608332fde1c56f8052d22469ed8e7dbf58e7162ef900e8b65814d45452dL30-R33) [[2]](diffhunk://#diff-b73e1d4215214e77606bf414d1ec9e02ef027f71ac26c375ee7168d7022f4c03L8-R28) [[3]](diffhunk://#diff-c4798e678306e6e4faa4c2c590518de8c72989c06c1243709bc5b1d2b67935e2L8-R22) [[4]](diffhunk://#diff-88e58150b1aec402bbbd73b9ffb1d2d195f17c72e070c724e0f4068cee1cccf6L8-R21)
* Added `#[cfg_attr(test, allow(dead_code, reason = "..."))]` to the same structs to suppress dead code warnings only in test configurations, clarifying that these implementations are intentionally unused outside of tests. [[1]](diffhunk://#diff-26f72608332fde1c56f8052d22469ed8e7dbf58e7162ef900e8b65814d45452dL30-R33) [[2]](diffhunk://#diff-b73e1d4215214e77606bf414d1ec9e02ef027f71ac26c375ee7168d7022f4c03L8-R28) [[3]](diffhunk://#diff-c4798e678306e6e4faa4c2c590518de8c72989c06c1243709bc5b1d2b67935e2L8-R22) [[4]](diffhunk://#diff-88e58150b1aec402bbbd73b9ffb1d2d195f17c72e070c724e0f4068cee1cccf6L8-R21)
* Removed unnecessary `#![allow(dead_code)]` attributes from the top of provider modules, as dead code is now only allowed in test builds. [[1]](diffhunk://#diff-26f72608332fde1c56f8052d22469ed8e7dbf58e7162ef900e8b65814d45452dL10-L12) [[2]](diffhunk://#diff-b73e1d4215214e77606bf414d1ec9e02ef027f71ac26c375ee7168d7022f4c03L8-R28) [[3]](diffhunk://#diff-c4798e678306e6e4faa4c2c590518de8c72989c06c1243709bc5b1d2b67935e2L8-R22) [[4]](diffhunk://#diff-88e58150b1aec402bbbd73b9ffb1d2d195f17c72e070c724e0f4068cee1cccf6L8-R21)
* Removed unused imports of `mockall::automock` from provider modules, since the macro is now conditionally applied. [[1]](diffhunk://#diff-26f72608332fde1c56f8052d22469ed8e7dbf58e7162ef900e8b65814d45452dL21) [[2]](diffhunk://#diff-b73e1d4215214e77606bf414d1ec9e02ef027f71ac26c375ee7168d7022f4c03L8-R28) [[3]](diffhunk://#diff-c4798e678306e6e4faa4c2c590518de8c72989c06c1243709bc5b1d2b67935e2L8-R22) [[4]](diffhunk://#diff-88e58150b1aec402bbbd73b9ffb1d2d195f17c72e070c724e0f4068cee1cccf6L8-R21)

**Cargo.toml Dependency Adjustments:**

* Moved the `mockall` dependency from `[dependencies]` to `[dev-dependencies]` in `crates/cargo-wdk/Cargo.toml`, ensuring it is only included for testing. [[1]](diffhunk://#diff-a6ed6baf7034b515714a7665cfb4ad2c0f373b95ca772b907d164b3bce4e8dbaL21) [[2]](diffhunk://#diff-a6ed6baf7034b515714a7665cfb4ad2c0f373b95ca772b907d164b3bce4e8dbaR31)

### DIFF
--- a/crates/cargo-wdk/Cargo.toml
+++ b/crates/cargo-wdk/Cargo.toml
@@ -18,7 +18,6 @@ clap = { workspace = true, features = ["derive"] }
 clap-verbosity-flag.workspace = true
 fs4.workspace = true
 include_dir.workspace = true
-mockall.workspace = true
 mockall_double.workspace = true
 serde_json.workspace = true
 thiserror.workspace = true
@@ -29,6 +28,7 @@ wdk-build.workspace = true
 [dev-dependencies]
 assert_cmd.workspace = true
 assert_fs.workspace = true
+mockall.workspace = true
 predicates.workspace = true
 sha2.workspace = true
 
--- a/crates/cargo-wdk/src/providers/exec.rs
+++ b/crates/cargo-wdk/src/providers/exec.rs
@@ -7,9 +7,6 @@
 
 // Suppression added for mockall as it generates mocks with env_vars: &Option
 #![allow(clippy::ref_option_ref)]
-// Warns the run method is not used, however it is used.
-// The intellisense confusion seems to come from automock
-#![allow(dead_code)]
 #![allow(clippy::unused_self)]
 
 use std::{
@@ -18,7 +15,6 @@ use std::{
 };
 
 use anyhow::Result;
-use mockall::automock;
 use tracing::debug;
 
 use super::error::CommandError;
@@ -27,7 +23,7 @@ use super::error::CommandError;
 #[derive(Debug, Default)]
 pub struct CommandExec {}
 
-#[automock]
+#[cfg_attr(test, mockall::automock)]
 impl CommandExec {
     pub fn run<'a>(
         &self,
--- a/crates/cargo-wdk/src/providers/exec.rs
+++ b/crates/cargo-wdk/src/providers/exec.rs
@@ -24,6 +24,13 @@ use super::error::CommandError;
 pub struct CommandExec {}
 
 #[cfg_attr(test, mockall::automock)]
+#[cfg_attr(
+    test,
+    allow(
+        dead_code,
+        reason = "This implementation is mocked in test configuration."
+    )
+)]
 impl CommandExec {
     pub fn run<'a>(
         &self,
--- a/crates/cargo-wdk/src/providers/fs.rs
+++ b/crates/cargo-wdk/src/providers/fs.rs
@@ -5,18 +5,12 @@
 //! operations such as reading, writing, copying, and checking file existence.
 //! It also integrates with `mockall` to enable mocking for unit tests.
 
-// Warns the methods are not used, however they are used.
-// The intellisense confusion seems to come from automock
-#![allow(dead_code)]
 #![allow(clippy::unused_self)]
-
 use std::{
     fs::{copy, create_dir, read_dir, rename, DirEntry, File, FileType, OpenOptions},
     io::{Read, Write},
     path::{Path, PathBuf},
 };
-
-use mockall::automock;
 
 use super::error::FileError;
 
@@ -24,7 +18,7 @@ use super::error::FileError;
 #[derive(Default)]
 pub struct Fs {}
 
-#[automock]
+#[cfg_attr(test, mockall::automock)]
 impl Fs {
     pub fn canonicalize_path(&self, path: &Path) -> Result<PathBuf, FileError> {
         path.canonicalize()
--- a/crates/cargo-wdk/src/providers/fs.rs
+++ b/crates/cargo-wdk/src/providers/fs.rs
@@ -19,6 +19,13 @@ use super::error::FileError;
 pub struct Fs {}
 
 #[cfg_attr(test, mockall::automock)]
+#[cfg_attr(
+    test,
+    allow(
+        dead_code,
+        reason = "This implementation is mocked in test configuration."
+    )
+)]
 impl Fs {
     pub fn canonicalize_path(&self, path: &Path) -> Result<PathBuf, FileError> {
         path.canonicalize()
--- a/crates/cargo-wdk/src/providers/metadata.rs
+++ b/crates/cargo-wdk/src/providers/metadata.rs
@@ -5,19 +5,14 @@
 //! `mockall` crate to enable mocking of its methods, facilitating easier unit
 //! testing.
 
-// Warns the get_cargo_metadata_at_path method is not used, however it is used.
-// The intellisense confusion seems to come from automock
-#![allow(dead_code)]
 #![allow(clippy::unused_self)]
 
 use std::path::Path;
 
-use mockall::automock;
-
 #[derive(Default)]
 pub struct Metadata {}
 
-#[automock]
+#[cfg_attr(test, mockall::automock)]
 impl Metadata {
     /// Get the Cargo metadata at a given path.
     ///
--- a/crates/cargo-wdk/src/providers/metadata.rs
+++ b/crates/cargo-wdk/src/providers/metadata.rs
@@ -13,6 +13,13 @@ use std::path::Path;
 pub struct Metadata {}
 
 #[cfg_attr(test, mockall::automock)]
+#[cfg_attr(
+    test,
+    allow(
+        dead_code,
+        reason = "This implementation is mocked in test configuration."
+    )
+)]
 impl Metadata {
     /// Get the Cargo metadata at a given path.
     ///
--- a/crates/cargo-wdk/src/providers/wdk_build.rs
+++ b/crates/cargo-wdk/src/providers/wdk_build.rs
@@ -12,6 +12,13 @@
 pub struct WdkBuild {}
 
 #[cfg_attr(test, mockall::automock)]
+#[cfg_attr(
+    test,
+    allow(
+        dead_code,
+        reason = "This implementation is mocked in test configuration."
+    )
+)]
 impl WdkBuild {
     pub fn detect_wdk_build_number(&self) -> Result<u32, wdk_build::ConfigError> {
         wdk_build::detect_wdk_build_number()
--- a/crates/cargo-wdk/src/providers/wdk_build.rs
+++ b/crates/cargo-wdk/src/providers/wdk_build.rs
@@ -5,17 +5,13 @@
 //! It leverages the `mockall` crate to enable mocking of the `WdkBuild` struct
 //! for improved testability in unit tests.
 
-// Warns the detect_wdk_build_number method is not used, however it is used.
-// The intellisense confusion seems to come from automock
-#![allow(dead_code)]
 #![allow(clippy::unused_self)]
-use mockall::automock;
 
 /// Provides limited access to wdk-build crate methods
 #[derive(Default)]
 pub struct WdkBuild {}
 
-#[automock]
+#[cfg_attr(test, mockall::automock)]
 impl WdkBuild {
     pub fn detect_wdk_build_number(&self) -> Result<u32, wdk_build::ConfigError> {
         wdk_build::detect_wdk_build_number()


crates/cargo-wdk/src/providers/exec.rs
```diff
@@ -27,7 +23,14 @@ use super::error::CommandError;
 #[derive(Debug, Default)]
 pub struct CommandExec {}
 
-#[automock]
+#[cfg_attr(test, mockall::automock)]
+#[cfg_attr(
+    test,
+    allow(
+        dead_code,
+        reason = "This implementation is mocked in test configuration."
```
gurry: The reason is not clear. It does not quite explain _why_ we are allowing dead code.

krishnakumar4a4: Okay. How about this?

"The implementation is mocked, so the actual implementation will appear as dead code in test configuration"

gurry: That's better. Slight improvement: "Tests use mocked impl so this impl becomes dead code in test configuration"

crates/cargo-wdk/Cargo.toml
```diff
@@ -18,7 +18,6 @@ clap = { workspace = true, features = ["derive"] }
 clap-verbosity-flag.workspace = true
 fs4.workspace = true
 include_dir.workspace = true
-mockall.workspace = true
 mockall_double.workspace = true
```
gurry: Should this also be moved to dev deps?

crates/cargo-wdk/src/providers/exec.rs
krishnakumar4a4: I updated the reason to

`Tests use mocked implementation, so this implementation becomes dead code in test configuration.`

gurry: "impl" is more precise than "implementation" in the context of Rust and shorter, But it's not a big deal.

crates/cargo-wdk/Cargo.toml
krishnakumar4a4: Actually, it looks like possible. I have made changes and pushed. Will wait for pipeline to double confirm this.

gurry (approved):

wmmc88:
@krishnakumar4a4 please title the pr with conventional commit convention

crates/cargo-wdk/src/providers/exec.rs
```diff
@@ -27,8 +18,24 @@ use super::error::CommandError;
...
+    test,
+    allow(
+        dead_code,
+        reason = "Tests use mocked implementation, so this implementation becomes dead code in \
+                  test configuration."
+    )
+)]
+#[allow(
+    clippy::unused_self,
+    reason = "Mocking associated functions complicates testing; using instance methods instead"
```
wmmc88: I don't understand the justification here. could you elaborate?

crates/cargo-wdk/src/providers/exec.rs
```diff
@@ -27,8 +18,24 @@ use super::error::CommandError;
...
+)]
 impl CommandExec {
+    #[allow(
```
wmmc88: this should only be allow during `cfg(test)`, right?

crates/cargo-wdk/src/providers/metadata.rs
```diff
@@ -5,19 +5,24 @@
...
+#[cfg_attr(test, mockall::automock)]
+#[cfg_attr(
+    test,
+    allow(
+        dead_code,
+        reason = "Tests use mocked implementation, so this implementation becomes dead code in \
+                  test configuration."
+    )
+)]
+#[allow(
```
wmmc88, line 22: ditto

crates/cargo-wdk/src/providers/fs.rs
```diff
@@ -5,26 +5,31 @@
...
+#[cfg_attr(test, mockall::automock)]
+#[cfg_attr(
+    test,
+    allow(
+        dead_code,
+        reason = "Tests use mocked implementation, so this implementation becomes dead code in \
+                  test configuration."
+    )
+)]
+#[allow(
```
wmmc88, line 29: cfg(test)?

crates/cargo-wdk/src/providers/exec.rs
krishnakumar4a4: Good catch, this one can be `cfg(test, allow(...))`. I will make the change.

crates/cargo-wdk/src/providers/exec.rs
krishnakumar4a4: This is one of the clippy lint issues I wanted to get your thoughts on, since having &self just for testing doesn’t feel quite right to me either. The only reason we originally did it was because, when we were writing providers, we thought we’d need to keep some state at the struct level—but it turned out we didn’t.

So, going back to why I made this choice, when I looked at this page in the past
https://docs.rs/mockall/latest/mockall/#static-methods

The first paragraph gave me the impression that "mocks generated for static or associated functions on a struct would be global, and their expectations would be global too, and synchronization is needed" which worried me (plus the possible need to keep state). That’s why I went with instance methods.

But after playing around with Copilot, it seems easy to make this work. If you create a separate context for each test, you can set expectations differently in each test’s scope. I still need to code it out to confirm, though.

@wmmc88 Do you also see it the same way?

crates/cargo-wdk/src/providers/metadata.rs
krishnakumar4a4: If you are referring to the use of `unused_self` here, this lint suppression is also needed in non-test configuration as its not being used at all. 

This lint suppression could be removed if we change the instance methods to associated functions as I mentioned here https://github.com/microsoft/windows-drivers-rs/pull/476#discussion_r2307114227

crates/cargo-wdk/src/providers/fs.rs
krishnakumar4a4: Same reason as this one - https://github.com/microsoft/windows-drivers-rs/pull/476#discussion_r2307121694

crates/cargo-wdk/src/providers/exec.rs
krishnakumar4a4: I've tested this locally and noticed that the expectations for the associated functions are set at the global process level. I initially thought the context would help restrict the expectation to the relevant unit test, but it seems the context only serves to clean up expectations when it's dropped. 
Synchronizing unit tests was one reason I hesitated to use associated functions, and the context doesn't address this issue. So, we're back to needing this lint, and I believe the `reason` provided explains why it's necessary here.

wmmc88: I don't think we should introduce useless self args in the api, just so that the mocked test is a little easier. The linked docs show an easy way to get ctx on statics to work in tests: https://github.com/asomers/mockall/blob/master/mockall/examples/synchronization.rs

also the reason itself in the lint is unclear. I should be able to understand exactly why a lint is ignored based on the comment. 
>     reason = "Mocking associated functions complicates testing; using instance methods instead"

How and why does it complicate testing? instance methods used for what? None of these details were clear

krishnakumar4a4: Let me quickly sum up my points and the options I see. First off, the lint we're seeing isn't new from this PR—it's just been moved from the file level to where it's actually needed, and I've tried to add a reason for it in this PR.

This link https://github.com/asomers/mockall/blob/master/mockall/examples/synchronization.rs shows that we need a lock to sync tests when using associated functions. Otherwise, if the unit tests run in parallel, the mock expectations can override each other since any expectations set on the context are global and not isolated per test. I have checked this locally and I can confirm the expectations are being overridden.

So, here are the two options I see:

1. Switch from instance methods to associated functions, which would mean serializing all unit tests in the actions module because every unit test will touch the provider methods one way or the other. We could use a Mutex lock for that, as shown in this link https://github.com/asomers/mockall/blob/master/mockall/examples/synchronization.rs
2. Stick with the instance methods and the lint as it is (which is the current state), and focus on improving the reason for the lint. The reason could be look like this 
>
```
reason = "The methods within the struct do not need to take self and can be defined as associated functions. However, implementing them as instance methods allows for per-unit test mocking and avoid synchronizing the unit tests"
```

@wmmc88 Which of these two options do you prefer? Both have their ups and downs.

Option 1 pros: One less lint suppression per provider module in the dev code.
Cons: It makes things slightly more complicated (extra context use) and ends up serializing the unit tests. Also, would serialization be a potential problem for cargo nexttest?

Option 2 cons: We'll need one additional awkward looking lint suppression per provider module.
Pros: Keeps the already verbose unit tests from getting more complex or serialized.

gurry: Given a choice I would prefer we didn't have to use `&self` just to make the mock framework happy. But if that means our unit tests run slower, then I'd say it's fine. Since this is an internal API and not a public one, we have some leeway here.

wmmc88: IMO go with #1. I do not think we should be adding lint exceptions for unit tests.

It won't interfere with nextest. The way nextest works is it runs each test in a separate process. If users find unit tests are a bottle neck (which imo rn they aren't. They're fairly quick especially compared to the integration tests), they should be able to just run under nextest for a speedup

krishnakumar4a4: Sure, that makes sense. 

Unit tests are located in multiple modules under `src`, so a module-level mutex wouldn't work for serializing unit tests. Also, since nextest may run tests in multiple processes, using a file lock instead of a mutex seems like a better option for now.

The `crates/cargo-wdk/src` folder appears to be a suitable directory for lock. Other options like `file!` or `CARGO_MANIFEST_PATH` don't seem as useful. I would avoid relying on `CARGO_MANIFEST_PATH` solely for unit tests, since there are integration tests as well.

wmmc88: > Unit tests are located in multiple modules under src, so a module-level mutex wouldn't work for serializing unit tests. Also, since nextest may run tests in multiple processes, using a file lock instead of a mutex seems like a better option for now.

the context is global to the process (since nextest is test-per-process), so synchronization doesn't matter for nextest. in fact, flocking ends up introducing unnecessary locking for nextest.

Could you just use a Mutex in `crate::tests` thats shared among unit tests?

crates/cargo-wdk/src/providers/exec.rs
```diff
@@ -4,33 +4,38 @@
...
+        reason = "Tests use mocked implementation, so this implementation becomes dead code in \
+                  test configuration."
+    )
+)]
 impl CommandExec {
+    #[cfg_attr(
+        test,
+        allow(
+            clippy::ref_option_ref,
+            reason = "Mock adds an extra reference to Option<HashMap<&str, &str>>."
```
gurry, line 36: The "reason" string mentions the type `Option<HashMap<&str, &str>>`, but the `env_vars` parameter's type is `Option<&HashMap<&str, &str>>`. Is there a mistake here?

gurry (commented):
Please merge/rebase with main and resolve conflicts

krishnakumar4a4:
Update: I have paused this work for now, as there are ongoing PRs that conflict with it and require rework. I will get back to this later.

PR #430, opened 2025-08-01
chore!: bump to Rust 2024 Edition

leon-xd:
**Breaking change**:
- MSRV bumped: 1.84 => 1.85

**Other changes**
- `unsafe` wrapper around `env::set_var`, `env::remove_var` functions, and `no_mangle` and `export_name` attributes
- Ran `clippy` on wdk crate -- this was not done in the past PRs as no automatic uncommenting of WDK configuration happens in pipeline
- general `fmt` & `clippy` fixes 
- added `cargo-sort` installation to `CONTRIBUTING.md`

PR #520, opened 2025-09-29
feat: support stampinf version override

krishnakumar4a4:
This pull request introduces support for overriding the `DriverVer` version in the `stampinf` tool via an environment variable when specific unstable build flags are enabled. It also adds a parameterized test to ensure this behavior works as intended, and updates configuration files to support and lint for the new conditional compilation flags.

**Feature: Conditional DriverVer override via environment variable**
- Updated `PackageTask::run_stampinf` in `package_task.rs` to allow the `STAMPINF_VERSION` environment variable to override the auto-generated `DriverVer` when both `wdk_build_unstable` and `allow_stampinf_version_env_override` cfg flags are enabled; otherwise, auto-generation is always used.

**Testing: Parameterized test for new behavior**
- Added a parameterized test `run_stampinf_parameterized_env_overrides` in `package_task.rs` to verify correct handling of the `STAMPINF_VERSION` environment variable under different cfg flag and environment conditions.

**Configuration: Support and linting for new cfg flag**
- Added `allow_stampinf_version_env_override` as a commented-out unstable cfg option in `.cargo/config.toml` for discoverability.
- Updated `Cargo.toml` to warn on unexpected cfgs and explicitly allow the new `allow_stampinf_version_env_override` flag, improving build hygiene and discoverability.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -9,4 +9,5 @@ rustflags = [
 
   # Unstable cfg options:
   # "--cfg", "skip_umdf_static_crt_check",
+  # "--cfg", "allow_stampinf_version_env_override"
 ]
--- a/crates/cargo-wdk/src/actions/build/package_task.rs
+++ b/crates/cargo-wdk/src/actions/build/package_task.rs
@@ -315,11 +315,23 @@ impl<'a> PackageTask<'a> {
             &cat_file_path,
         ];
 
-        if let Ok(version) = std::env::var("STAMPINF_VERSION") {
-            // If STAMPINF_VERSION is set we rely on stampinf reading it; do NOT add -v.
-            // Otherwise we pass -v * (see else) to let stampinf auto-generate the version.
-            info!("Using STAMPINF_VERSION env var to set DriverVer: {version}");
-        } else {
+        // DriverVer handling:
+        // 1. When BOTH cfg flags (wdk_build_unstable + allow_stampinf_version_env_override) are
+        //    enabled, allow an external override via STAMPINF_VERSION env var. If the env var is
+        //    absent we fall back to auto-generation (-v *).
+        // 2. Otherwise (stable / default builds) always request auto-generation (-v *).
+        #[cfg(all(wdk_build_unstable, allow_stampinf_version_env_override))]
+        {
+            if let Ok(version) = std::env::var("STAMPINF_VERSION") {
+                // Rely on stampinf reading STAMPINF_VERSION; do NOT add -v.
+                info!("Using STAMPINF_VERSION env var to set DriverVer: {version}");
+            } else {
+                args.push("-v");
+                args.push("*");
+            }
+        }
+        #[cfg(not(all(wdk_build_unstable, allow_stampinf_version_env_override)))]
+        {
             args.push("-v");
             args.push("*");
         }
--- a/crates/cargo-wdk/src/actions/build/package_task.rs
+++ b/crates/cargo-wdk/src/actions/build/package_task.rs
@@ -313,9 +313,17 @@ impl<'a> PackageTask<'a> {
             &arch,
             "-c",
             &cat_file_path,
-            "-v",
-            "*",
         ];
+
+        if let Ok(version) = std::env::var("STAMPINF_VERSION") {
+            // If STAMPINF_VERSION is set we rely on stampinf reading it; do NOT add -v.
+            // Otherwise we pass -v * (see else) to let stampinf auto-generate the version.
+            info!("Using STAMPINF_VERSION env var to set DriverVer: {version}");
+        } else {
+            args.push("-v");
+            args.push("*");
+        }
+
         if !wdf_version_flags.is_empty() {
             args.append(&mut wdf_version_flags.iter().map(String::as_str).collect());
         }
--- a/crates/cargo-wdk/src/actions/build/package_task.rs
+++ b/crates/cargo-wdk/src/actions/build/package_task.rs
@@ -651,4 +651,77 @@ mod tests {
 
         PackageTask::new(package_task_params, &wdk_build, &command_exec, &fs);
     }
+
+    // Parameterized test for run_stampinf covering env present/absent.
+    // Compile-time cfgs determine expectation via cfg! so a single test body suffices.
+    #[test]
+    fn run_stampinf_parameterized_env_overrides() {
+        use std::process::{ExitStatus, Output};
+
+        let scenarios = [("env_set", Some("1.2.3.4")), ("env_unset", None)];
+
+        let both_cfgs = cfg!(all(wdk_build_unstable, allow_stampinf_version_env_override));
+
+        for (name, env_val) in scenarios {
+            if let Some(v) = env_val {
+                std::env::set_var("STAMPINF_VERSION", v);
+            } else {
+                std::env::remove_var("STAMPINF_VERSION");
+            }
+
+            let package_name = "param_pkg"; // reuse
+            let working_dir = PathBuf::from("C:/abs/wd");
+            let target_dir = PathBuf::from("C:/abs/target");
+            let arch = CpuArchitecture::Amd64;
+
+            let params = PackageTaskParams {
+                package_name,
+                working_dir: &working_dir,
+                target_dir: &target_dir,
+                target_arch: &arch,
+                driver_model: DriverConfig::Kmdf(KmdfConfig::default()),
+                sample_class: false,
+                verify_signature: false,
+            };
+
+            let wdk_build = WdkBuild::default();
+            let fs = Fs::default();
+            let mut command_exec = CommandExec::default();
+
+            let expect_skip_v = env_val.is_some() && both_cfgs; // skip -v only in override path
+
+            command_exec
+                .expect_run()
+                .withf(move |cmd: &str, args: &[&str], _, _| {
+                    if cmd != "stampinf" {
+                        return false;
+                    }
+                    let has_v = args.contains(&"-v");
+                    if expect_skip_v {
+                        !has_v
+                    } else {
+                        args.windows(2).any(|w| w == ["-v", "*"])
+                    }
+                })
+                .once()
+                .return_once(|_, _, _, _| {
+                    Ok(Output {
+                        status: ExitStatus::default(),
+                        stdout: vec![],
+                        stderr: vec![],
+                    })
+                });
+
+            let task = PackageTask::new(params, &wdk_build, &command_exec, &fs);
+            let result = task.run_stampinf();
+            assert!(
+                result.is_ok(),
+                "scenario {name} failed (cfgs_override_enabled={both_cfgs}, env_set={:?})",
+                env_val
+            );
+        }
+
+        // Cleanup
+        std::env::remove_var("STAMPINF_VERSION");
+    }
 }
--- a/crates/cargo-wdk/Cargo.toml
+++ b/crates/cargo-wdk/Cargo.toml
@@ -36,6 +36,10 @@ sha2.workspace = true
 missing_docs = "warn"
 unsafe_op_in_unsafe_fn = "forbid"
 
+[lints.rust.unexpected_cfgs]
+level = "warn"
+check-cfg = ["cfg(wdk_build_unstable)", "cfg(allow_stampinf_version_env_override)"]
+
 [lints.clippy]
 # Lint Groups
 all = "deny"


crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -313,9 +313,29 @@ impl<'a> PackageTask<'a> {
...
+        // 1. When BOTH cfg flags (wdk_build_unstable + allow_stampinf_version_env_override) are
+        //    enabled, allow an external override via STAMPINF_VERSION env var. If the env var is
+        //    absent we fall back to auto-generation (-v *).
+        // 2. Otherwise (stable / default builds) always request auto-generation (-v *).
+        #[cfg(all(wdk_build_unstable, allow_stampinf_version_env_override))]
+        {
+            if let Ok(version) = std::env::var("STAMPINF_VERSION") {
+                // Rely on stampinf reading STAMPINF_VERSION; do NOT add -v.
+                info!("Using STAMPINF_VERSION env var to set DriverVer: {version}");
+            } else {
```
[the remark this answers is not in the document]
krishnakumar4a4: Updated the comment.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -313,9 +313,28 @@ impl<'a> PackageTask<'a> {
...
+        // 2. Otherwise (stable / default builds) always request auto-generation (-v *).
+        cfg_if::cfg_if! {
+            if #[cfg(allow_stampinf_version_env_override)]
+            {
+                if let Ok(version) = std::env::var("STAMPINF_VERSION") && !version.trim().is_empty() {
+                    // When STAMPINF_VERSION is set we intentionally omit -v so stampinf reads it
+                    // and populates DriverVer.
+                    info!("Using STAMPINF_VERSION env var to set DriverVer: {version}");
+                } else {
+                    args.extend(["-v", "*"]);
```
wmmc88: What happens if you set `-v *` but also specify a `STAMPINF_VERSION` env var when **directly executing stampinf.exe**?

.cargo/config.toml
```diff
@@ -9,4 +9,5 @@ rustflags = [
 
   # Unstable cfg options:
   # "--cfg", "skip_umdf_static_crt_check",
+  # "--cfg", "allow_stampinf_version_env_override"
```
wmmc88: Since we're using the official env var of the tool here, i think its fine to not gate this behind unstable. Contingent on treating all the tool env vars in the same way as we treated `STAMPINF_VERSION` in this pr, i think removing the unstable cfg is fine. What are others' thoughts here @microsoft/windows-drivers-rs ?

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -313,9 +313,28 @@ impl<'a> PackageTask<'a> {
...
+
+        // DriverVer handling:
+        // 1. When allow_stampinf_version_env_override cfg is enabled, allow an external
+        //    override via STAMPINF_VERSION env var. If the env var is absent we fall
+        //    back to auto-generation (-v *).
+        // 2. Otherwise (stable / default builds) always request auto-generation (-v *).
+        cfg_if::cfg_if! {
+            if #[cfg(allow_stampinf_version_env_override)]
+            {
+                if let Ok(version) = std::env::var("STAMPINF_VERSION") && !version.trim().is_empty() {
```
wmmc88: imo pull out all env vars to consts (ex. so they can also be used elsewhere like in the log statement)

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -631,4 +650,77 @@ mod tests {
...
+        // verify both with and without the env var set scenarios
+        let scenarios = [("env_set", Some("1.2.3.4")), ("env_unset", None)];
+
+        // Compile-time evaluation: cfg!(...) => true only if
+        // `allow_stampinf_version_env_override` was enabled during build
+        let is_cfg_set = cfg!(allow_stampinf_version_env_override);
+
+        for (name, env_val) in scenarios {
+            if let Some(v) = env_val {
+                std::env::set_var("STAMPINF_VERSION", v);
```
wmmc88: may race with changes in #430 depending on who gets their changes in first. fyi @leon-xd

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -631,4 +650,77 @@ mod tests {
...
+            } else {
+                std::env::remove_var("STAMPINF_VERSION");
```
wmmc88: manipulating env variables in tests requires locking to prevent flaky tests

.cargo/config.toml
gurry: I agree. We were discussing the same thing yesterday. cfgs are a bit painful to set during cargo install as well. So just having the env var is enough.

Yes, we could generalize this scheme to other tools as well, but that overall UX design is yet to be done. Let's see how it plays out.

crates/cargo-wdk/src/actions/build/package_task.rs
krishnakumar4a4: If the environment variable takes precedence, it actually simplifies the implementation here, if that's what you are thinking too. However, the documentation states that the environment variable will be ignored if the -v flag is present. I will confirm this locally once.

crates/cargo-wdk/src/actions/build/package_task.rs
wmmc88: https://github.com/microsoft/windows-drivers-rs/blob/63f8dd8a53df0fc41080f62ec2b1d040d6d28ba7/crates/wdk-build/src/lib.rs#L1729 for an example

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -631,4 +653,76 @@ mod tests {
...
+                .withf(move |cmd: &str, args: &[&str], _, _| {
+                    if cmd != "stampinf" {
+                        return false;
+                    }
+                    let has_v = args.contains(&"-v");
+                    if expect_skip_v {
+                        !has_v
+                    } else {
+                        args.windows(2).any(|w| w == ["-v", "*"])
+                    }
```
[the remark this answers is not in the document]
krishnakumar4a4: No, the current one is better.

crates/cargo-wdk/src/actions/build/package_task.rs
krishnakumar4a4: I verified this, `-v *` takes the precedence in presence of both.

.cargo/config.toml
krishnakumar4a4: Removed the cfg.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -313,9 +314,19 @@ impl<'a> PackageTask<'a> {
...
+        match std::env::var(STAMPINF_VERSION_ENV_VAR) {
+            Ok(version) if !version.trim().is_empty() => {
+                // When STAMPINF_VERSION is set we intentionally omit -v so stampinf reads it
+                // and populates DriverVer.
+                info!("Using {STAMPINF_VERSION_ENV_VAR} env var to set DriverVer: {version}");
+            }
+            _ => {
+                args.extend(["-v", "*"]);
+            }
+        }
```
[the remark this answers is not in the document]
krishnakumar4a4: the cfg has been removed.

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -54,39 +57,49 @@ fn kmdf_driver_builds_successfully() {
...
 
 #[test]
 fn wdm_driver_builds_successfully() {
-    with_file_lock(|| build_driver_project("wdm"));
+    with_file_lock::<&str, &str, _>(&[], || clean_and_build_driver_project("wdm", None));
+}
+
+#[test]
+fn wdm_driver_builds_successfully_with_given_version() {
+    with_file_lock::<&str, &str, _>(&[(STAMPINF_VERSION_ENV_VAR, "5.1.0")], || {
```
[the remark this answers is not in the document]
krishnakumar4a4: This test setup is intentional.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -313,9 +314,19 @@ impl<'a> PackageTask<'a> {
...
             &cat_file_path,
-            "-v",
-            "*",
         ];
+
+        match std::env::var(STAMPINF_VERSION_ENV_VAR) {
+            Ok(version) if !version.trim().is_empty() => {
+                // When STAMPINF_VERSION is set we intentionally omit -v so stampinf reads it
+                // and populates DriverVer.
+                info!("Using {STAMPINF_VERSION_ENV_VAR} env var to set DriverVer: {version}");
```
wmmc88: Did you sanity test this behavior on the latest commit while compiling a driver with the env var set?

gurry: This doesn't have to be `info`. Better to move it to `debug`

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -97,7 +110,18 @@ fn build_driver_project(driver_type: &str) {
...
-    verify_driver_package_files(&driver_path, &driver_name, driver_binary_extension);
+    verify_driver_package_files(
+        &driver_path,
+        &driver_name,
+        driver_binary_extension,
+        driver_version,
+    );
+}
+
+fn run_clean_cmd(driver_path: &str) {
```
gurry:
```suggestion
fn run_cargo_clean(driver_path: &str) {
```

This name is clearer

crates/cargo-wdk/src/actions/build/package_task.rs
krishnakumar4a4: Yes, I did it as part of the newly added integration test https://github.com/microsoft/windows-drivers-rs/pull/520/files#diff-7f39e8d21b60832dceaba74ceb186dce58e45d2c660c6ce0674d1be3181a18e9R74

I checked this again by running manually on a couple of driver setups, as shown below

Screenshot 2025-10-07 103130
Screenshot 2025-10-07 105813
Screenshot 2025-10-07 105853

crates/cargo-wdk/tests/build_command_test.rs
krishnakumar4a4: Renamed it.

crates/cargo-wdk/src/actions/build/package_task.rs
krishnakumar4a4: Changed this statement to debug.

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -175,6 +202,23 @@ fn assert_file_hash(path1: &str, path2: &str) {
...
+    let file_content =
+        fs::read(format!("{package_path}/{driver_name}.inf")).expect("Unable to read inf file");
+    let file_content = String::from_utf8_lossy(&file_content);
+    file_content.lines().for_each(|line| {
+        // Example: DriverVer=09/13/2023,1.0.0.0
+        let driver_version_regex = driver_version.unwrap_or(r"\d+\.\d+\.\d+\.\d+");
+        let re = regex::Regex::new(&format!(
+            r"^DriverVer\s+=\s+\d+/\d+/\d+,{driver_version_regex}$"
+        ))
+        .unwrap();
```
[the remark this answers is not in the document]
krishnakumar4a4: This should be fine.

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -175,6 +202,37 @@ fn assert_file_hash(path1: &str, path2: &str) {
...
 }
 
+fn assert_driver_ver(package_path: &str, driver_name: &str, driver_version: Option<&str>) {
+    // Read the INF file as raw bytes and produce a best-effort UTF-8 string.
+    let file_content =
+        fs::read(format!("{package_path}/{driver_name}.inf")).expect("Unable to read inf file");
+    let file_content = if file_content.starts_with(&[0xFF, 0xFE]) {
+        // Handle UTF-16 LE (BOM 0xFF 0xFE).
+        let file_content = file_content
+            .chunks(2)
```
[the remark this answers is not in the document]
krishnakumar4a4: UTF16 encoded files would always contain even number of bytes. So, this is fine.

crates/cargo-wdk/src/actions/build/package_task.rs
krishnakumar4a4: Thanks @wmmc88. I’ve added this function to cargo-wdk and am using it in the current test. I’ve skipped it for the other unit tests in this module, as they aren’t affected by the presence or absence of this environment variable.

While this seemed fine in the current module, I realised that all the [tests in the build module](https://github.com/microsoft/windows-drivers-rs/blob/main/crates/cargo-wdk/src/actions/build/tests.rs) also require locking as each of them uses an expect on the stampinf invocation. It doesn’t fail on the CI pipeline at the moment, but it may fail when the unit tests in the build and package_task modules run in parallel enough times. I was originally planning to implement a form of locking as part of another PR (https://github.com/microsoft/windows-drivers-rs/pull/476#discussion_r2316751570), but it looks like we need to address it now. Do you also see it that way?

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -157,6 +164,18 @@ fn verify_driver_package_files(
...
+        .expect("Unable to read inf file")
+        .lines()
+        .for_each(|line| {
+            // Example: DriverVer=09/13/2023,1.0.0.0
+            let re =
+                regex::Regex::new(r"^DriverVer\s+=\s+\d+/\d+/\d+,\d+\.\d+\.\d+\.\d+$").unwrap();
+            if line.starts_with("DriverVer") {
+                assert!(re.captures(line).is_some());
+            }
+        });
```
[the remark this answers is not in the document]
krishnakumar4a4: This suggestion is obsolete now.

crates/cargo-wdk/src/actions/build/package_task.rs
krishnakumar4a4: I went ahead and modified the tests in the build module to also use the with_env method, so that locking occurs across tests in both the build and package_task modules.

crates/cargo-wdk/src/main.rs
```diff
@@ -36,3 +36,70 @@ fn main() -> Result<()> {
...
+    /// The environment is guaranteed to be not modified during the execution
+    /// of the function, and the environment is reset to its original state
+    /// after execution of the function. No testing asserts should be called in
+    /// the function, since a failing test will poison the mutex, and cause all
+    /// remaining tests to fail.
+    ///
+    /// # Panics
+    ///
+    /// Panics if called with duplicate environment variable keys.
+    pub fn with_env<K, V, F, R>(env_vars_key_value_pairs: &[(K, Option<V>)], f: F) -> R
```
gurry: This function doesn't belog in `main.rs`. Better to put it in `tests`.

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -175,6 +202,37 @@ fn assert_file_hash(path1: &str, path2: &str) {
...
+        let file_content = file_content
+            .chunks(2)
+            .map(|pair| u16::from_le_bytes([pair[0], pair[1]]))
+            .collect::<Vec<u16>>();
+        String::from_utf16_lossy(&file_content)
+    } else {
+        // Otherwise, treat the content as UTF-8; our test setups do not include UTF-16
+        // BE encoded .inx files.
+        String::from_utf8_lossy(&file_content).to_string()
+    };
```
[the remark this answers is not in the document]
krishnakumar4a4: https://github.com/microsoft/windows-drivers-rs/pull/520#discussion_r2410247801

The test setups we have contain only valid UTF16-LE encoded inx files. odd number of bytes in these inx files are invalid and should be caught with a panic.

crates/cargo-wdk/src/main.rs
krishnakumar4a4: As discussed, since it’s not possible to reference this method if moved to the integration tests folder (`tests`), I’ve moved it to src/test_utils.rs. 
A long-term strategy could be to create a `test-utils` crate that hosts common test utility methods, which can be shared across all workspace member crates to avoid duplication.

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -175,6 +202,37 @@ fn assert_file_hash(path1: &str, path2: &str) {
...
+
+    // Example: DriverVer = 09/13/2023,1.0.0.0
+    let driver_version_regex = driver_version.unwrap_or(r"\d+\.\d+\.\d+\.\d+");
+    let re = regex::Regex::new(&format!(
+        r"^DriverVer\s+=\s+\d+/\d+/\d+,{driver_version_regex}$"
+    ))
+    .unwrap();
```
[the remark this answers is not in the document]
krishnakumar4a4: This is inside the test scope, and the driver_version variable is set as part of the test, so it shouldn’t be a problem. However, I’ve added a regex escape step since it was trivial to include.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -313,9 +314,19 @@ impl<'a> PackageTask<'a> {
...
-            "-v",
-            "*",
         ];
+
+        match std::env::var(STAMPINF_VERSION_ENV_VAR) {
+            Ok(version) if !version.trim().is_empty() => {
+                // When STAMPINF_VERSION is set we intentionally omit -v so stampinf reads it
+                // and populates DriverVer.
+                debug!("Using {STAMPINF_VERSION_ENV_VAR} env var to set DriverVer: {version}");
+            }
```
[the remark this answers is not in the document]
krishnakumar4a4: Fixed it.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -313,9 +314,22 @@ impl<'a> PackageTask<'a> {
...
-            "*",
         ];
+
+        match std::env::var(STAMPINF_VERSION_ENV_VAR) {
+            Ok(version) if !version.trim().is_empty() => {
+                // When STAMPINF_VERSION is set we intentionally omit -v so stampinf reads it
+                // and populates DriverVer.
+                debug!(
+                    DriverVer = version,
+                    "Using {STAMPINF_VERSION_ENV_VAR} env var to set DriverVer"
```
wmmc88, line 326: Does `tracing` support [named parameters](https://doc.rust-lang.org/std/fmt/index.html#named-parameters) without an argumenl list like this properly? AKA does it properly expand `STAMPINF_VERSION_ENV_VAR`?

crates/cargo-wdk/tests/common.rs
```diff
@@ -33,13 +35,42 @@ pub fn set_crt_static_flag() {
 /// * Panics if the lock file cannot be created.
 /// * Panics if the lock cannot be acquired.
 /// * Panics if the lock cannot be released.
-pub fn with_file_lock<F>(f: F)
+pub fn with_file_lock<K, V, F>(env_vars_key_value_pairs: &[(K, V)], f: F)
```
wmmc88: `with_file_lock` wasn't meant to deal directly with env vars. Why not just have this env setting/resetting behavior in the `f` closure/function?

wmmc88: or you can create a seperate `with_env` function that calls `with_file_lock`. right now forcing with_file_lock to take ine env_vars is awkward for tests that don't touch env vars

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -175,6 +202,38 @@ fn assert_file_hash(path1: &str, path2: &str) {
...
+        fs::read(format!("{package_path}/{driver_name}.inf")).expect("Unable to read inf file");
+    let file_content = if file_content.starts_with(&[0xFF, 0xFE]) {
+        // Handle UTF-16 LE (BOM 0xFF 0xFE).
+        let file_content = file_content
+            .chunks(2)
+            .map(|pair| u16::from_le_bytes([pair[0], pair[1]]))
+            .collect::<Vec<u16>>();
+        String::from_utf16_lossy(&file_content)
+    } else {
+        // Otherwise, treat the content as UTF-8; our test setups do not include UTF-16
```
wmmc88: if we are not testing UTF16 files, we should not have unnecessary UTF16 logic in our test code. just assert your utf8 precondition and move on with assuming utf8.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -631,4 +653,76 @@ mod tests {
...
+
+            let wdk_build = WdkBuild::default();
+            let fs = Fs::default();
+            let mut command_exec = CommandExec::default();
+
+            let expect_skip_v = env_val.is_some() && both_cfgs; // skip -v only in override path
+
+            command_exec
+                .expect_run()
+                .withf(move |cmd: &str, args: &[&str], _, _| {
```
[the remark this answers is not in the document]
krishnakumar4a4: The closure requires captured variables to have a 'static lifetime if they are borrowed. The variable expect_skip_v does not have a 'static lifetime. Therefore, using the move keyword forces a copy of the expect_skip_v boolean which is the right thing to do here.

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -175,6 +202,38 @@ fn assert_file_hash(path1: &str, path2: &str) {
...
+        String::from_utf8_lossy(&file_content).to_string()
+    };
+
+    // Example: DriverVer = 09/13/2023,1.0.0.0
+    let driver_version_regex =
+        driver_version.map_or_else(|| r"\d+\.\d+\.\d+\.\d+".to_string(), regex::escape);
+    let re = regex::Regex::new(&format!(
+        r"^DriverVer\s+=\s+\d+/\d+/\d+,{driver_version_regex}$"
+    ))
+    .unwrap();
```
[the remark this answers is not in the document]
krishnakumar4a4: No, as per [map_or_else documentation](https://doc.rust-lang.org/std/option/enum.Option.html#method.map_or_else), the function provided as a second argument (i.e: `regex::escape`)is only applied when driver_version is Some, otherwise the function provided in the first argument (i.e: `|| r"\d+\.\d+\.\d+\.\d+".to_string()`) is computed and returned. So `regex::escape` won't be applied to first argument.

crates/cargo-wdk/src/actions/build/package_task.rs
krishnakumar4a4: Yes, this is how I see the log 

`DEBUG Using STAMPINF_VERSION env var to set DriverVer DriverVer="1.1.1.1"`

> AKA does it properly expand STAMPINF_VERSION_ENV_VAR?

It appears to expand STAMPINF_VERSION_ENV_VAR to the value STAMPINF_VERSION, and the DriverVer field is printed separately.

> Does tracing support [named parameters](https://doc.rust-lang.org/std/fmt/index.html#named-parameters) without an argumenl list like this properly?

I found a reference in the tracing documentation that supports named parameter-style fields: https://docs.rs/tracing/latest/tracing/#recording-fields

The reason for choosing to print DriverVer as a named field comes from a Copilot suggestion to prevent log injection. Refer to: https://github.com/microsoft/windows-drivers-rs/pull/520#discussion_r2413836362

crates/cargo-wdk/tests/build_command_test.rs
krishnakumar4a4: We do have a couple of test projects with UTF16-LE encoded inx files 
- `crates/cargo-wdk/tests/umdf-driver`
- `crates/cargo-wdk/mixed-package-kmdf-workspace`

krishnakumar4a4:
```
        // Otherwise, treat the content as UTF-8; our test setups do not include UTF-16
        // BE encoded .inx files.
```
The first line of the comment might have given the impression that UTF-16 files did not exist. However, I noticed that the comment was split into two lines at an incorrect place. I'm fixing this now.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -313,9 +314,22 @@ impl<'a> PackageTask<'a> {
...
             "-c",
             &cat_file_path,
-            "-v",
-            "*",
         ];
+
+        match std::env::var(STAMPINF_VERSION_ENV_VAR) {
+            Ok(version) if !version.trim().is_empty() => {
+                // When STAMPINF_VERSION is set we intentionally omit -v so stampinf reads it
+                // and populates DriverVer.
```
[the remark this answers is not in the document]
krishnakumar4a4: Fixed the comment.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -313,9 +314,23 @@ impl<'a> PackageTask<'a> {
...
+                // DriverVer. (Whitespace-only values are ignored.)
+                debug!(
+                    DriverVer = version,
+                    "Using {STAMPINF_VERSION_ENV_VAR} env var to set DriverVer"
+                );
+            }
+            _ => {
+                args.extend(["-v", "*"]);
+            }
+        }
```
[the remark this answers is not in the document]
krishnakumar4a4: I think it's unnecessary to log the trimmed version. It's only meant to filter out empty or whitespace-only values explicitly passed to stampinf.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -313,9 +314,23 @@ impl<'a> PackageTask<'a> {
             &arch,
             "-c",
             &cat_file_path,
-            "-v",
-            "*",
         ];
+
+        match std::env::var(STAMPINF_VERSION_ENV_VAR) {
+            Ok(version) if !version.trim().is_empty() => {
```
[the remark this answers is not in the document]
krishnakumar4a4: https://github.com/microsoft/windows-drivers-rs/pull/520#discussion_r2417771807

crates/cargo-wdk/tests/test_utils/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+        .expect("Unable to create lock file for cargo-wdk tests");
+    FileExt::lock_exclusive(&lock_file).expect("Unable to cargo-wdk-test.lock file");
+    let result = f();
+    FileExt::unlock(&lock_file).expect("Unable to unlock cargo-wdk-test.lock file");
+    result
+}
+
+// FIXME: For whatever reason, clippy flags this function as dead code, even
+// though it is used in build_command_test.rs
+#[allow(dead_code)]
```
krishnakumar4a4: Callout: This looks like a clippy issue which is treating with_env as unused even though it being used inside build_command_test.rs

Error reference: https://github.com/microsoft/windows-drivers-rs/actions/runs/18387336798/job/52388944528?pr=520

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -175,6 +201,38 @@ fn assert_file_hash(path1: &str, path2: &str) {
...
+        String::from_utf16_lossy(&file_content)
+    } else {
+        // Otherwise, treat the content as UTF-8; our test setups do not include
+        // UTF16-BE encoded .inx files.
+        String::from_utf8_lossy(&file_content).to_string()
+    };
+
+    // Example: DriverVer = 09/13/2023,1.0.0.0
+    let driver_version_regex =
+        driver_version.map_or_else(|| r"\d+\.\d+\.\d+\.\d+".to_string(), regex::escape);
```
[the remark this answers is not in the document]
krishnakumar4a4: Nope.

crates/cargo-wdk/tests/common.rs
krishnakumar4a4: Makes sense. I’ve modified with_env to call with_file_lock, while still allowing with_file_lock to be used independently. Code changes have been made accordingly.

crates/cargo-wdk/tests/_test_utils.rs
```diff
@@ -0,0 +1,109 @@
+//! Common methods for tests.
+
+#![allow(clippy::literal_string_with_formatting_args)]
```
gurry: Better to apply this on the specific location of the issue rather than the entire file.

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -1,28 +1,28 @@
 //! System level tests for cargo wdk build flow
 #![allow(clippy::literal_string_with_formatting_args)]
-mod common;
+mod _test_utils;
```
gurry: Does this mod have the underscore in the name to avoid "unused" warnings?

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -1,28 +1,28 @@
 //! System level tests for cargo wdk build flow
 #![allow(clippy::literal_string_with_formatting_args)]
```
gurry: Given that the code in in this file has changed worth checking if we still need this lint supression. If it's still needed, move to that specific location.

crates/cargo-wdk/tests/new_command_test.rs
```diff
@@ -1,11 +1,11 @@
 //! System level tests for cargo wdk new flow
 #![allow(clippy::literal_string_with_formatting_args)]
```
gurry: Move it to the specific location where needed

crates/cargo-wdk/tests/test_utils/mod.rs
gurry: Assuming it is used only in tests, could it be happening because the function is being compiled even when `cfg(test)` is not on?

crates/cargo-wdk/tests/build_command_test.rs
krishnakumar4a4: Yes, it was an attempt to tell Cargo that this file is not an integration test crate (that contains tests), but a utility module (with methods being called from other integration tests). I realised that the name has no role to play, so I’m going to remove the underscore. I am planning to retain the module rename to test_utils as it seems more apt.

One more thing I discovered — and am still exploring — is that changing the structure from `test_utils.rs` to `test_utils/mod.rs` tells Cargo to treat this module as a non-integration test.

crates/cargo-wdk/tests/test_utils/mod.rs
krishnakumar4a4: As far as I understand, all the `.rs` files inside the tests folder are compiled as individual test crates. So, within the test utils module, there are no tests that use these methods, which could be a potential reason for them showing up as dead code.
Oddly, the dead code warning only appears for one method and not for others. I’m still investigating this to see if I can find a solution or determine whether it’s a Clippy bug.

crates/cargo-wdk/tests/_test_utils.rs
krishnakumar4a4: Thanks for pointing this out, I found that this lint is stale and removed it.

crates/cargo-wdk/tests/build_command_test.rs
krishnakumar4a4: https://github.com/microsoft/windows-drivers-rs/pull/520#discussion_r2419035080

crates/cargo-wdk/tests/new_command_test.rs
krishnakumar4a4: https://github.com/microsoft/windows-drivers-rs/pull/520#discussion_r2419035080

crates/cargo-wdk/tests/test_utils/mod.rs
krishnakumar4a4: I identified why this is failing. This particular method is used in only one of the two integration test modules, while the remaining methods are used in both. So, when Clippy is run with --all-targets, it flags dead_code when compiling the other integration test. It appears there’s no way to encode this into a lint suppression, so I’ve added #[allow(dead_code)] with a reason. This suppression can be removed once the other integration test also uses this method.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -313,9 +314,23 @@ impl<'a> PackageTask<'a> {
...
+                // When STAMPINF_VERSION is set to a non-empty, non-whitespace value, we
+                // intentionally omit -v so stampinf reads it and populates
+                // DriverVer. (Whitespace-only values are ignored.)
+                debug!(
+                    DriverVer = version,
+                    "Using {STAMPINF_VERSION_ENV_VAR} env var to set DriverVer"
+                );
```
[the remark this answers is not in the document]
krishnakumar4a4: Nope, `stampinf` directly reads the env var, the idea is to print the same value, but not the trimmed value. The trimming is only meant as a check to decide when to pass `-v *`

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -175,6 +200,38 @@ fn assert_file_hash(path1: &str, path2: &str) {
...
+        String::from_utf16_lossy(&file_content)
+    } else {
+        // Otherwise, treat the content as UTF-8; our test setups do not include
+        // UTF16-BE encoded .inx files.
+        String::from_utf8_lossy(&file_content).to_string()
+    };
+
+    // Example: DriverVer = 09/13/2023,1.0.0.0
+    let driver_version_regex =
+        driver_version.map_or_else(|| r"\d+\.\d+\.\d+\.\d+".to_string(), regex::escape);
```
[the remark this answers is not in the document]
krishnakumar4a4: Nope.

crates/cargo-wdk/src/test_utils.rs
```diff
@@ -0,0 +1,63 @@
...
+/// The environment is guaranteed to be not modified during the execution
+/// of the function, and the environment is reset to its original state
+/// after execution of the function. No testing asserts should be called in
+/// the function, since a failing test will poison the mutex, and cause all
+/// remaining tests to fail.
+///
+/// # Panics
+///
+/// Panics if called with duplicate environment variable keys.
+pub fn with_env<K, V, F, R>(env_vars_key_value_pairs: &[(K, Option<V>)], f: F) -> R
```
leon-xd, line 17: I have a similar function in my PR but this one has more features -- I didn't account for multiple key/env pairs and this one has a mutex. I may borrow this for my PR.

crates/cargo-wdk/src/test_utils.rs
```diff
@@ -0,0 +1,63 @@
...
+            let insert_result = original_env_vars.insert(key, original_value);
+            assert!(
+                insert_result.is_none(),
+                "Duplicate environment variable keys were provided"
+            );
+        }
+
+        // Remove the env var if value is None
+        if let Some(value) = value {
+            std::env::set_var(key, value);
```
leon-xd, line 42: in the 2024 edition these functions are unsafe which I have dealt with by triggering a compile error if not built with windows as a target. given that this PR is likely to go through first I'll update this function on my end

crates/cargo-wdk/src/test_utils.rs
gurry: There are a few functions that are duplicated like this. At some point we may have to put them in a shared crate maybe.

PR #527, opened 2025-10-07
fix: move cargo-wdk output to stderr

### LINKED ISSUE #526: Move `cargo-wdk` output to stderr
`cargo build`, `rustc` and other build tools like `gcc` emit their human readable output to to stderr instead of stdout. This allows stdout to be available for machine readable output (e.g. `cargo build --message-format=json`). I believe we should move `cargo wdk` output to stderr as well. 
 
The rationale is:

- Using stderr is consistent with the aforementioned convention for build tools
- Without it you could end up "splitting" the output when stdout or stderr is redirected to a file. E.g. if you run `cargo wdk build > output.log` then all the cargo wdk lines will go to the file and all the `Compiling ...` lines from `cargo build` will appear on the console and not go into the file. We need to use the same stream as `cargo` to avoid this.
- It leaves the door open for supporting machine-readable output in the future

cnaples79:
## Summary
- Configures `cargo-wdk` to write its output to stderr instead of stdout
- Aligns with standard build tool conventions (cargo, rustc, gcc)

## Rationale
As noted in #526, build tools like `cargo build`, `rustc`, and `gcc` emit their human-readable output to stderr, leaving stdout available for machine-readable output (e.g., `cargo build --message-format=json`). This PR brings `cargo-wdk` in line with this convention.

## Changes
- Modified `crates/cargo-wdk/src/trace.rs` to add `.with_writer(std::io::stderr)` to the `tracing_subscriber` configuration
- This redirects all `info!`, `debug!`, `warn!`, and `error!` macro output to stderr

## Benefits
- **Consistency**: Matches the behavior of other Rust build tools
- **No output splitting**: When redirecting streams (e.g., `cargo wdk build > output.log`), all build output now goes to the same stream as `cargo build` output
- **Future-proof**: Leaves stdout available for potential machine-readable output formats

## Testing
The change uses the standard `tracing_subscriber::fmt::SubscriberBuilder::with_writer` API, which is well-documented and widely used. CI will verify the build passes on supported platforms.

Fixes #526

### DIFF
--- a/crates/cargo-wdk/src/trace.rs
+++ b/crates/cargo-wdk/src/trace.rs
@@ -33,6 +33,7 @@ pub fn init_tracing(verbosity_level: clap_verbosity_flag::Verbosity) {
         .without_time()
         .with_target(false)
         .with_file(false)
+        .with_writer(std::io::stderr)
         .with_env_filter(tracing_filter)
         .init();
 }


gurry:
Thanks for the PR @cnaples79. We will have a look at it in the next few days.

gurry:
@cnaples79 The checks didn't pass bercause many of `cargo-wdk`'s tests are failing after the change. That is because these tests try to verify what's emitted on stdout. They need to be changed to check stderr now.

Another change you may have to make is around the output of the Windows tools that `cargo-wdk` invokes e.g. `stampinf`, `signtool` etc. These tools emit to stdout instead of stderr. We will need to redirect them to stderr as well.

gurry:
> @cnaples79 The checks didn't pass bercause many of `cargo-wdk`'s tests are failing after the change. That is because these tests try to verify what's emitted on stdout. They need to be changed to check stderr now.
> 
> Another change you may have to make is around the output of the Windows tools that `cargo-wdk` invokes e.g. `stampinf`, `signtool` etc. These tools emit to stdout instead of stderr. We will need to redirect them to stderr as well.

Hey @cnaples79. Will you be able to address this? If not, I will close this PR and create a new one fixing the issue.

gurry:
Since there has been no response from the author, closing in favour of https://github.com/microsoft/windows-drivers-rs/pull/599

PR #551, opened 2025-10-15
test: add unit tests for cargo wdk new subcommand action flow

### LINKED ISSUE #405: Add unit tests for cargo-wdk new subcommand action flow
We have unit tests for the Build subcommand action flow. We need to cover unit tests for the new subcommand action flow as well. This issue is about that.

krishnakumar4a4:
Test improvements:

* Added a unit test module to `crates/cargo-wdk/src/trace.rs` that verifies the correct mapping of `Verbosity` levels to Cargo flag options, covering default, quiet, and multiple verbose levels.
* Added unit tests for new action flow

Closes #405

### DIFF
--- a/crates/cargo-wdk/src/actions/new/mod.rs
+++ b/crates/cargo-wdk/src/actions/new/mod.rs
@@ -472,9 +472,9 @@ mod tests {
     }
 
     #[test]
-    fn when_create_inx_file_called_with_invalid_path_then_run_returns_invalid_driver_crate_name() {
+    fn when_invalid_path_provided_then_run_returns_invalid_driver_crate_name() {
         set_up_and_assert(
-            // Use an empty path component so file_name returns None
+            // Use an empty path component so that calling file_name() on it returns None
             Path::new(""),
             DriverType::Kmdf,
             Verbosity::default(),


crates/cargo-wdk/src/trace.rs
```diff
@@ -57,3 +57,32 @@ pub fn get_cargo_verbose_flags<'a>(
...
+    use clap_verbosity_flag::Verbosity;
+
+    #[test]
+    fn map_input_clap_verbosity_flags_to_cargo_flags() {
+        // (incoming verbosity, expected cargo flag Option)
+        let cases = vec![
+            // Input: Default (no -v / -q)
+            (Verbosity::default(), None),
+            // Input: Quiet (-q)
+            (Verbosity::new(0, 1), Some("-q")),
```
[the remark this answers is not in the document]
krishnakumar4a4: Nope, the mapping is correct.

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,476 @@ impl<'a> NewAction<'a> {
...
+        let path = Path::new("test_driver_fail_cargo_new");
+        let driver_type = DriverType::Kmdf;
+        let verbosity_level = Verbosity::default();
+
+        // Set up mocks with failure at cargo new step
+        let test_new_action = TestNewAction::new(path).expect_cargo_new(
+            Some(Output {
+                status: ExitStatus::from_raw(1),
+                stdout: vec![],
+                stderr: "some error".into(),
```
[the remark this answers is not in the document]
krishnakumar4a4: Nope, there is `From<&str>` trait implemented for Vec<u8>. So, calling into should be fine here.

crates/cargo-wdk/src/trace.rs
```diff
@@ -57,3 +57,32 @@ pub fn get_cargo_verbose_flags<'a>(
...
+            // Input: Single verbose (-v)
+            (Verbosity::new(1, 0), Some("-v")),
+            // Input: Double verbose (-vv)
+            (Verbosity::new(2, 0), Some("-vv")),
+            // Input: Triple verbose (-vvv)
+            (Verbosity::new(3, 0), Some("-vv")),
+            // Input: Quadruple verbose (-vvvv)
+            (Verbosity::new(4, 0), Some("-vv")),
```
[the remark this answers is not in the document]
krishnakumar4a4: Here is our expectation and the code matches that

|cli_input|cargo_verbosity_flag|cargo build flag expectation|
|--|--|--|
|no -v/-q| Error level | no -v/-q should be passed|
|-q| Off level | -q should be passed|
|-v| Warn level| -v should be passed|
|-vv| Info level| -vv should be passed|
|-vvv/-vvvv| Debug/Trace level| -vv should be passed|

crates/cargo-wdk/src/trace.rs
```diff
@@ -57,3 +57,32 @@ pub fn get_cargo_verbose_flags<'a>(
...
 }
+
+#[cfg(test)]
+mod tests {
+    use clap_verbosity_flag::Verbosity;
+
+    #[test]
+    fn map_input_clap_verbosity_flags_to_cargo_flags() {
+        // (incoming verbosity, expected cargo flag Option)
+        let cases = vec![
```
gurry, line 68: What happens if someone instantiated `Verbosity` with illegal values e.g. `15` for the first argument or `3` for the second argument?

Do we want to add tests for some of these values?

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,476 @@ impl<'a> NewAction<'a> {
...
+        assert!(
+            matches!(
+                result,
+                Err(NewActionError::FileSystem(FileError::WriteError(_, _)))
+            ),
+            "Expected FileSystem WriteError from copy_cargo_config step"
+        );
+    }
+
+    struct TestNewAction<'a> {
```
gurry:
```suggestion
    struct TestSetup<'a> {
```

`TestSetup` is a clearer name. The problem with `TestNewAction` is that it conveys almost no meaning at all.

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,476 @@ impl<'a> NewAction<'a> {
...
+        providers::{
+            error::{CommandError, FileError},
+            exec::MockCommandExec,
+            fs::MockFs,
+        },
+    };
+
+    // Helper struct for setting up different failure scenarios in the
+    // `update_cargo_toml` step.
+    struct UpdateCargoTomlCaseSetupHelper {
```
gurry:
```suggestion
    struct UpdateCargoTomlTestCase {
```

What this struct seems to hold is pre-conditions and post-conditions of a test. So a good name for it is just `*TestCase`. 

That said, I don't think you need a struct for this. You can just use an unnamed tuple like you have done on line 330.

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,476 @@ impl<'a> NewAction<'a> {
...
+            matches!(
+                result,
+                Err(NewActionError::FileSystem(FileError::WriteError(_, _)))
+            ),
+            "Expected FileSystem WriteError from copy_lib_rs_template"
+        );
+    }
+
+    #[test]
+    fn when_update_cargo_toml_fails_for_multiple_cases_then_returns_filesystem_error() {
```
gurry:
```suggestion
    fn when_update_cargo_toml_fails_then_returns_filesystem_error() {
```

`_for_multiple_cases` doesn't add anything useful

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,476 @@ impl<'a> NewAction<'a> {
...
+        )
+        .run();
+        assert!(
+            matches!(result, Err(NewActionError::CargoNewCommand(_))),
+            "Expected CargoNewCommand error"
+        );
+    }
+
+    #[test]
+    fn when_copy_lib_rs_template_fails_then_returns_filesystem_error() {
```
gurry: Most of these tests seem to be following the same pattern:

1. Declare three variables
2. Set up expectations
3. Create `NewAction`
4. Call `run`
5. Assert `result`

with only steps 2 and 5 varying from test to test. Can you try refactoring all the common stuff into a single method? Then it becomes a lot easier to review as well.

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,476 @@ impl<'a> NewAction<'a> {
...
+    use std::{
+        io::Error,
+        os::windows::process::ExitStatusExt,
+        path::{Path, PathBuf},
+        process::{ExitStatus, Output},
+    };
+
+    use clap_verbosity_flag::Verbosity;
+
+    use super::error::NewActionError;
```
leon-xd: is this able to be imported with the `crate` prefix?

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,476 @@ impl<'a> NewAction<'a> {
...
+        );
+    }
+
+    #[test]
+    fn when_update_cargo_toml_fails_for_multiple_cases_then_returns_filesystem_error() {
+        let path = Path::new("test_driver_fail_cargo_toml_read");
+        let driver_type = DriverType::Kmdf;
+        let verbosity_level = Verbosity::default();
+
+        let cases: [UpdateCargoTomlCaseSetupHelper; 3] = [
```
leon-xd: In addition to Gurinder's comment above, storing each of these as booleans would make more sense if we wanted to test every single combination of the failures and their potential outputs. either we should choose a better way to represent this or test all 8 possible true/false combos

crates/cargo-wdk/src/trace.rs
krishnakumar4a4: The first argument represents the count of verbose flags, and the second indicates the count of quiet flags. An [internal offset](https://github.com/clap-rs/clap-verbosity-flag/blob/v3.0.2/src/lib.rs#L208) is calculated as `first - second`. When this value exceeds 3, it sets the trace level. In our case, we have one such scenario that covers that i.e `(4, 0)`.

crates/cargo-wdk/src/actions/new/mod.rs
krishnakumar4a4: Originally, I used an unnamed tuple and encountered `type_complexity` lint error (screenshot below), which led me to create a struct instead.
Screenshot 2025-10-21 164448

I tried declaring the struct inside the test itself, as it's the only place it's used, but then ran into another clippy error `items_after_statements`. So I had to move the struct declaration to the top of the module.

As discussed, since these are tests, I will go ahead with adding a lint suppression and revert to using unnamed tuples.

crates/cargo-wdk/src/actions/new/mod.rs
krishnakumar4a4: It can be done, I've changed it.

crates/cargo-wdk/src/actions/new/mod.rs
krishnakumar4a4: I'm reverting to using unnamed tuples as discussed here https://github.com/microsoft/windows-drivers-rs/pull/551#discussion_r2447844109

crates/cargo-wdk/src/actions/new/mod.rs
gurry: This lint seems to be about type nesting. Can you try creating a type alias for the function pointer type and use that alias in the tuple? That might stop the warning.

krishnakumar4a4: Yes, copilot also suggested this as well - https://github.com/microsoft/windows-drivers-rs/pull/551#discussion_r2449084198 and I tried it locally; it seemed to work.

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,485 @@ impl<'a> NewAction<'a> {
...
+            let expected_file_to_write = cargo_toml_path.clone();
+            let expected_file_to_append = cargo_toml_path.clone();
+            self.mock_fs
+                .expect_read_file_to_string()
+                .withf(move |path| path == cargo_toml_path)
+                .returning(move |_| {
+                    if is_cargo_toml_read_success {
+                        Ok("[package]\nname = \"test_driver\"\nversion = \
+                            \"0.1.0\"\n\n[dependencies]\n"
+                            .to_string())
```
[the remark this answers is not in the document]
krishnakumar4a4: Updated it.

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,485 @@ impl<'a> NewAction<'a> {
...
+    fn when_update_cargo_toml_fails_then_returns_filesystem_error() {
+        let path = Path::new("test_driver_fail_cargo_toml_read");
+        let driver_type = DriverType::Kmdf;
+        let verbosity_level = Verbosity::default();
+
+        #[allow(
+            clippy::type_complexity,
+            reason = "This suppression is required to avoid creating a type for the below complex \
+                      type declaration."
+        )]
```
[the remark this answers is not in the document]
krishnakumar4a4: This worked, updated it.

crates/cargo-wdk/src/actions/new/mod.rs
krishnakumar4a4: Thanks for the suggestion. I refactored the code to reduce duplications.

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,576 @@ impl<'a> NewAction<'a> {
...
+                    Some(output) => match output.status.code() {
+                        Some(0) => Ok(Output {
+                            status: ExitStatus::from_raw(0),
+                            stdout: vec![],
+                            stderr: vec![],
+                        }),
+                        _ => Err(CommandError::from_output("cargo", &[], &output)),
+                    },
+                    None => Ok(Output {
+                        status: ExitStatus::from_raw(0),
```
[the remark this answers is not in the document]
krishnakumar4a4: Contradicting copilot suggestion https://github.com/microsoft/windows-drivers-rs/pull/551#discussion_r2434430894

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,576 @@ impl<'a> NewAction<'a> {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::{
+        io::Error,
```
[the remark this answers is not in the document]
krishnakumar4a4: Modified imports as below, so that tests can be run on both windows and non-windows hosts. Assuming the majority of non-Windows hosts typically used for builds and tests going to be from Unix family.

```rust
    #[cfg(not(windows))]
    use std::os::unix::process::ExitStatusExt;
    #[cfg(windows)]
    use std::os::windows::process::ExitStatusExt;
```

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,576 @@ impl<'a> NewAction<'a> {
...
+
+        setup_and_assert(
+            path,
+            driver_type,
+            verbosity_level,
+            |test_setup| {
+                // Set up mocks with failure at cargo new step
+                test_setup.set_expectations_with(
+                    Some(FailureStep::CargoNew(Output {
+                        status: ExitStatus::from_raw(1),
```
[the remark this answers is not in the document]
krishnakumar4a4: I think we can keep this. 
Changed imports as mentioned here https://github.com/microsoft/windows-drivers-rs/pull/551#discussion_r2454022787

crates/cargo-wdk/src/actions/new/mod.rs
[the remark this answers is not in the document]
krishnakumar4a4: Contradicting copilot suggestion https://github.com/microsoft/windows-drivers-rs/pull/551#discussion_r2434430894

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,548 @@ impl<'a> NewAction<'a> {
...
+                            PathBuf::from("/some_random_path/Cargo.toml"),
+                            Error::other("Append error"),
+                        ))
+                    }
+                });
+            self
+        }
+
+        fn expect_create_inx_file(mut self, is_create_success: bool) -> Self {
+            let driver_crate_name = self.path.file_name().unwrap().to_string_lossy().to_string();
```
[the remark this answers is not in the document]
krishnakumar4a4: Accepted the suggestion.

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,548 @@ impl<'a> NewAction<'a> {
...
+                        && args.len() >= 5
+                        && args[0] == "new"
+                        && args[1] == "--lib"
+                        && args[2] == expected_path
+                        && args[3] == "--vcs"
+                        && args[4] == "none";
+
+                    expected_flag
+                        .clone()
+                        .map_or(matched, |flag| matched && args.len() > 5 && args[5] == flag)
```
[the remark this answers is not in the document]
krishnakumar4a4: Accepted the suggestion.

crates/cargo-wdk/src/trace.rs
gurry: Okay, can you add another one which does not have a `0` in it because `0` is an edge condition as well?

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,553 @@ impl<'a> NewAction<'a> {
...
+                        Err(NewActionError::FileSystem(FileError::WriteError(_, _)))
+                    ),
+                    "Expected FileSystem WriteError from copy_lib_rs_template"
+                );
+            },
+        );
+    }
+
+    /// Type alias for assertion functions used in tests.
+    type AssertionFn = fn(Result<(), NewActionError>);
```
gurry: Slightly better if you move this inside the only function where it is used, but no big deal. I'm fine with it here as well.

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,553 @@ impl<'a> NewAction<'a> {
...
+    }
+
+    /// Helper function to set up mock expectations and assert on the result.
+    ///
+    /// This function takes a closure to configure the test setup (e.g., mock
+    /// expectations) and another closure to perform assertions on the
+    /// result of running the action. Usage: pass a closure to
+    /// `set_expectations_fn` to configure mocks, and a closure to `assert_fn`
+    /// to check the outcome.
+    fn setup_and_assert(
```
gurry:
```suggestion
    fn set_up_and_assert(
```
Nit: "setup" is a noun and "set up" is a verb.

crates/cargo-wdk/src/trace.rs
krishnakumar4a4: I missed to address this comment, I have added an additional case - `(10, 5)` for this now.

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,552 @@ impl<'a> NewAction<'a> {
...
+            mut self,
+            failure_step: Option<FailureStep>,
+            expected_flag: Option<String>,
+        ) -> Self {
+            if let Some(FailureStep::CargoNew(override_output)) = failure_step {
+                return self.expect_cargo_new(Some(override_output), expected_flag);
+            }
+            self = self.expect_cargo_new(None, expected_flag);
+
+            if matches!(failure_step, Some(FailureStep::CopyLibRsTemplate)) {
```
gurry, line 634: Nit: use `if let Some(...) = failure_step` for all these checks instead of `matches!`

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,552 @@ impl<'a> NewAction<'a> {
...
+
+        fn expect_copy_lib_rs_template(mut self, is_copy_success: bool) -> Self {
+            let lib_rs_path = self.path.join("src").join("lib.rs");
+            self.mock_fs
+                .expect_write_to_file()
+                .withf(move |path, _| path == lib_rs_path)
+                .returning(move |_, _| {
+                    if !is_copy_success {
+                        return Err(FileError::WriteError(
+                            PathBuf::from("/some_random_path/src/lib.rs"),
```
gurry: Can we use the src path passed to the function here instead of a random path?

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,552 @@ impl<'a> NewAction<'a> {
...
+            let expected_file_to_append = cargo_toml_path.clone();
+            self.mock_fs
+                .expect_read_file_to_string()
+                .withf(move |path| path == cargo_toml_path)
+                .returning(move |_| {
+                    if is_cargo_toml_read_success {
+                        Ok(r#"[package]
+                               name = "test_driver"
+                               version = "0.1.0"
+                              [dependencies]
```
gurry, line 744: Maybe add a crate or two under `[dependencies]` as well to make it bit more realistic

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,552 @@ impl<'a> NewAction<'a> {
...
+                });
+            self.mock_fs
+                .expect_write_to_file()
+                .withf(move |path, content| path == expected_file_to_write && !content.is_empty())
+                .returning(move |_, _| {
+                    if is_dep_section_removal_success {
+                        Ok(())
+                    } else {
+                        Err(FileError::WriteError(
+                            PathBuf::from("/some_random_path/Cargo.toml"),
```
gurry: Maybe use the actual path passed to the write function. Do this at all other similar places as well.

crates/cargo-wdk/src/actions/new/mod.rs
krishnakumar4a4: 1. One intention was to demonstrate that the path value does not influence the logic of the run method.
2. Another, less ideal reason within the test scope: since `lib_rs_path` variable has to be moved into the mock method rather than by reference, I’d need to clone it again — which I was trying to avoid.

crates/cargo-wdk/src/actions/new/mod.rs
krishnakumar4a4: The method response we are trying to mock here is supposed to read the `Cargo.toml` file created by `cargo new` command.

The content of such Cargo.toml will look like the below
```toml
[package]
name = "plain-tmp-cargo-project"
version = "0.1.0"
edition = "2024"

[dependencies]
```
I could add the missing edition part to make it more realistic.

crates/cargo-wdk/src/actions/new/mod.rs
krishnakumar4a4: This is a clippy suggestion to use matches! here

```
error: this pattern matching can be expressed using `matches!`
   --> crates\cargo-wdk\src\actions\new\mod.rs:634:16
    |
634 |             if let Some(FailureStep::CopyLibRsTemplate) = failure_step {
    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `matches!(failure_step, Some(FailureStep::CopyLibRsTemplate))`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#equatable_if_let
  ```
  
Looking at documentation, I guess the reason is most likely to avoid `if let` where a simple match is possible.

File: crates/cargo-wdk/src/actions/new/mod.rs (hunk 22 above)
Comment (krishnakumar4a4) [reply crates/cargo-wdk/src/actions/new/mod.rs:744]: Added the edition entry.

File: crates/cargo-wdk/src/actions/new/mod.rs (hunk 23 above)
Comment (krishnakumar4a4) [reply crates/cargo-wdk/src/actions/new/mod.rs]: https://github.com/microsoft/windows-drivers-rs/pull/551#discussion_r2489040475

File: crates/cargo-wdk/src/actions/new/mod.rs (hunk 21 above)
Comment (gurry) [reply crates/cargo-wdk/src/actions/new/mod.rs]: It's just that it makes anyone reading the code wonder why we're returning a random path. 

The actual commands will never return a wrong path so testing for wrong path not influencing the logic is not useful. Cloing is fine in tests if you can't avoid it.

File: crates/cargo-wdk/src/actions/new/mod.rs (hunk 20 above)
Comment (gurry) [reply crates/cargo-wdk/src/actions/new/mod.rs:634]: Okay cool. Let's keep it

File: crates/cargo-wdk/src/actions/new/mod.rs [hunk 24]
```diff
@@ -291,3 +291,563 @@ impl<'a> NewAction<'a> {
...
+                });
+            self
+        }
+
+        fn expect_copy_build_rs_template(mut self, is_copy_success: bool) -> Self {
+            let build_rs_path = self.path.join("build.rs");
+            let expected_build_rs_path = self.path.join("build.rs");
+            self.mock_fs
+                .expect_write_to_file()
+                .withf(move |path, _| path == expected_build_rs_path)
```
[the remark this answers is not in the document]
krishnakumar4a4: we cannot do `build_rs_path.clone()` as it attempts to move `self.path` into the closure as it is not a Copy type.

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,563 @@ impl<'a> NewAction<'a> {
...
+                });
+            self
+        }
+
+        fn expect_copy_cargo_config(mut self, is_copy_success: bool) -> Self {
+            let cargo_config_path = self.path.join(".cargo").join("config.toml");
+            let expected_cargo_config_path = self.path.join(".cargo").join("config.toml");
+            self.mock_fs
+                .expect_write_to_file()
+                .withf(move |path, _| path == expected_cargo_config_path)
```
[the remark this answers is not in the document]
krishnakumar4a4: Similar reason as this https://github.com/microsoft/windows-drivers-rs/pull/551#discussion_r2490311228

gurry (approved):

svasista-ms (approved):
LGTM

leon-xd (approved):

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -291,3 +291,561 @@ impl<'a> NewAction<'a> {
...
+                .expect_read_file_to_string()
+                .withf(move |path| path == expected_file_to_read)
+                .returning(move |_| {
+                    if is_cargo_toml_read_success {
+                        Ok(r#"[package]
+                               name = "test_driver"
+                               version = "0.1.0"
+                               edition = "2024"
+                              [dependencies]
+                              "#
```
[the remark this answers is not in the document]
krishnakumar4a4: This change affects readability of test code. The current formatting is fine since it doesn't affect correctness.

gurry (approved):

svasista-ms (approved):

Issue #252, opened 2024-12-10
Support `--target-dir` in cargo make invocations


wmmc88:
Currently, this is not supported for 2 reasons:
1.  the option is not currently parsed by our cargo-make plugin: https://github.com/microsoft/windows-drivers-rs/blob/10a8e3742b7834e816f6a2ce630b46533bcec357/crates/wdk-build/src/cargo_make.rs#L322

This is a simple enhancement^

2. The implementation of find_top_level_cargo_manifest does not work when a `--target-dir` is supplied: https://github.com/microsoft/windows-drivers-rs/blob/10a8e3742b7834e816f6a2ce630b46533bcec357/crates/wdk-build/src/lib.rs#L875

AFAIU, this 2nd issue is blocked by having support for mutually-exclusive cargo features. The pre-RFC for that idea is [here](https://internals.rust-lang.org/t/pre-rfc-mutually-excusive-global-features/19618). See previous discussion on this issue: https://github.com/rust-lang/cargo/issues/13484 

For the time being, as far as i can tell, there is no path forward to support `--target-dir` builds until progress is made on mutually exclusive cargo features

PR #554, opened 2025-10-21
fix: resolve target directory lookup for emulated cargo-make workspaces

yuvald-sweet-security:
Assuming that the target directory is in `${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}\target` is incorrect since `cargo make` has a feature called "emulated workspaces" where you build a workspace-like structure with `cargo make` that's built out of ordinary cargo crates (where each has it's own root and target directory), and then that variable points to the *cargo make workspace* (which is not a cargo workspace, and therefore has no target directory). Using the `CARGO_MAKE_CRATE_TARGET_DIRECTORY` variable is a better way to locate the current build's target directory.

### DIFF
--- a/crates/wdk-build/src/cargo_make.rs
+++ b/crates/wdk-build/src/cargo_make.rs
@@ -56,6 +56,7 @@ const CARGO_MAKE_CARGO_PROFILE_ENV_VAR: &str = "CARGO_MAKE_CARGO_PROFILE";
 const CARGO_MAKE_CRATE_TARGET_TRIPLE_ENV_VAR: &str = "CARGO_MAKE_CRATE_TARGET_TRIPLE";
 const CARGO_MAKE_CRATE_CUSTOM_TRIPLE_TARGET_DIRECTORY_ENV_VAR: &str =
     "CARGO_MAKE_CRATE_CUSTOM_TRIPLE_TARGET_DIRECTORY";
+const CARGO_MAKE_CRATE_TARGET_DIRECTORY_ENV_VAR: &str = "CARGO_MAKE_CRATE_TARGET_DIRECTORY";
 const CARGO_MAKE_RUST_DEFAULT_TOOLCHAIN_ENV_VAR: &str = "CARGO_MAKE_RUST_DEFAULT_TOOLCHAIN";
 const CARGO_MAKE_CRATE_NAME_ENV_VAR: &str = "CARGO_MAKE_CRATE_NAME";
 const CARGO_MAKE_CRATE_FS_NAME_ENV_VAR: &str = "CARGO_MAKE_CRATE_FS_NAME";
@@ -816,8 +817,9 @@ pub fn copy_to_driver_package_folder<P: AsRef<Path>>(path_to_copy: P) -> Result<
 ///
 /// # Panics
 ///
-/// This function will panic if the `CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY`
-/// environment variable is not set
+/// This function will panic if neither the
+/// `CARGO_MAKE_CRATE_TARGET_DIRECTORY` nor the
+/// `CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY` environment variables are set.
 pub fn load_rust_driver_makefile() -> Result<(), ConfigError> {
     load_wdk_build_makefile(RUST_DRIVER_MAKEFILE_NAME)
 }
@@ -840,8 +842,9 @@ pub fn load_rust_driver_makefile() -> Result<(), ConfigError> {
 ///
 /// # Panics
 ///
-/// This function will panic if the `CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY`
-/// environment variable is not set
+/// This function will panic if neither the
+/// `CARGO_MAKE_CRATE_TARGET_DIRECTORY` nor the
+/// `CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY` environment variables are set.
 pub fn load_rust_driver_sample_makefile() -> Result<(), ConfigError> {
     load_wdk_build_makefile(RUST_DRIVER_SAMPLE_MAKEFILE_NAME)
 }
@@ -870,8 +873,9 @@ pub fn load_rust_driver_sample_makefile() -> Result<(), ConfigError> {
 ///
 /// # Panics
 ///
-/// This function will panic if the `CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY`
-/// environment variable is not set
+/// This function will panic if neither the
+/// `CARGO_MAKE_CRATE_TARGET_DIRECTORY` nor the
+/// `CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY` environment variables are set.
 #[instrument(level = "trace")]
 fn load_wdk_build_makefile<S: AsRef<str> + AsRef<Utf8Path> + AsRef<Path> + fmt::Debug>(
     makefile_name: S,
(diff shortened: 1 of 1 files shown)


yuvald-sweet-security:

gurry:
Thanks for the PR @yuvald-sweet-security. We will look into it soon.

yuvald-sweet-security:
@gurry no problem, also note that I ended up falling back to `CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY` if `CARGO_MAKE_CRATE_TARGET_DIRECTORY` doesn't exist because it seems that cargo make doesn't always set `CARGO_MAKE_CRATE_TARGET_DIRECTORY` early enough (but weirdly enough, it does set that variable when the crates are built through emulated workspaces)

wmmc88 (changes requested):
Thanks for submitting a PR! What is desired behaviour you're looking for here @yuvald-sweet-security?

This change wouldn't affect that fact that custom target directories for the actual builds are not supported and will probably not be supported for quite some time. See https://github.com/microsoft/windows-drivers-rs/issues/252 (especailly the 2nd point).

> CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY doesn't necessarily point to the root of the cargo workspace, since cargo make has a feature called "emulated workspaces" where you build a workspace-like structure with cargo make that's built out of ordinary cargo crates (where each has it's own root and target directory), and then that variable points to the cargo make workspace (which is not a cargo workspace, and therefore has no target directory)

In this emulated workspace scenario, the existing behavior is desirable since it means that the makefile only needs to be copied to one place shared by all members of the emulated workspace. 

Maybe I'm missing it, but I don't see the point of the change proposed here. Could you mock up some examples of something failing beforehand, and then passing after this change?

yuvald-sweet-security:
@wmmc88 maybe I wasn't clear; the current behavior with regards to emulated workspaces is simply broken - it doesn't build because it cannot deduce the correct target directory. For example, with this directory structure:

`Makefile.toml`:
```toml
[env]
# this tells cargo-make that this directory acts as a workspace root
CARGO_MAKE_WORKSPACE_EMULATION = true

# a list of crate members. since we do not have a Cargo.toml, we will need to specify this in here.
CARGO_MAKE_CRATE_WORKSPACE_MEMBERS = ["um", "km"]
```

`um`:
standard Rust binary project initialized with `cargo new um` from the root directory

`km`:
standard WDK project initialized with `cargo wdk new --wdm km` from the root directory

`km/Makefile.toml`:
```toml
extend = "target/rust-driver-makefile.toml"

[config]
load_script = '''
#!@rust
//! ```cargo
//! [dependencies]
//! wdk-build = "0.4.0"
//! ```
#![allow(unused_doc_comments)]

wdk_build::cargo_make::load_rust_driver_makefile()?
'''
```

Running `cargo make` **from the root directory** fails with this error:
 ```
[cargo-make][1] INFO - Build Done in 1.80 seconds.
[cargo-make][1] INFO - Execute Command: "rust-script" "target\\_cargo_make_temp\\persisted_scripts\\c6f46e81df625caf4dfbb235d91efefcff124b65a2b99ad0e7baff59898dd02a.rs"
Error: IoError(Os { code: 3, kind: NotFound, message: "The system cannot find the path specified." })
[cargo-make][1] ERROR - Unable to execute rust code.
[cargo-make][1] WARN - Build Failed.
[cargo-make] ERROR - Error while running duckscript: Source: Unknown Line: 8 - Error while executing command, exit code: 1
[cargo-make] WARN - Build Failed.
```

This is because `CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY` is now set to the **root directory**, which has no `target`. However, even if we go and manually create a `target` directory at the root, we fail with the following error:
```
[cargo-make][1] INFO - Execute Command: "rust-script" "target\\_cargo_make_temp\\persisted_scripts\\c6f46e81df625caf4dfbb235d91efefcff124b65a2b99ad0e7baff59898dd02a.rs"
[cargo-make][1] ERROR - Descriptor file: ".\\target/rust-driver-makefile.toml" not found.
[cargo-make][1] WARN - Build Failed.
[cargo-make] ERROR - Error while running duckscript: Source: Unknown Line: 8 - Error while executing command, exit code: 1
[cargo-make] WARN - Build Failed.
``` 
This is because wdk-build placed the auto-generated makefile into `target/rust-driver-makefile.toml` but the `Makefile.toml` inside the WDK project requests extending `target/rust-driver-makefile.toml` relative to itself, i.e. `km/target/rust-driver-makefile.toml` which doesn't exist.

I don't think that sharing `rust-driver-makefile.toml` between emulated workspaces is a good idea, because they might be different projects or different drivers with different versions of that makefile and there's no prima facie reason to force them to share the same build logic. However, if you wish to take that approach, then we still somehow need to tell `km/Makefile.toml` to extend a makefile from the workspace root - setting it to `../target/rust-driver-makefile.toml` breaks building the project without the workspace (i.e. running `cargo make` inside `km`) and it doesn't seem that the `extend` property allows evaluation of environment variables, and also create the `target` directory if it doesn't exist in `load_wdk_build_makefile`.

wmmc88:
> @wmmc88 maybe I wasn't clear; the current behavior with regards to emulated workspaces is simply broken - it doesn't build because it cannot deduce the correct target directory. For example, with this directory structure:
> 
> `Makefile.toml`:
> 
> ```toml
> [env]
> # this tells cargo-make that this directory acts as a workspace root
> CARGO_MAKE_WORKSPACE_EMULATION = true
> 
> # a list of crate members. since we do not have a Cargo.toml, we will need to specify this in here.
> CARGO_MAKE_CRATE_WORKSPACE_MEMBERS = ["um", "km"]
> ```
> 
> `um`: standard Rust binary project initialized with `cargo new um` from the root directory
> 
> `km`: standard WDK project initialized with `cargo wdk new --wdm km` from the root directory
> 
> `km/Makefile.toml`:
> 
> ```toml
> extend = "target/rust-driver-makefile.toml"
> 
> [config]
> load_script = '''
> #!@rust
> //! ```cargo
> //! [dependencies]
> //! wdk-build = "0.4.0"
> //! ```
> #![allow(unused_doc_comments)]
> 
> wdk_build::cargo_make::load_rust_driver_makefile()?
> '''
> ```
> 
> Running `cargo make` **from the root directory** fails with this error:
> 
> ```
> [cargo-make][1] INFO - Build Done in 1.80 seconds.
> [cargo-make][1] INFO - Execute Command: "rust-script" "target\\_cargo_make_temp\\persisted_scripts\\c6f46e81df625caf4dfbb235d91efefcff124b65a2b99ad0e7baff59898dd02a.rs"
> Error: IoError(Os { code: 3, kind: NotFound, message: "The system cannot find the path specified." })
> [cargo-make][1] ERROR - Unable to execute rust code.
> [cargo-make][1] WARN - Build Failed.
> [cargo-make] ERROR - Error while running duckscript: Source: Unknown Line: 8 - Error while executing command, exit code: 1
> [cargo-make] WARN - Build Failed.
> ```
> 
> This is because `CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY` is now set to the **root directory**, which has no `target`. However, even if we go and manually create a `target` directory at the root, we fail with the following error:
> 
> ```
> [cargo-make][1] INFO - Execute Command: "rust-script" "target\\_cargo_make_temp\\persisted_scripts\\c6f46e81df625caf4dfbb235d91efefcff124b65a2b99ad0e7baff59898dd02a.rs"
> [cargo-make][1] ERROR - Descriptor file: ".\\target/rust-driver-makefile.toml" not found.
> [cargo-make][1] WARN - Build Failed.
> [cargo-make] ERROR - Error while running duckscript: Source: Unknown Line: 8 - Error while executing command, exit code: 1
> [cargo-make] WARN - Build Failed.
> ```
> 
> This is because wdk-build placed the auto-generated makefile into `target/rust-driver-makefile.toml` but the `Makefile.toml` inside the WDK project requests extending `target/rust-driver-makefile.toml` relative to itself, i.e. `km/target/rust-driver-makefile.toml` which doesn't exist.
> 
> I don't think that sharing `rust-driver-makefile.toml` between emulated workspaces is a good idea, because they might be different projects or different drivers with different versions of that makefile and there's no prima facie reason to force them to share the same build logic. However, if you wish to take that approach, then we still somehow need to tell `km/Makefile.toml` to extend a makefile from the workspace root - setting it to `../target/rust-driver-makefile.toml` breaks building the project without the workspace (i.e. running `cargo make` inside `km`) and it doesn't seem that the `extend` property allows evaluation of environment variables, and also create the `target` directory if it doesn't exist in `load_wdk_build_makefile`.

thanks for the thorough explanation.

So this PR is not about support custom target directories, its about fixing loading the driver makefile for emulated cargo-make workspaces. Could you update PR title and description as such.

> also note that I ended up falling back to CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY if CARGO_MAKE_CRATE_TARGET_DIRECTORY doesn't exist because it seems that cargo make doesn't always set CARGO_MAKE_CRATE_TARGET_DIRECTORY

I'd like to understand in what situations `CARGO_MAKE_CRATE_TARGET_DIRECTORY` does not exist here. I would rather not need to have fallbacks in place when determining paths from cargo-make variables- it can make debugging issues more difficult

yuvald-sweet-security:
@wmmc88 done

>I'd like to understand in what situations CARGO_MAKE_CRATE_TARGET_DIRECTORY does not exist here. I would rather not need to have fallbacks in place when determining paths from cargo-make variables- it can make debugging issues more difficult

I would also, but alas I'm not a developer of cargo-make, I'm only trying to get a build of a driver working

wmmc88:
> @wmmc88 done
> 
> > I'd like to understand in what situations CARGO_MAKE_CRATE_TARGET_DIRECTORY does not exist here. I would rather not need to have fallbacks in place when determining paths from cargo-make variables- it can make debugging issues more difficult
> 
> I would also, but alas I'm not a developer of cargo-make, I'm only trying to get a build of a driver working

Okay. i did some digging and it looks like it doesnt get set at the time of load_script. fix the pipeline failure, and ill merge the fix

yuvald-sweet-security:
@wmmc88 should be fixed now

PR #556, opened 2025-10-28
ci: Add code coverage

gurry:
This PR enables code coverage for tests and adds the logic to upload coverage data to [CodeCov](https://about.codecov.io/).

## Scope
Coverage is currently limited only to tests in the top level workspace. It does not include tests from the `tests` folder because they primarily verify macro expansion which is hard to get coverage for. We could focus on them at a later point.

Also we have enabled coverage only for x64 because [Windows arm64 is currently not supported by `cargo-llvm-cov`](https://github.com/taiki-e/cargo-llvm-cov/issues/436), which is the tool we are using.

## Changes
Apart from changing the `test.yaml` GitHub action, I also had to make some changes in `cargo-wdk` tests. It was to make sure Cargo and Rust environment variables from the `cargo llvm-cov` process do not get inherited by `cargo wdk build` process because they can interfere with its behaviour. For instance, `cargo llvm-cov` sets an environment variable to enable instrumentation in the code being built. If it is allowed to be inherited by `cargo wdk build` instrumentation will be enabled for the test driver projects being built as well which will impact the test. 

This argument can be generalized and extended to _any_ Cargo or Rust variables. Therefore, before calling `cargo-wdk` in the tests we now make sure all Cargo and Rust variables are removed from its environment.

## Results

To see what code coverage will look like after this PR, see [this example from an earlier run]( https://app.codecov.io/gh/gurry/windows-drivers-rs/tree/add-code-coverage).

Closes https://github.com/microsoft/windows-drivers-rs/issues/431

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -37,7 +37,8 @@ jobs:
         rust_toolchain:
           - stable
           - beta
-          - nightly
+          # Pin nightly version temporarily due to bug in macrotest: https://github.com/eupn/macrotest/issues/131
+          - nightly-2025-11-20
 
         cargo_profile:
           - dev
@@ -99,6 +100,12 @@ jobs:
       # See issue: https://github.com/taiki-e/cargo-llvm-cov/issues/436
       - name: Run Cargo Test with Coverage (workspace)
         if: matrix.runner.arch != 'arm64'
+        env:
+          # cargo-wdk new tests try to build newly generated
+          # driver projects which can fail in release-plz PRs
+          # because the dependencies are not yet published.
+          # This env var skips such builds for release-plz PRs
+          SKIP_BUILD_IN_CARGO_WDK_NEW_TESTS: ${{ startsWith(github.head_ref, 'release-plz-') && '1' || '' }}
         run: cargo +${{ matrix.rust_toolchain }} llvm-cov --workspace --codecov --output-path target/codecov.info --locked --profile ${{ matrix.cargo_profile }} --all-features
 
       - name: Upload Coverage to Codecov
@@ -111,6 +118,9 @@ jobs:
 
       - name: Run Cargo Test (workspace)
         if: matrix.runner.arch == 'arm64'
+        env:
+          # Skips build in cargo-wdk new tests for release-plz PRs
+          SKIP_BUILD_IN_CARGO_WDK_NEW_TESTS: ${{ startsWith(github.head_ref, 'release-plz-') && '1' || '' }}
         run: cargo +${{ matrix.rust_toolchain }} test --locked --profile ${{ matrix.cargo_profile }} --all-features
 
       - name: Run Cargo Test (Top-Level tests Folder via Cargo Make)
--- a/crates/cargo-wdk/tests/new_command_test.rs
+++ b/crates/cargo-wdk/tests/new_command_test.rs
@@ -59,65 +59,26 @@ fn help_works() {
 }
 
 fn project_is_created(driver_type: &str) {
-    let (stdout, _stderr) = create_and_build_new_driver_project(driver_type);
-    assert!(
-        stdout.contains(
-            "Required directive Provider missing, empty, or invalid in [Version] section."
-        )
-    );
-    assert!(
-        stdout
-            .contains("Required directive Class missing, empty, or invalid in [Version] section.")
-    );
-    assert!(
-        stdout
-            .contains("Invalid ClassGuid \"\", expecting {XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX}.")
-    );
-    assert!(stdout.contains("INF is NOT VALID"));
-}
+    let tmp_dir = TempDir::new().expect("Unable to create new temp dir for test");
+    let project_path = verify_project_creation(driver_type, &tmp_dir);
 
-fn test_command_invocation<F: FnOnce(&str, &str)>(
-    args: &[&str],
-    add_path_arg: bool,
-    command_succeeds: bool,
-    assert: F,
-) {
-    let mut args = args
-        .iter()
-        .map(ToString::to_string)
-        .collect::<Vec<String>>();
-    args.insert(0, String::from("new"));
-
-    if add_path_arg {
-        let driver_name = "test-driver";
-        let tmp_dir = TempDir::new().expect("Unable to create new temp dir for test");
-        println!("Temp dir: {}", tmp_dir.path().display());
-        let driver_path = tmp_dir.join(driver_name);
-        args.push(driver_path.to_string_lossy().to_string());
-    }
-
-    let mut cmd = Command::cargo_bin("cargo-wdk").expect("unable to find cargo-wdk binary");
-    cmd.args(args);
-
-    let cmd_assertion = cmd.assert();
-    let cmd_assertion = if command_succeeds {
-        cmd_assertion.success()
+    // Build the project only if SKIP_BUILD_IN_CARGO_WDK_NEW_TESTS is not set.
+    // This env var is used in release-plz PRs, wherein it is set to skip the
+    // project build because it would fail due to not yet released
+    // dependencies
+    if std::env::var("SKIP_BUILD_IN_CARGO_WDK_NEW_TESTS").unwrap_or_default() == "1" {
+        println!(
+            "Skipping driver build due to SKIP_BUILD_IN_CARGO_WDK_NEW_TESTS environment variable"
+        );
     } else {
-        cmd_assertion.failure()
-    };
-    let output = cmd_assertion.get_output();
-    let stdout = String::from_utf8_lossy(&output.stdout);
-    println!("stdout: {stdout}");
-    let stderr = String::from_utf8_lossy(&output.stderr);
-    println!("stderr: {stderr}");
-
-    assert(&stdout, &stderr);
+        verify_project_build(&project_path);
+    }
 }
 
-fn create_and_build_new_driver_project(driver_type: &str) -> (String, String) {
+fn verify_project_creation(driver_type: &str, tmp_dir: &TempDir) -> PathBuf {
     let driver_name = format!("test-{driver_type}-driver");
     let driver_name_underscored = driver_name.replace('-', "_");
-    let tmp_dir = TempDir::new().expect("Unable to create new temp dir for test");
+
     println!("Temp dir: {}", tmp_dir.path().display());
 
     let driver_path = tmp_dir.join(driver_name.clone());
@@ -195,15 +156,69 @@ fn create_and_build_new_driver_project(driver_type: &str) -> (String, String) {
         .child(driver_name_path.join(".cargo").join("config.toml"))
         .assert(predicates::str::contains("target-feature=+crt-static"));
 
+    driver_path
+}
+
+fn verify_project_build(path: &std::path::Path) {
     // assert if cargo wdk build works on the created driver project
-    let driver_path = tmp_dir.join(&driver_name); // Root dir for tests
-    let mut cmd = create_cargo_wdk_cmd("build", None, Some(&driver_path));
+    let mut cmd = create_cargo_wdk_cmd("build", None, Some(path));
+
     let cmd_assertion = cmd.assert().failure();
-    tmp_dir
-        .close()
-        .expect("Unable to close temp dir after test");
     let output = cmd_assertion.get_output();
     let stdout: String = String::from_utf8_lossy(&output.stdout).into();
-    let stderr: String = String::from_utf8_lossy(&output.stderr).into();
-    (stdout, stderr)
+
+    // Assert build output contains expected errors (the INF file is intentionally
+    // incomplete)
+    assert!(
+        stdout.contains(
+            "Required directive Provider missing, empty, or invalid in [Version] section."
+        )
+    );
+    assert!(
+        stdout
+            .contains("Required directive Class missing, empty, or invalid in [Version] section.")
+    );
+    assert!(
+        stdout
+            .contains("Invalid ClassGuid \"\", expecting {XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX}.")
+    );
+    assert!(stdout.contains("INF is NOT VALID"));
+}
+
+fn test_command_invocation<F: FnOnce(&str, &str)>(
+    args: &[&str],
+    add_path_arg: bool,
+    command_succeeds: bool,
+    assert: F,
+) {
+    let mut args = args
+        .iter()
+        .map(ToString::to_string)
+        .collect::<Vec<String>>();
+    args.insert(0, String::from("new"));
+
+    if add_path_arg {
+        let driver_name = "test-driver";
+        let tmp_dir = TempDir::new().expect("Unable to create new temp dir for test");
+        println!("Temp dir: {}", tmp_dir.path().display());
+        let driver_path = tmp_dir.join(driver_name);
+        args.push(driver_path.to_string_lossy().to_string());
+    }
+
+    let mut cmd = Command::cargo_bin("cargo-wdk").expect("unable to find cargo-wdk binary");
+    cmd.args(args);
+
+    let cmd_assertion = cmd.assert();
+    let cmd_assertion = if command_succeeds {
+        cmd_assertion.success()
+    } else {
+        cmd_assertion.failure()
+    };
+    let output = cmd_assertion.get_output();
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    println!("stdout: {stdout}");
+    let stderr = String::from_utf8_lossy(&output.stderr);
+    println!("stderr: {stderr}");
+
+    assert(&stdout, &stderr);
 }
--- a/.github/workflows/local-development-makefile.yaml
+++ b/.github/workflows/local-development-makefile.yaml
@@ -87,4 +87,10 @@ jobs:
           tool: cargo-make
 
       - name: Test wdk-pre-commit-flow cargo-make task
+        env:
+          # cargo-wdk new tests try to build newly generated
+          # driver projects which can fail in release-plz PRs
+          # because the dependencies are not yet published.
+          # This env var skips such builds for release-plz PRs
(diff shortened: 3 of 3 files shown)


crates/cargo-wdk/tests/test_utils/mod.rs
```diff
@@ -3,31 +3,11 @@
...
-        println!("RUSTFLAGS set, adding the +crt-static: {rustflags:?}");
-    } else {
-        set_var("RUSTFLAGS", "-C target-feature=+crt-static");
-        println!(
-            "No RUSTFLAGS set, setting it to: {:?}",
-            std::env::var("RUSTFLAGS").expect("RUSTFLAGS not set")
-        );
-    }
-}
-
```
gurry, line 40: This function was removed because we now set the `crt-static` flag in the `create_cargo_wdk_cmd` function below

crates/cargo-wdk/tests/test_utils/mod.rs
```diff
@@ -180,3 +160,101 @@ where
...
+/// the parent might set cargo-related environment
+/// variables that might affect with the child.
+///
+/// This function wipes the slate clean and ensures
+/// the child runs in a clean environment.
+///
+/// In particular, this function removes:
+/// - All env vars starting with "CARGO" or "RUST"
+/// - Entries added to the "PATH" variable by cargo
+fn sanitize_env_vars(cmd: &mut Command) {
```
gurry, line 287: This is the function that is reponsible for ensuring the Rust/Cargo env variables from the parent process do not get inherited by cargo-wdk invocations

crates/cargo-wdk/tests/test_utils/mod.rs
```diff
@@ -180,3 +160,101 @@ where
...
+    });
+
+    for var in vars_to_remove {
+        cmd.env_remove(var);
+    }
+
+    // Remove paths in the PATH variable that were
+    // added by cargo
+    const PATH_VAR: &str = "PATH";
+    let path_value = env::var(PATH_VAR).expect(&format!("{PATH_VAR} env var not found"));
```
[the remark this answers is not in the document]
gurry: Fixed

crates/cargo-wdk/tests/test_utils/mod.rs
```diff
@@ -180,3 +160,101 @@ where
...
+    let paths = path_value.split(";");
```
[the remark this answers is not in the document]
gurry: Fixed

crates/cargo-wdk/tests/test_utils/mod.rs
```diff
@@ -180,3 +160,102 @@ where
...
+            // unlikely to depend on those false positives either
+            !(normalized_path.ends_with("target\\debug")
+                || normalized_path.ends_with("target\\debug\\deps")
+                || normalized_path.ends_with("target\\release")
+                || normalized_path.ends_with("target\\release\\deps")
+                || normalized_path.contains(".rustup\\toolchain"))
+        })
+        .collect::<Vec<_>>();
+
+    let new_value = paths_to_keep.join(";");
```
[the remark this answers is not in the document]
gurry: This is okay

crates/cargo-wdk/tests/test_utils/mod.rs
```diff
@@ -180,3 +160,102 @@ where
...
+    });
+
+    for var in vars_to_remove {
+        cmd.env_remove(var);
+    }
+
+    // Remove paths in the PATH variable that were
+    // added by cargo
+    let path_value = env::var(PATH_VAR).expect("PATH env var not found");
+    let paths = path_value.split(';');
```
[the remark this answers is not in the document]
gurry: Fixed

crates/cargo-wdk/tests/test_utils/mod.rs
```diff
@@ -180,3 +160,102 @@ where
...
+///
+/// * `cmd_name` - Name of the cargo-wdk command. Can be only "new" or "build"
+/// * `cmd_args` - Optional args for the command
+/// * `cwd` - Optional current working directory for the command
+pub fn create_cargo_wdk_cmd<P: AsRef<Path>>(
+    cmd_name: &str,
+    cmd_args: Option<&[&str]>,
+    cwd: Option<P>,
+) -> Command {
+    assert!(cmd_name == "build" || cmd_name == "new"); // The only commands supported right now
```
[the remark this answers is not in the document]
gurry: Fixed

crates/cargo-wdk/tests/new_command_test.rs
```diff
@@ -2,10 +2,10 @@
 mod test_utils;
 use std::path::PathBuf;
 
-use assert_cmd::Command;
+use assert_cmd::{Command, assert::OutputAssertExt};
```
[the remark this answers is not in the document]
gurry: It is being used

crates/cargo-wdk/tests/test_utils/mod.rs
```diff
@@ -180,3 +160,103 @@ where
...
+        cmd.env_remove(var);
+    }
+
+    // Remove paths in the PATH variable that were
+    // added by cargo
+    let path_value = env::var(PATH_VAR).expect("PATH env var not found");
+    let paths = env::split_paths(&path_value);
+
+    let paths_to_keep = paths.filter(|path| {
+        println!("Examining PATH entry: {}", path.display());
```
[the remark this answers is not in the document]
gurry: Fixed

crates/cargo-wdk/tests/test_utils/mod.rs
```diff
@@ -3,31 +3,11 @@
 //! subdirectory prevents Cargo from treating this as an independent integration
 //! test crate and instead lets other tests import it as a regular module.
 
-use std::{collections::HashMap, ffi::OsStr};
+use std::{collections::HashMap, env, ffi::OsStr, path::Path, process::Command, vec};
```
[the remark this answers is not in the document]
gurry: Fixed

crates/cargo-wdk/tests/test_utils/mod.rs
```diff
@@ -180,3 +160,102 @@ where
...
+        // cargo-llvm-cov, which may be used to run tests, and
+        // Rust toolchain paths
+        !(path.ends_with("target/llvm-cov-target/debug")
+            || path.ends_with("target/llvm-cov-target/debug/deps")
+            || path.ends_with("target/llvm-cov-target/release")
+            || path.ends_with("target/llvm-cov-target/release/deps")
+            || path
+                .to_string_lossy()
+                .replace('\\', "/")
+                .contains(".rustup/toolchain"))
```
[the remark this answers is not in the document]
gurry: Nonsense. Short circuiting prevents this.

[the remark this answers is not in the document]
gurry: It short circuits.

crates/cargo-wdk/tests/test_utils/mod.rs
```diff
@@ -180,3 +160,102 @@ where
...
+        if var_upper.starts_with("CARGO") || var_upper.starts_with("RUST") {
+            Some(var)
+        } else {
+            None
+        }
+    });
+
+    for var in vars_to_remove {
+        cmd.env_remove(var);
+    }
```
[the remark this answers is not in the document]
gurry: This shouldn't be a problem because `remove_vars` doesn't actually remove any actual env vars because the process hasn't started yet. It merely removes entries from a `BTreeMap` inside `std::process::Command`.

.github/workflows/test.yaml
```diff
@@ -72,10 +72,17 @@ jobs:
...
 
-      - name: Install Rust Toolchain (${{ matrix.rust_toolchain }})
+      - name: Install Rust Toolchain (${{ matrix.rust_toolchain }}) and LLVM tools
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.rust_toolchain }}
+          components: llvm-tools-preview
+
+      - name: Install cargo-llvm-cov
+        if: matrix.runner.name == 'windows-2025'
```
wmmc88: maybe all of these `if: matrix.runner.name == 'windows-2025'` would be better as `if: matrix.runner.arch== 'amd64'` since the intention is tied to arch

.github/workflows/test.yaml
```diff
@@ -87,7 +94,23 @@ jobs:
...
+      - name: Upload Coverage to Codecov
+        if: matrix.runner.name == 'windows-2025'
+        uses: codecov/codecov-action@v5
+        with:
+          files: target/codecov.info
+          token: ${{ secrets.CODECOV_TOKEN }}
+          fail_ci_if_error: true
+
       - name: Run Cargo Test (workspace)
+        if: matrix.runner.name == 'windows-11-arm'
```
wmmc88: maybe do `if: matrix.runner.arch != 'amd64'`

.github/workflows/test.yaml
```diff
@@ -87,7 +94,23 @@ jobs:
...
+      - name: Run Cargo Test with Coverage (workspace)
+        if: matrix.runner.name == 'windows-2025'
+        run: cargo +${{ matrix.rust_toolchain }} llvm-cov --workspace --codecov --output-path target/codecov.info --locked --profile ${{ matrix.cargo_profile }} --all-features
+
+      - name: Upload Coverage to Codecov
+        if: matrix.runner.name == 'windows-2025'
+        uses: codecov/codecov-action@v5
+        with:
+          files: target/codecov.info
+          token: ${{ secrets.CODECOV_TOKEN }}
```
wmmc88, line 116: Does this actually work for our fork based workflow? 

from https://github.com/codecov/codecov-action:
Image

wmmc88 (commented):
I want to make sure we're aligned on the goal here. I think what we want is:
* see overall coverage from tests
* see diff in coverage from tests in PR
* block pr if coverage below some value
* be able to run code coverage evaluation locally

Does this accomplish all of this, or does the tool used in this pr have the ability to accomplish that?

wmmc88:
> I want to make sure we're aligned on the goal here. I think what we want is:
> 
> * see overall coverage from tests
> * see diff in coverage from tests in PR
> * block pr if coverage below some value
> * be able to run code coverage evaluation locally
> 
> Does this accomplish all of this, or does the tool used in this pr have the ability to accomplish that?

Other questions I have:
* Can we run the action/tool multiple times and combine the results? or at least see the results from all the runs?
    1. to see unit test vs integration test coverage
    2. to be able to run multiple times with different wdk configurations active and combine results (or at least be able to see all the results)
* Is the line coverage, branch coverage, or Modified Condition/Decision coverage?

.github/workflows/test.yaml
gurry: I was thinking that but went with `name` for consistency with what we had.

.github/workflows/test.yaml
gurry: You can obtain a token with a personal Codecov account and use that in your fork. Codecov will then record coverage separately for the fork. Here's an example of where it recorded coverage for my fork of WDR: https://app.codecov.io/gh/gurry/windows-drivers-rs/tree/add-code-coverage

gurry:
> > I want to make sure we're aligned on the goal here. I think what we want is:
> > 
> > * see overall coverage from tests
> > * see diff in coverage from tests in PR
> > * block pr if coverage below some value
> > * be able to run code coverage evaluation locally
> > 
> > Does this accomplish all of this, or does the tool used in this pr have the ability to accomplish that?
> 
> Other questions I have:
> 
> * Can we run the action/tool multiple times and combine the results? or at least see the results from all the runs?
>   
>   1. to see unit test vs integration test coverage
>   2. to be able to run multiple times with different wdk configurations active and combine results (or at least be able to see all the results)
> * Is the line coverage, branch coverage, or Modified Condition/Decision coverage?

Code coverage implemented here relies on two tools: [`cargo-llvm-cov`](https://crates.io/crates/cargo-llvm-cov) and [Codecov](https://about.codecov.io/). `cargo-llvm-cov` is a drop-in replacement for `cargo test` that internally invokes `cargo test` to run the tests and then emits coverage. Codecov is an online visualization service. The typical workflow is to generate a coverage report with `cargo-llvm-cov` and upload it to Codecov which is what we're doing in `test.yaml`.

Given that background, here are the answers to your questions:

>  * see overall coverage from tests

The report will contain coverage for whatever tests we run with `cargo llvm-cov`. Since we running all the workspace tests you will see coverage for all of them in Codecov.

>  * see diff in coverage from tests in PR

Codecov can show [percentage change in coverage ](https://docs.codecov.com/docs/coverage-percentages) on the PR page itself and allows [comparing coverage across commits](https://docs.codecov.com/docs/comparing-commits). It also shows a [graph of coverage changes over time](https://docs.codecov.com/docs/graphs).

> * block pr if coverage below some value

Yes, [that is possible](https://docs.codecov.com/docs/commit-status).

> * be able to run code coverage evaluation locally

You can run `cargo llvm-cov` locally and it will print coverage on stdout

> * Can we run the action/tool multiple times and combine the results? or at least see the results from all the runs?
>   
>   1. to see unit test vs integration test coverage
>   2. to be able to run multiple times with different wdk configurations active and combine results (or at least be able to see all the results)

Codecov [automatically merges](https://docs.codecov.com/docs/merging-reports) uploads from multiple runs into a single report. But I am not sure if it can show separate coverage for individual test groups like unit and integration tests.

> * Is the line coverage, branch coverage, or Modified Condition/Decision coverage?

`cargo-llvm-cov` supports line and region coverage. Branch coverage is supported but [is unstable and requires nightly](https://github.com/taiki-e/cargo-llvm-cov/issues/8). We will enable it when it becomes stable.

.github/workflows/test.yaml
gurry: Done

.github/workflows/test.yaml
gurry: Done, but I'm using `arm64` everywhere in `test.yaml` instead of `amd64` as that made more sense to me.

.github/workflows/test.yaml
```diff
@@ -87,7 +94,23 @@ jobs:
...
+      - name: Run Cargo Test with Coverage (workspace)
+        if: matrix.runner.arch != 'arm64'
+        run: cargo +${{ matrix.rust_toolchain }} llvm-cov --workspace --codecov --output-path target/codecov.info --locked --profile ${{ matrix.cargo_profile }} --all-features
+
+      - name: Upload Coverage to Codecov
+        if: matrix.runner.arch != 'arm64'
+        uses: codecov/codecov-action@v5
+        with:
+          files: target/codecov.info
+          token: ${{ secrets.CODECOV_TOKEN }}
```
krishnakumar4a4, line 116: What would be the behaviour of this action be in case this token is not available? Would it skip the upload with a warning (or) fail the workflow?

IMO, Failing the workflow would be odd for the forked repos, as this step is not critical for them. It would be better to treat it as a warning if that's not currently the behaviour.

crates/cargo-wdk/tests/test_utils/mod.rs
```diff
@@ -180,3 +160,102 @@ where
...
+    cmd_name: &str,
+    cmd_args: Option<&[&str]>,
+    curr_working_dir: Option<P>,
+) -> Command {
+    assert!(
+        cmd_name == "build" || cmd_name == "new",
+        "Only 'build' and 'new' commands are supported"
+    );
+
+    let mut cmd = Command::cargo_bin("cargo-wdk").expect("unable to find cargo-wdk binary");
```
krishnakumar4a4, line 251: nit: Not related to your change, but a suggestion if you would like to incorporate

```suggestion
    let mut cmd =
        Command::cargo_bin(env!("CARGO_PKG_NAME")).expect("unable to find cargo-wdk binary");
```

or 

```suggestion
    let mut cmd = Command::cargo_bin(env!("CARGO_PKG_NAME"))
        .expect(&format!("unable to find {} binary", env!("CARGO_PKG_NAME")));
```

-------

FYI, There is even a better way, which is to call [assert_cmd::cargo::cargo_bin_cmd!()](https://docs.rs/assert_cmd/latest/assert_cmd/cargo/macro.cargo_bin_cmd.html) but that requires latest asssert_cmd 2.1.1. Something we could do only after a bump up.

.github/workflows/test.yaml
wmmc88: May want to mention needing to do this in `contributing.md`

crates/cargo-wdk/tests/test_utils/mod.rs
gurry: Let us do it after upgrading to 2.1.1

.github/workflows/test.yaml
gurry: That's a good point. I have never relied on workflows running in my own fork. If I need to get some tests run I open a draft PR on the upstream repo instead. That said, others may have different preferences so let me try to address this.

gurry: I have changed the workflow to not attempt upload to codecov in forks

.github/workflows/test.yaml
gurry: I have changed the workflow to not attempt upload to codecov in forks. So this should not be a problem any more.

.github/workflows/test.yaml
krishnakumar4a4: How about skipping upload when secrets.CODECOV_TOKEN token is not available, instead of doing it based on the repository? This way, it may still work on forked repos if someone has set their personal token.

gurry: I'd thought of that but it'll mean we'll publish to their codecov account without their knowledge. So perhaps not a good idea.

crates/cargo-wdk/tests/test_utils/mod.rs
```diff
@@ -5,39 +5,22 @@
 
 use std::{
     collections::HashMap,
+    env,
     ffi::{CStr, CString, OsStr},
     marker::PhantomData,
+    path::Path,
+    process::Command,
+    vec,
```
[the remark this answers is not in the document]
gurry: Removed

krishnakumar4a4 (dismissed):

wmmc88 (approved):

PR #561, opened 2025-11-05
chore: reduce dependency dupes in lockfiles

### LINKED ISSUE #500: Fix duplicate windows-sys versions and enforce in CI
This PR also seems to introduce 2 different versions of `windows-sys` into the build. This is not ideal for build times etc, but we can push that fix to a followup to unblock pipelines rn. It should just involving determining what's bringing in the duplicate versions and trying to uses `cargo update` to target an update to one. It might involve submitting a pr to loosen windows-sys version in a dependency, but i doubt that would be needed in this case since there are no new deps introduced in this pr.

Not sure why clippy isn't catching the duplicate though since we should already have https://rust-lang.github.io/rust-clippy/master/index.html#multiple_crate_versions enabled

_Originally posted by @wmmc88 in https://github.com/microsoft/windows-drivers-rs/pull/492#pullrequestreview-3197762002_

wmmc88:
Removes duplicate dependency versions in workspace, examples, and tests. This should slightly speed up build, test, pipelines, etc

closes #500

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -47,7 +47,7 @@ version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6c8bdeb6047d8983be085bab0ba1472e6dc604e7041dbf6fcd5e71523014fae9"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys",
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


krishnakumar4a4 (approved):
lgtm. 

Things I have focussed on this review are 
- windows-sys 0.52.0 is removed
- nu-ansi-term bumped to 0.50.3 in which windows-sys dep range is widened https://github.com/nushell/nu-ansi-term/pull/69

I believe rest of the changes are automatic updates done due to cargo update run.

gurry (approved):

PR #574, opened 2025-11-12
fix: path construction in `wdk-build` `get_wdk_tools_root` and `get_wdk_bin_root`

svasista-ms:
A bug in `PATH` system env variable setup was identified in the latest release. This bug affects eWDK scenarios for BOTH `cargo-make` and `cargo-wdk` scenarios. 

image

The above image shows the difference in paths that are setup by `wdk-build v0.4.0` vs `wdk-build v0.5.0`. 

This PR fixes the issue by appending the `Version String` only in the default case, i.e, when `WDKToolRoot` / `WDKBinRoot` are not set.[[1]](https://github.com/microsoft/windows-drivers-rs/compare/main...svasista-ms:windows-drivers-rs:fix-wdktoolroot?expand=1#diff-da58729eb32b74ae10c700766228300e3637a2f1141f291b8cd128497071bde3R599-R603)
In case of eWDK env, the values include the version string and need not be appended. Since Nuget WDK environment also depend on the same variables, in the `install-wdk` action, the `WDKToolRoot` and `WDKBinRoot` variables are now set with the `sdkVersion` component. [[2]](https://github.com/microsoft/windows-drivers-rs/compare/main...svasista-ms:windows-drivers-rs:fix-wdktoolroot?expand=1#diff-d9007f75de09530fc55ec3e8651e7f45ddf9d1db50f1b342923eb03444e1da39R221-R222)
The PR also includes a `clippy` suggestion to borrow `sdk_version` (`&str`) parameter in these functions instead of owned `String`.

### DIFF
--- a/crates/wdk-build/src/cargo_make.rs
+++ b/crates/wdk-build/src/cargo_make.rs
@@ -13,7 +13,7 @@
 
 use core::{fmt, ops::RangeFrom};
 use std::{
-    env,
+    env::{self, VarError},
     panic::UnwindSafe,
     path::{Path, PathBuf, absolute},
     process::Command,
@@ -547,7 +547,7 @@ pub fn setup_path() -> Result<impl IntoIterator<Item = String>, ConfigError> {
     let host_arch = CpuArchitecture::try_from_cargo_str(env::consts::ARCH)
         .expect("The rust standard library should always set env::consts::ARCH");
 
-    let wdk_bin_root = get_wdk_bin_root(&wdk_content_root, &sdk_version);
+    let wdk_bin_root = get_wdk_bin_root(&wdk_content_root, &sdk_version)?;
 
     let host_windows_sdk_ver_bin_path = {
         let path = wdk_bin_root.join(host_arch.as_windows_str());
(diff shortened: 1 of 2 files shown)


crates/wdk-build/src/cargo_make.rs
```diff
@@ -596,16 +596,18 @@ pub fn setup_path() -> Result<impl IntoIterator<Item = String>, ConfigError> {
     Ok([PATH_ENV_VAR].map(ToString::to_string))
 }
 
-fn get_wdk_tools_root(wdk_content_root: &Path, sdk_version: String) -> PathBuf {
-    env::var("WDKToolRoot")
-        .map_or_else(|_| wdk_content_root.join("tools"), PathBuf::from)
-        .join(sdk_version)
+fn get_wdk_tools_root(wdk_content_root: &Path, sdk_version: &str) -> PathBuf {
+    env::var("WDKToolRoot").map_or_else(
+        |_| wdk_content_root.join("tools").join(sdk_version),
```
krishnakumar4a4: Perhaps good to perform `join()` only for `VarError::NotPresent` case. For the other case, it may be better to simply throw error in setup_path.

crates/wdk-build/src/cargo_make.rs
```diff
@@ -596,16 +596,18 @@ pub fn setup_path() -> Result<impl IntoIterator<Item = String>, ConfigError> {
     Ok([PATH_ENV_VAR].map(ToString::to_string))
 }
 
-fn get_wdk_tools_root(wdk_content_root: &Path, sdk_version: String) -> PathBuf {
-    env::var("WDKToolRoot")
-        .map_or_else(|_| wdk_content_root.join("tools"), PathBuf::from)
-        .join(sdk_version)
+fn get_wdk_tools_root(wdk_content_root: &Path, sdk_version: &str) -> PathBuf {
+    env::var("WDKToolRoot").map_or_else(
```
krishnakumar4a4: Add a comment here explaining why the version needs to be joined.

wmmc88 (approved):

PR #579, opened 2025-11-13
fix: fallback to detection logic when WDK env vars have invalid unicode

gurry:
This PR removes the new error variant introduced in #574 and thus avoids breaking compat.

Instead of returning error we now merely trace the failure and fall back to returning the default value.

### DIFF
--- a/crates/wdk-build/src/cargo_make.rs
+++ b/crates/wdk-build/src/cargo_make.rs
@@ -13,7 +13,7 @@
 
 use core::{fmt, ops::RangeFrom};
 use std::{
-    env::{self, VarError},
+    env,
     panic::UnwindSafe,
     path::{Path, PathBuf, absolute},
     process::Command,
@@ -547,7 +547,7 @@ pub fn setup_path() -> Result<impl IntoIterator<Item = String>, ConfigError> {
     let host_arch = CpuArchitecture::try_from_cargo_str(env::consts::ARCH)
         .expect("The rust standard library should always set env::consts::ARCH");
 
-    let wdk_bin_root = get_wdk_bin_root(&wdk_content_root, &sdk_version)?;
+    let wdk_bin_root = get_wdk_bin_root(&wdk_content_root, &sdk_version);
 
     let host_windows_sdk_ver_bin_path = {
         let path = wdk_bin_root.join(host_arch.as_windows_str());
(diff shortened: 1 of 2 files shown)


crates/wdk-build/src/cargo_make.rs
```diff
@@ -596,24 +596,36 @@ pub fn setup_path() -> Result<impl IntoIterator<Item = String>, ConfigError> {
...
-    }
+fn get_wdk_bin_root(wdk_content_root: &Path, sdk_version: &str) -> PathBuf {
+    get_path_from_env("WDKBinRoot", wdk_content_root, "bin", sdk_version)
+}
+
+/// Reads path from the given env variable or falls back to
+/// constructing it from WDK content root.
+///
+/// The path in `env_var` is already the full path because eWDK
+/// and Nuget CLI set it that way. In the fallback however we
```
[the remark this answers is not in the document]
gurry: Not important

svasista-ms (approved):

gurry:
As per @svasista-ms's testing everything's working correctly. @wmmc88 please go ahead and merge it if you approve. 

The tests PR https://github.com/microsoft/windows-drivers-rs/pull/578 should go on top of this.

crates/wdk-build/src/cargo_make.rs
```diff
@@ -596,24 +596,36 @@ pub fn setup_path() -> Result<impl IntoIterator<Item = String>, ConfigError> {
...
+/// have to append `sub_folder` and `sdk_version` manually.
+fn get_path_from_env(
+    env_var: &str,
+    wdk_content_root: &Path,
+    sub_folder: &str,
+    sdk_version: &str,
+) -> PathBuf {
+    env::var(env_var).map_or_else(
+        |e| {
+            trace!(
```
wmmc88, line 621: Not going to block merge on this, so that we can cut the patch release, but I have some thoughts here:

1. I think you should distinguish between the "env var not present" case and the "env var has invalid unicode" case
2. in the former, it should be `trace`. in the latter it should be `warn`

wmmc88: https://github.com/microsoft/windows-drivers-rs/issues/580

gurry: Agreed

Issue #582, opened 2025-11-14
Run `cargo wdk new` tests in CI


wmmc88:
> We need to figure out a way to run these ignored tests, but ignore them for release prs 

 _Originally posted by @wmmc88 in [#577](https://github.com/microsoft/windows-drivers-rs/pull/577/files#r2525705442)_

There is a desire to only update the templates when doing release prs, but this breaks some cargo wdk new tests since the crates aren't released yet. The easiest solution i see here is to run these `#[ignore]`'ed tests only in **PRs** that don't come from a `release-plz-` branch

PR #583, opened 2025-11-14
chore: re-enable temporarily disabled `cargo-wdk new` tests

gurry:
They were disabled in order to make the release PR checks pass. 

We need to find a more permananent solution to the problem of upgrading dep versions in `cargo-wdk new` templates while releasing those dep crates at the same time.


wmmc88 (approved):
Would it be possible to implement #582 instead? (or as a followup soon)

gurry:
> Would it be possible to implement #582 instead? (or as a followup soon)

Yeah, I'll look into it

Issue #580, opened 2025-11-13
Warn on invalid values for WDK Env Vars


wmmc88:
Not going to block merge on this, so that we can cut the patch release, but I have some thoughts here:

1. I think you should distinguish between the "env var not present" case and the "env var has invalid unicode" case
2. in the former, it should be `trace`. in the latter it should be `warn`

_Originally posted by @wmmc88 in https://github.com/microsoft/windows-drivers-rs/pull/579#discussion_r2524357482_

wmmc88:
after #579, invalid WDK env var values will silently fall back to the detection logic. The log is only a trace. In the case that the env var has an invalid value, I think it should come up by default (ie. `warn`) so that the user isn't confused why their env var isn't working

PR #596, opened 2025-12-23
fix: warn when WDK env vars are non-UTF8

gurry:
Fixes https://github.com/microsoft/windows-drivers-rs/issues/580

### DIFF
--- a/crates/wdk-build/src/cargo_make.rs
+++ b/crates/wdk-build/src/cargo_make.rs
@@ -621,7 +621,7 @@ fn get_path_from_env(
             const FALLBACK_MSG: &str = "Constructing path from WDK content root instead";
             match e {
                 env::VarError::NotPresent => {
-                    trace!("Env var '{env_var}' not found. {FALLBACK_MSG}")
+                    trace!("Env var '{env_var}' not found. {FALLBACK_MSG}");
                 }
                 env::VarError::NotUnicode(val) => {
                     warn!(


crates/wdk-build/src/cargo_make.rs
```diff
@@ -618,10 +618,15 @@ fn get_path_from_env(
 ) -> PathBuf {
     env::var(env_var).map_or_else(
         |e| {
-            trace!(
-                "Could not read env var '{env_var}': {e:?}. Constructing path from WDK content \
-                 root"
-            );
+            const COMMON_MSG: &str = "Constructing path from WDK content root instead";
```
wmmc88: nit: more descriptive name for this constant

wmmc88 (dismissed):
LGTM, just one minor nit. Merge is going to have to wait for #597 and probably #598 to go in

crates/wdk-build/src/cargo_make.rs
gurry: I can call it `COMMON_ERR_MSG` because that's what it really is. What do you think?

gurry: Renamed it to `FALLBACK_MSG`

wmmc88 (approved):

crates/wdk-build/src/cargo_make.rs
```diff
@@ -618,10 +618,17 @@ fn get_path_from_env(
...
-                 root"
-            );
+            const FALLBACK_MSG: &str = "Constructing path from WDK content root instead";
+            match e {
+                env::VarError::NotPresent => {
+                    trace!("Env var '{env_var}' not found. {FALLBACK_MSG}")
+                }
+                env::VarError::NotUnicode(val) => {
+                    warn!(
+                        "Env var '{env_var}' contains non-UTF8 characters: {val:?}. {FALLBACK_MSG}"
```
krishnakumar4a4, line 628: I think the message should say `non-unicode` instead of `non-UTF8`.

gurry: This error occurs only when the env var value is non-UTF8 because the function `var` needs to return a `String` which is always UTF8. So this message is accurate.

krishnakumar4a4: Okay, what we need as a return value of this `get_path_from_env` function is a UTF8 string. 

So, the message could either mention that the environment variable contains a non-unicode value (not UTF8 or UTF16), or it could explain that because it has non-unicode content, it can't be a valid UTF8 string, which is required as our return value. From your explanation, it looks like we're choosing the second approach.

gurry: `get_path_from_env` returns a `PathBuf` not a string and `PathBuf` works with non-UTF8 strings too. So `get_path_from_env` itself is not a problem. The problem is `env::var`. It needs to return a `String` which is always UTF8.

krishnakumar4a4: Sorry, I overlooked the return value of the `get_path_from_env` function. What I should have said is that it’s the return value of `env::var` itself. But anyways, I think I got the answer I am looking for.

gurry: To be fair, we can construct a `PathBuf` from a non-UTF8 and even a non-unicode string and return that instead of falling back, but eventually that `PathBuf` would be converted to a `String` somewhere and then that call to `to_string_lossy` will fail. So better to fallback right here and ensure the `PathBuf` is always UTF8.

krishnakumar4a4 (approved):

PR #602, opened 2025-12-29
fix: replace direct call to `std::fs::create_dir_all` with `fs` provider `fs::create_dir_all`

svasista-ms:
Calling `std::fs::create_dir_all` from the `NewAction::copy_cargo_config` method led to bypassing the `Fs` provider mock during test execution. This led to creation of temporary directories:

image

In this PR, a `create_dir_all` method is added to `Fs` provider and all direct calls to `std::fs::create_dir_all` have been replaced by this method. This allows the mocks to work as expected during test execution.


crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -830,6 +828,13 @@ mod tests {
         }
 
         fn expect_copy_cargo_config(mut self, is_copy_success: bool) -> Self {
+            let cargo_dir = self.path.join(".cargo");
+            let expected_cargo_dir = cargo_dir;
+            self.mock_fs
+                .expect_create_dir_all()
+                .withf(move |path| path == expected_cargo_dir)
```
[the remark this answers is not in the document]
gurry: Yeah, I don't think we need that redundant variable `expected_cargo_dir`. We should accept Copilot's suggestion

crates/cargo-wdk/src/providers/fs.rs
```diff
@@ -38,6 +48,10 @@ impl Fs {
         create_dir(path).map_err(|e| FileError::CreateDirError(path.to_owned(), e))
     }
 
+    pub fn create_dir_all(&self, path: &Path) -> Result<(), FileError> {
+        create_dir_all(path).map_err(|e| FileError::CreateDirError(path.to_owned(), e))
+    }
```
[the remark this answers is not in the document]
gurry: I think we can ignore this one

crates/cargo-wdk/src/providers/fs.rs
```diff
@@ -82,3 +96,30 @@ impl Fs {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use assert_fs::{TempDir, prelude::*};
+
+    use super::Fs;
+    use crate::providers::error::FileError;
+
```
gurry: Need a test for the success case as well i.e. when the dir is created

crates/cargo-wdk/src/providers/fs.rs
```diff
@@ -82,3 +96,30 @@ impl Fs {
...
+    #[test]
+    fn create_dir_all_returns_create_dir_error_when_target_is_not_a_dir() {
```
gurry:
```suggestion
    fn create_dir_all_returns_error_when_target_is_not_a_dir() {
```

Nit: no need to mention the type of the error in the name. Makes it a bit shorter

crates/cargo-wdk/src/actions/new/mod.rs
```diff
@@ -830,6 +828,12 @@ mod tests {
         }
 
         fn expect_copy_cargo_config(mut self, is_copy_success: bool) -> Self {
+            let cargo_dir = self.path.join(".cargo");
+            self.mock_fs
+                .expect_create_dir_all()
+                .withf(move |path| path == cargo_dir)
```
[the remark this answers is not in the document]
gurry: Ignore it

crates/cargo-wdk/src/providers/fs.rs
```diff
@@ -82,3 +96,46 @@ impl Fs {
...
+        let temp = TempDir::new().expect("failed to create temp dir");
+        let nested_dir = temp.child("a").child("b");
+
+        nested_dir.assert(predicate::path::missing());
+
+        let fs = Fs::default();
+        fs.create_dir_all(nested_dir.path())
+            .expect("expected create_dir_all to succeed");
+
+        temp.child("a").assert(predicate::path::is_dir());
```
krishnakumar4a4: nit: This check looks redundant. `nested_dir` check below should be enough.

svasista-ms: As per our discussion, I removed the tests for `Fs::create_dir_all` as it is a thin wrapper method. Marking the conversation "Resolved"

crates/cargo-wdk/src/providers/fs.rs
svasista-ms: As per our discussion, the tests for `Fs::create_dir_all` were removed as they are thin wrappers over `std` lib calls. Marking the conversation "Resolved"

crates/cargo-wdk/src/providers/fs.rs
svasista-ms: Tests for `Fs::create_dir_all` were removed as they are thin wrappers over std lib calls. Marking the conversation "Resolved"

PR #604, opened 2026-01-28
fix(ci): use latest nightly toolchain in the Test workflow, fix `assert_matches` import and the `clippy` warning

svasista-ms:
The Lint workflow is failing with:
1. `error[E0432]: unresolved import std::assert_matches`[[1]](https://github.com/microsoft/windows-drivers-rs/actions/runs/21394819155/job/61590415286)
2. `clippy::map_unwrap_or` (pedantic) warning promoted to failure in the job configuration. [[2]](https://github.com/microsoft/windows-drivers-rs/actions/runs/21284030225/job/61260678184#step:7:277)

I updated my local dev machine to match the toolchain versions used by the pipeline and reproduced the same failure locally.

Also, the issue in `macrotest`[[3]](https://github.com/eupn/macrotest/issues/131) due to which the Test flow's nightly toolchain version was pinned (to nightly-2025-11-20) is now resolved. 

In this PR: 
1. Test workflow is updated to use the latest nightly toolchain
2. The `assert_matches` import is fixed.
3. `clippy` pedantic warning in `wdk-build` is fixed by replacing `env::var(...).map(...).unwrap_or(false)` with `Result::is_ok_and(...)` in `is_cargo_make_color_disabled` fn.

### DIFF
--- a/crates/wdk-build/src/lib.rs
+++ b/crates/wdk-build/src/lib.rs
@@ -1548,7 +1548,7 @@ pub fn detect_wdk_build_number() -> Result<u32, ConfigError> {
 #[cfg(test)]
 mod tests {
     #[cfg(nightly_toolchain)]
-    use std::assert_matches::assert_matches;
+    use std::assert_matches;
     use std::{collections::HashMap, ffi::OsStr, sync::Mutex};
 
     use super::*;


crates/wdk-build/src/lib.rs
```diff
@@ -1547,8 +1547,6 @@ pub fn detect_wdk_build_number() -> Result<u32, ConfigError> {
 
 #[cfg(test)]
 mod tests {
-    #[cfg(nightly_toolchain)]
```
wmmc88, line 1550: revert these changes. unrelated to pr. Also, assert_matches provides better diagnostics and is about to be stabilized: https://github.com/rust-lang/rust/pull/137487

wmmc88 (approved):

PR #610, opened 2026-02-10
feat: add dbg! macro to wdk

Alan632:
This PR adds a no_std implementation of the `dbg!` macro to wdk.

Refer to the macro [dbg!](https://doc.rust-lang.org/std/macro.dbg.html) for the std crate's `dbg!` details.


Alan632:
@microsoft-github-policy-service agree company="Microsoft"

crates/wdk/src/dbg.rs
gurry, line 1: I wonder if it would be better to put this in `print.rs` instead of a separate mod/source file

Alan632: I can get behind this since `dbg!` is closely related to print and `dbg.rs` only contained the single macro anyway.

Macro has been moved to `print.rs`.

Alan632:
Quick verification test using the sample kmdf example in `windows-drivers-rs`.

behavior in DebugView:
image_1770864083904_0

`std` environment behavior:
image_1770864158715_0

Alan632:
[Issue-616](https://github.com/microsoft/windows-drivers-rs/issues/616) created to track the TODO in `dbg!`.

PR #611, opened 2026-02-13
feat!: stack-based format buffer for `wdk` crate 

leon-xd:
Adds and integrates `wdk::fmt::FormatBuffer`.

## Summary
- Introduced `crates/wdk/src/fmt.rs` with a public `wdk::fmt` module containing two types:
  - `FormatBuffer<N>` — a fixed-size, stack-allocated formatting buffer implementing `fmt::Write`
  - `FlushableFormatBuffer<F, N>` — a wrapper that auto-flushes via closure on overflow and on drop
- Swapped `print.rs` WDM/KMDF path to use `FlushableFormatBuffer` instead of `DbgPrintBufWriter`; removed that module and its tests.
- Turns the `alloc` feature from the `wdk` crate into a no-op — it had no remaining uses after the stack-buffer migration. Avoids removal of feature to prevent a breaking change
- Bumps MSRV from 1.85 to 1.91 due to use of [`str::char_floor_boundary`](https://doc.rust-lang.org/core/primitive.str.html#method.floor_char_boundary)

## Details
- `FormatBuffer<N>` stores a zero-initialized `[u8; N]` and tracks `used` bytes. The last byte is reserved for a NUL terminator, so usable capacity is `N - 1`. `N` must be at least 2; smaller values will not compile (`const { assert!(N >= 2) }`).
- All buffer mutations go through a single `append_bytes` helper that copies data and sets `buffer[used] = 0`, centralizing the NUL terminator invariant.
- `as_str()` returns an infallible UTF-8 view — only valid UTF-8 enters via `write_str` (both `FormatBuffer::write_str` and `FlushableFormatBuffer::write_str` copy from `&str` sources).
- `as_c_str()` returns a `&CStr` view up to the first NUL. Infallible in practice since the NUL invariant is always maintained.
- `fmt::Write` clamps overflow writes at a UTF-8 char boundary via `floor_char_boundary` and signals truncation via `fmt::Error`.
- `FlushableFormatBuffer<F, N>` wraps `FormatBuffer` and auto-flushes via a closure when the buffer fills. Remaining content is flushed on drop. This allows arbitrarily long formatted output to be processed in fixed-size chunks.
- `print.rs` WDM/KMDF path now formats into `FlushableFormatBuffer<_, 512>` and calls `DbgPrint` via `%s` using the resulting `CStr`.
- Removed the `alloc` feature gate from the WDM/KMDF print path — `print!`/`println!` no longer require heap allocation and work at any IRQL where `DbgPrint` is valid.
- Custom `Debug` impl on `FormatBuffer` shows only the used content, not stale bytes after `clear()`.

## Testing
29 tests covering:
- basic write, `as_str()`, `as_c_str()` usage
- overflow, multi-write, and exact-fit scenarios
- empty writes and min-sized buffers (`N = 2`)
- clear-then-shorter-write regression (NUL terminator correctness)
- multi-byte UTF-8 char boundary handling in both buffer types
- explicit `flush()` mid-stream and flush-on-drop
- interior NUL truncation via `as_c_str()`
- `compile_fail` doctest for `N < 2`

```pwsh
cargo test --package wdk --lib --all-features -- fmt --nocapture
```

## Notes / Follow-ups
- `_print` ignores `fmt::write` errors — partial output is acceptable for debug printing. Errors from individual `Display` impls are silently dropped.
- The old `DbgPrintBufWriter` stripped interior NUL bytes and continued printing. The new WDM/KMDF path truncates each chunk at the first NUL (via `as_c_str`). UMDF still strips NULs and prints the remainder.


crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,334 @@
...
+///
+/// let mut buf = WdkFormatBuffer::<16>::new();
+/// write!(&mut buf, "hello {}", 42).unwrap();
+///
+/// let s = buf.as_str().unwrap();
+/// assert_eq!(s, "hello 42");
+///
+/// let c = buf.as_cstr().unwrap();
+/// assert_eq!(c.to_bytes(), b"hello 42");
+/// ```
```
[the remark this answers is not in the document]
leon-xd: the crate does expose a public `fmt` module, closing this.

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,334 @@
...
+        //     unreachable!("Buffer should always contain a trailing null
+        // byte"); }
+        // cstr.unwrap_or_else(| _err | {
+        //     // SAFETY: We pass a single null byte into
+        // `from_bytes_with_nul_unchecked`. This means that it is null
+        // terminated and has no interior null bytes.     unsafe{
+        //         CStr::from_bytes_with_nul_unchecked(b"\0")
+        //     }
+        // })
+    }
```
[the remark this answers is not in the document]
leon-xd: I will remove this after getting comments from a human :)

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,334 @@
...
+    /// Ensures termination by writing a `NUL` if the buffer is completely
+    /// filled.
+    ///
+    /// # Errors
+    /// Returns an error only if no terminator is found, e.g. if `T == 0`.
+    pub const fn as_cstr(&mut self) -> Result<&CStr, FromBytesUntilNulError> {
+        if self.used == T && T != 0 {
+            self.buffer[self.used - 1] = 0;
+        }
+        CStr::from_bytes_until_nul(&self.buffer)
```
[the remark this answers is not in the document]
leon-xd: yeah fair, I'll wait for human input for this one. want to balance the expectations of having Rust-only strings, but given the developer experience required to build in WDR this might be over compensating for a non-helpful use case. had a comment somewhere about this design decision.

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,334 @@
...
+impl<const T: usize> fmt::Write for WdkFormatBuffer<T> {
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        if s.len() + self.used >= T {
+            self.buffer[self.used..T].copy_from_slice(&s.as_bytes()[..T - self.used]);
+            self.used = T;
+            return Err(fmt::Error);
+        }
+        self.buffer[self.used..self.used + s.len()].copy_from_slice(s.as_bytes());
+        self.used += s.len();
+        Ok(())
```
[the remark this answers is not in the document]
leon-xd: this is reflected in my test cases as ideal behavior since converting to cstr will truncate. just need feedback on design before making any changes.

examples/sample-kmdf-driver/src/lib.rs
```diff
@@ -92,7 +92,7 @@ pub unsafe extern "system" fn driver_entry(
     };
 
     let driver_attributes = WDF_NO_OBJECT_ATTRIBUTES;
-    let driver_handle_output = WDF_NO_HANDLE.cast::<WDFDRIVER>();
+    let driver_handle_output: *mut *mut wdk_sys::WDFDRIVER__ = WDF_NO_HANDLE.cast::<WDFDRIVER>();
```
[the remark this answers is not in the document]
leon-xd: `cargo fmt` added this line, so it's staying in

crates/wdk/src/print.rs
```diff
@@ -88,14 +90,24 @@ macro_rules! println {
...
-            if fmt::write(&mut buffered_writer, args).is_ok() {
-                buffered_writer.flush();
-            } else {
-                unreachable!("DbgPrintBufWriter should never fail to write");
+            unsafe {
+                wdk_sys::ntddk::DbgPrint(
+                    c"%s".as_ptr().cast(),
+                    cstr_buffer.as_ptr().cast::<wdk_sys::CHAR>(),
+                );
             }
```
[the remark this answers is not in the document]
leon-xd: great point!

crates/wdk/src/print.rs
```diff
@@ -142,21 +146,31 @@ macro_rules! dbg {
...
+            let cstr_buffer = match buffered_writer.as_cstr() {
+                Ok(cstr) => cstr,
+                Err(_e) => return, // silently return on error, no null terminator. (Should this be a placeholder string for debugging purposes?)
+            };
 
-            if fmt::write(&mut buffered_writer, args).is_ok() {
-                buffered_writer.flush();
-            } else {
-                unreachable!("DbgPrintBufWriter should never fail to write");
+            unsafe {
```
Alan632: Does this need a safety doc/comment?

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,334 @@
...
+        // &CStr. But handling this error here caused a panic which
+        // feels too strict given the driver environment since that would cause
+        // a bugcheck for kernel drivers. Determined that the most
+        // elegant way was for the user to always handle the errors but I'm open
+        // to feedback here.
+        //
+        // // if I do implement this way how should I handle an error case here?
+        // feels too strong to panic but this case cannot happen as the current
+        // logic stands. if cstr.is_err() {
+        //     unreachable!("Buffer should always contain a trailing null
```
Alan632: I am a fan of this for its clarity.

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,334 @@
...
+///
+/// let c = buf.as_cstr().unwrap();
+/// assert_eq!(c.to_bytes(), b"hello 42");
+/// ```
+pub struct WdkFormatBuffer<const T: usize = DEFAULT_WDK_FORMAT_BUFFER_SIZE> {
+    buffer: [u8; T],
+    used: usize,
+}
+
+impl<const T: usize> WdkFormatBuffer<T> {
```
gurry: `N` is a more idiomatic name for const generics representing size instead of `T` as seen in the stdlib array type: https://doc.rust-lang.org/std/primitive.array.html#implementations

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    /// Returns a C string view up to the first `NUL` byte.
+    ///
+    /// Ensures termination by writing a `NUL` if the buffer is completely
+    /// filled.
+    ///
+    /// # Errors
+    /// Returns an error only if no terminator is found, e.g. if `T == 0`.
+    pub const fn as_cstr(&mut self) -> Result<&CStr, FromBytesUntilNulError> {
```
gurry: This function should take `&self` instead of `&mut self` because from the outside it is a "read" operation. We should add the terminator in `write_str()` instead.

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,334 @@
...
+impl<const T: usize> fmt::Write for WdkFormatBuffer<T> {
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        if s.len() + self.used >= T {
+            self.buffer[self.used..T].copy_from_slice(&s.as_bytes()[..T - self.used]);
+            self.used = T;
+            return Err(fmt::Error);
+        }
+        self.buffer[self.used..self.used + s.len()].copy_from_slice(s.as_bytes());
+        self.used += s.len();
+        Ok(())
```
gurry: A comment or comments at the right places explaining the logic will be very helpful.

examples/sample-kmdf-driver/src/lib.rs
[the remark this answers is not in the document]
wmmc88: huh. are you sure `cargo fmt` added this? this seems unlikely

crates/wdk/src/fmt.rs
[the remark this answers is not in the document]
gurry: My suggestion is to add the needed null terminator in `write_str` while writing the string instead of here. Then this function can return a `&Cstr` without mutating the buffer or worrying about errors e.g.:
```rust
pub const fn as_cstr(&mut self) -> &CStr {
      match CStr::from_bytes_until_nul(&self.buffer) {
        Ok(cstr) => cstr,
        Err(_) => unreachable!("self.buffer is guaranteed to be null-terminated"),
    }
}
```

crates/wdk/src/fmt.rs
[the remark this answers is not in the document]
leon-xd: Removed

crates/wdk/src/fmt.rs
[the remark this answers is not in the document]
leon-xd: Fixed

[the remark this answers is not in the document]
leon-xd: I kept the error because it is still able to trigger when N = 0

crates/wdk/src/fmt.rs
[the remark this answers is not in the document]
leon-xd: fixed by reserving N-1 for C strings

crates/wdk/src/print.rs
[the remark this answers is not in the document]
leon-xd: Fixed

examples/sample-kmdf-driver/src/lib.rs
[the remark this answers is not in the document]
leon-xd: ok after testing cargo fmt did not add this. Not sure what happened, maybe some nightly business? Reverted this.

crates/wdk/src/print.rs
leon-xd: Added

crates/wdk/src/print.rs
```diff
@@ -17,8 +17,10 @@ requirements of that function apply. In particular, this should only be called a
 `IRQL` <= `DIRQL`, and calling it at `IRQL` > `DIRQL` can cause deadlocks due to
 the debugger's use of IPIs (Inter-Process Interrupts).
 
-[`wdk_sys::ntddk::DbgPrint`]'s 512 byte limit does not apply to this macro, as it will
-automatically buffer and chunk the output if it exceeds that limit.
+Output is formatted into a fixed 512-byte stack buffer and passed to
+[`wdk_sys::ntddk::DbgPrint`] in a single call. If the formatted output exceeds 511 bytes
+(reserving one byte for the NUL terminator), it is silently truncated. Interior NUL bytes
+in the formatted output will cause the string to be truncated at the first NUL.
 "
```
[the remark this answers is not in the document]
leon-xd: Fixed

crates/wdk/src/fmt.rs
leon-xd: N = 0 still creates this error so we are keeping it

crates/wdk/src/fmt.rs
leon-xd: Replaced

crates/wdk/src/fmt.rs
leon-xd: Fixed

crates/wdk/src/fmt.rs
leon-xd: Added

crates/wdk/src/fmt.rs
```diff
@@ -59,52 +60,42 @@ impl<const T: usize> WdkFormatBuffer<T> {
 
     /// Returns a C string view up to the first `NUL` byte.
     ///
-    /// Ensures termination by writing a `NUL` if the buffer is completely
-    /// filled.
+    /// The buffer always contains a NUL terminator because `write_str`
+    /// reserves the last byte. This method does not modify the buffer.
```
gurry: Nit: we don't need to say "This method does not modify the buffer" because users won't expect it to now.

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,307 @@
...
+    used: usize,
+}
+
+impl<const N: usize> WdkFormatBuffer<N> {
+    /// Creates a zeroed formatting buffer with capacity `N`.
+    ///
+    /// The buffer starts empty (`used == 0`) and is ready for `fmt::Write`.
+    #[must_use]
+    pub const fn new() -> Self {
+        Self {
```
gurry, line 56:
```suggestion
        const { assert!(N > 0, "N must be at least 1"); }
        Self {
```
Adding this const block will make it impossible to construct the type with size `0`, removing the need for `N == 0` checks elsewhere

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,307 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<const N: usize> fmt::Write for WdkFormatBuffer<N> {
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        if N == 0 {
+            return Err(fmt::Error);
+        }
```
gurry: We can remove this check if we add the `assert!(N > 0,...)` in `new`

crates/wdk/src/fmt.rs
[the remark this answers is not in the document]
gurry: We can remove the error if we add the `assert!(N > 0,...)` in `new` because `N` will never be `0` then.

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,307 @@
...
+
+        // The last byte (buffer[N-1]) is reserved for the NUL terminator
+        // so that `as_cstr` can read without mutating the buffer.
```
gurry:
```suggestion
        // so that the buffer always contains a valid `CStr`
```

Nit: no one would have expected `as_cstr` to mutate the buffer to begin with so the current comment looks odd.

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,307 @@
...
+        let capacity = N - 1;
+        let remaining = capacity - self.used;
+
+        // Overflow: copy what fits into the remaining space and signal error.
+        if s.len() > remaining {
+            self.buffer[self.used..self.used + remaining]
+                .copy_from_slice(&s.as_bytes()[..remaining]);
+            self.used = capacity;
+            return Err(fmt::Error);
```
gurry, line 164: This one is a bit subjective, but I feel we should not both update the buffer and return an error. Either we accept the input as valid and mutate the buffer with it or we reject it completely with an error. Doing both may be confusing to the user.

I feel it is fine to just silently drop extra characters here as that is effectively what the contract of this type is.

wmmc88: From the [write_str docs](https://doc.rust-lang.org/std/fmt/trait.Write.html#tymethod.write_str):
> This method can only succeed if the entire string slice was successfully written, and this method will not return until all data has been written or an error occurs.

wmmc88: But i think more pertinent is that returning err with partial write gives the user of the buffer an option to flush the buffer and continue writing (similar to the old behavior of dbgprint)

wmmc88 (commented):
@leon-xd is there a reason that you removed the buffered printing and flushing functionality?

wmmc88:
@leon-xd per conventional commit spec, could you add a footer explaining the breaking change being made

crates/wdk/src/fmt.rs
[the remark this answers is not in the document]
leon-xd: Added this too.

crates/wdk/src/fmt.rs
leon-xd: We have gotten rid of this by asserting N > 0.

crates/wdk/src/fmt.rs
leon-xd: Removed

crates/wdk/src/fmt.rs
leon-xd: Added this

crates/wdk/src/fmt.rs
leon-xd: Done

crates/wdk/src/fmt.rs
leon-xd: changed this

crates/wdk/src/fmt.rs
leon-xd: Now that I pushed `WdkFlushableFormatBuffer` does this change the behavior that we would expect out of `WdkFormatBuffer`?

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,305 @@
...
+
+        // Normal write: append the full string.
+        self.buffer[self.used..self.used + s.len()].copy_from_slice(s.as_bytes());
+        self.used += s.len();
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod test {
```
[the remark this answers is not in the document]
leon-xd: done

crates/wdk/src/print.rs
```diff
@@ -142,21 +148,33 @@ macro_rules! dbg {
...
-
-            if fmt::write(&mut buffered_writer, args).is_ok() {
-                buffered_writer.flush();
-            } else {
-                unreachable!("DbgPrintBufWriter should never fail to write");
+            let mut buffered_writer: crate::WdkFormatBuffer = crate::WdkFormatBuffer::new();
+            // We do not care whether this is `Ok` or `Err` right now. If `Err` will simply write all until overflow.
+            // TODO: create custom `flush` param in `WdkFormatBuffer` to preserve old functionality
+            let _ = fmt::write(&mut buffered_writer, args);
+
```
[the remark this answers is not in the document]
leon-xd: Removed

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,305 @@
...
+
+    /// Returns a UTF-8 view over the written bytes.
+    ///
+    /// Only the bytes successfully written are interpreted as UTF-8.
+    ///
+    /// # Errors
+    /// Returns an error if the written bytes are not valid UTF-8.
+    pub fn as_str(&self) -> Result<&str, Utf8Error> {
+        core::str::from_utf8(&self.buffer[..self.used])
+    }
```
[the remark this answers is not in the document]
leon-xd: &str can be created from a buffer which does not have to be guaranteed as valid UTF-8. this is incorrect.

leon-xd:
> @leon-xd is there a reason that you removed the buffered printing and flushing functionality?

@wmmc88  Thought about this for a bit and realized I should just prevent the breaking change if possible. So added a new type, `WdkFlushableFormatBuffer` as a wrapper around `WdkFormatBuffer` that includes a `flush` function that the user defines. Print should work as it did before, making this PR not contain a breaking change. Adjusting the PR title accordingly.

crates/wdk/src/lib.rs
```diff
@@ -28,14 +26,26 @@ pub use wdk_sys::NT_SUCCESS as nt_success;
...
+    driver_model__driver_type = "UMDF"
+))]
+mod fmt;
+
+#[cfg(any(
+    driver_model__driver_type = "WDM",
+    driver_model__driver_type = "KMDF",
+    driver_model__driver_type = "UMDF"
+))]
+pub use fmt::{WdkFlushableFormatBuffer, WdkFormatBuffer};
```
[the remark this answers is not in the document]
leon-xd: I think I'm comfortable just exporting it as `wdk::WdkFormatBuffer` and `wdk::WdkFlushableFormatBufer`. not sure if anyone else has any opinions

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,473 @@
...
+            let space = capacity - self.format_buffer.used;
+            self.format_buffer.buffer[self.format_buffer.used..self.format_buffer.used + space]
+                .copy_from_slice(&remaining[..space]);
+            self.format_buffer.used = capacity;
+
+            (self.flush_fn)(&self.format_buffer);
+            self.format_buffer.reset();
+
+            remaining = &remaining[space..];
+        }
```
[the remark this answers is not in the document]
leon-xd: Fixed

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,473 @@
...
+    /// use wdk::WdkFormatBuffer;
+    /// let _ = WdkFormatBuffer::<0>::new();
+    /// ```
+    #[must_use]
+    pub const fn new() -> Self {
+        const { assert!(N > 0, "N must be at least 1") }
+        Self {
+            buffer: [0; N],
+            used: 0,
+        }
```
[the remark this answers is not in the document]
leon-xd: This is valid Rust — `const { assert!(...) }` is a const block expression, and a trailing expression without a semicolon before the next statement is fine here. The code compiles and all tests pass.

crates/wdk/src/fmt.rs
gurry: Given the contract of `write_str` shared by @wmmc88 we must return an error in this case and given our unique requirement we cannot avoid a partial write even when we error. So let's keep the current behaviour then.

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,481 @@
...
+            self.format_buffer.reset();
+
+            remaining = &remaining[space..];
+        }
+
+        // Remaining bytes fit in the buffer.
+        self.format_buffer.buffer
+            [self.format_buffer.used..self.format_buffer.used + remaining.len()]
+            .copy_from_slice(remaining);
+        self.format_buffer.used += remaining.len();
```
gurry: We should auto-flush the remainder as well. Leaving that responsibility on the caller seems unnecessary and makes the type just a bit awkward to use.

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,481 @@
...
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
gurry: @leon-xd Copilot is right. We should end at a clean char boundary not a byte boundary. `as_str` returning an error is not the only problem. A bigger problem would be `DbgPrint` emitting garbled output if the last char gets cut in the middle.

We should also add a test for this.

crates/wdk/src/lib.rs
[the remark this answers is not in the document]
wmmc88: i agree with copilot. should reduce the duplication and its typically convention to do stuff like `wdk::FormatBuffer`

crates/wdk/src/fmt.rs
wmmc88: i think i disagree here. if you're calling `write!` multiple times in a row with long strings, i don't see why it should be flushing the remainders at every call. a flush on drop would make sense though imo

gurry: If this is a general-purpose type I would agree with you. If it's meant to be used only in `println!` then `_print` is already doing that i.e. flushing on every call.

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,481 @@
...
+        }
+        Self {
+            buffer: [0; N],
+            used: 0,
+        }
+    }
+
+    /// Resets the buffer to its initial empty state.
+    pub fn reset(&mut self) {
+        self.buffer = [0; N];
```
wmmc88: open question: should this even bother to zero out the buffer since the buffer is private?

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,481 @@
...
+    /// Returns an error if the written bytes are not valid UTF-8.
+    pub fn as_str(&self) -> Result<&str, Utf8Error> {
+        core::str::from_utf8(&self.buffer[..self.used])
+    }
+
+    /// Returns a C string view up to the first `NUL` byte.
+    ///
+    /// The buffer always contains a NUL terminator because `write_str`
+    /// reserves the last byte.
+    pub const fn as_cstr(&self) -> &CStr {
```
wmmc88: what happens when we have internal null since that's valid in rust string?

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,481 @@
...
+}
+
+impl<F: FnMut(&WdkFormatBuffer<N>), const N: usize> fmt::Write for WdkFlushableFormatBuffer<F, N> {
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        let capacity = N - 1;
+        let mut remaining = &s.as_bytes()[..];
+
+        // Fill what fits, flush, continue with the rest.
+        while remaining.len() > capacity - self.format_buffer.used {
+            let space = capacity - self.format_buffer.used;
```
wmmc88: nit:
```suggestion
            let remaining_space= capacity - self.format_buffer.used;
```

crates/wdk/src/fmt.rs
wmmc88: These types are meant to be generic. they're `no_std` compatible ways of doing string formatting and can be useful in many places outside of just the print.rs since its `pub`

crates/wdk/src/fmt.rs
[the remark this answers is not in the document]
wmmc88: yeah, i think this should be chunking off of graphemes

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,481 @@
...
+        }
+
+        // Normal write: append the full string.
+        self.buffer[self.used..self.used + s.len()].copy_from_slice(s.as_bytes());
+        self.used += s.len();
+        Ok(())
+    }
+}
+
+/// A [`WdkFormatBuffer`] wrapper that auto-flushes on overflow.
```
wmmc88: implement flush on drop?

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,481 @@
...
+            )
+        }
+        Self {
+            buffer: [0; N],
+            used: 0,
+        }
+    }
+
+    /// Resets the buffer to its initial empty state.
+    pub fn reset(&mut self) {
```
wmmc88: nit: Rust ecosystem convention is `clear()` for "empty the contents" (`Vec::clear()`, `String::clear()`, etc.). `reset()` implies broader state reset

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,481 @@
...
+    }
+
+    /// Returns a UTF-8 view over the written bytes.
+    ///
+    /// Only the bytes successfully written are interpreted as UTF-8.
+    ///
+    /// # Errors
+    /// Returns an error if the written bytes are not valid UTF-8.
+    pub fn as_str(&self) -> Result<&str, Utf8Error> {
+        core::str::from_utf8(&self.buffer[..self.used])
```
wmmc88: this is guaranteed to always succeed

wmmc88 (changes requested):

crates/wdk/src/fmt.rs
gurry: Okay. Then it's fine.

crates/wdk/src/fmt.rs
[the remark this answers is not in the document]
gurry: Ideally it should be graphemes not chars, but that will require pulling in the unicode_segmentation crate and might increase binary size a bit as well.

crates/wdk/src/fmt.rs
gurry: There's no need to zero out.

crates/wdk/src/fmt.rs
gurry: We could skip any embedded nulls in the incoming string when writing to the buffer. Or replace them with `?`s.

crates/wdk/src/fmt.rs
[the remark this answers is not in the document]
leon-xd: I ended up chunking based off code points, not graphemes. Hope that's an ok middle ground? Otherwise I can try pulling in the unicode_segmentation crate or adding some functionality myself. Let me know @gurry @wmmc88

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,481 @@
...
+
+    #[test]
+    fn ref_sanity_check() {
+        let mut fmt_buffer: WdkFormatBuffer = WdkFormatBuffer::new();
+        let world: &str = "world";
+        assert!(write!(&mut fmt_buffer, "Hello {world}!").is_ok());
+
+        // borrow fmt_buffer -- while this is in scope we cannot edit fmt_buffer
+        let buf_str = fmt_buffer.as_str().unwrap();
+        // write!(&mut fmt_buffer, "buf_str is still in scope so we cannot edit!)
```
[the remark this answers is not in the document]
leon-xd: fixed

crates/wdk/src/fmt.rs
leon-xd: after a flush the format buffer calls reset. i'd like to keep it if possible?

crates/wdk/src/fmt.rs
leon-xd: it writes until the first null, I believe this is documented

crates/wdk/src/fmt.rs
leon-xd: fixed

crates/wdk/src/fmt.rs
leon-xd: renamed

crates/wdk/src/fmt.rs
leon-xd: yessir added

crates/wdk/src/fmt.rs
leon-xd: renamed

crates/wdk/src/fmt.rs
leon-xd: i see -- nothing gets beyond used so it doesn't really matter what's in the rest of the buffer. removed.

crates/wdk/src/lib.rs
wmmc88: do these need to be gated?

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,582 @@
...
+        // into the buffer.
+        unsafe { core::str::from_utf8_unchecked(&self.buffer[..self.used]) }
+    }
+
+    /// Returns a C string view up to the first `NUL` byte.
+    ///
+    /// The buffer always contains a NUL terminator because `write_str`
+    /// reserves the last byte.
+    #[must_use]
+    pub const fn as_cstr(&self) -> &CStr {
```
wmmc88: nit: as_c_str

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,582 @@
...
+impl<const N: usize> fmt::Write for WdkFormatBuffer<N> {
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        // The last byte (buffer[N-1]) is reserved for the NUL terminator
+        // so that the buffer always contains a valid `CStr`.
+        let capacity = N - 1;
+        let remaining = capacity - self.used;
+
+        // Overflow: copy what fits at a char boundary and signal error.
+        if s.len() > remaining {
+            let fit = s.floor_char_boundary(remaining);
```
gurry, line 162: `WdkFormatBuffer` is just dumb `cstr`. It should not have any logic related to splitting at Utf8 char boundary. Otherwise we risk silently dropping parts of the string. Only `WdkFlushableFormatBuffer` should have this logic because it can adequately deal with the situation by flushing.

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,582 @@
...
+}
+
+impl<const N: usize> Default for WdkFormatBuffer<N> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<const N: usize> fmt::Write for WdkFormatBuffer<N> {
+    fn write_str(&mut self, s: &str) -> fmt::Result {
```
gurry, line 154: No one is using this method so do we need it? If not, do we need the the whole `WdkFormatBuffer` type?

`WdkFlushableFormatBuffer` could easily get away with an array and a `used` field directly inside it. Not clear what value `WdkFormatBuffer` is adding.

crates/wdk/src/fmt.rs
leon-xd: Hm I suppose but do you believe that there's an issue with having different functionality between the flushable/non-flushable buffer types?

crates/wdk/src/fmt.rs
leon-xd: I think there's still good reason to keep WdkFormatBuffer around on its own. Any WDK call that takes a `PCSTR` can just use `as_cstr()` directly from a `WdkFormatBuffer`.

In the future if we add wide buffer support for `UNICODE_STRING` the flushable buffer can wrap that one too.

crates/wdk/src/lib.rs
leon-xd: removed

crates/wdk/src/print.rs
```diff
@@ -142,20 +154,36 @@ macro_rules! dbg {
 /// detail and should never be called directly, but must be public to be useable
 /// by the print! and println! macro
 ///
-/// # Panics
-///
-/// Panics if an internal null byte is passed in
+/// Interior NUL bytes in the formatted output will cause the string to be
+/// truncated at the first NUL.
```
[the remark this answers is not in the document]
leon-xd: fixed

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,576 @@
...
+    /// let _ = WdkFormatBuffer::<1>::new();
+    /// ```
+    #[must_use]
+    pub const fn new() -> Self {
+        const {
+            assert!(
+                N >= 2,
+                "N must be at least 2 (one byte of content plus the NUL terminator)"
+            )
+        }
```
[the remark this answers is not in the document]
leon-xd: fixed

crates/wdk/src/fmt.rs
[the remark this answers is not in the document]
wmmc88: i think this is fine for now. Code-point chunking via floor_char_boundary is a good middle ground. can reevalutate this in the future

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,582 @@
...
+        if s.len() > remaining {
+            let fit = s.floor_char_boundary(remaining);
+            self.buffer[self.used..self.used + fit].copy_from_slice(s[..fit].as_bytes());
+            self.used += fit;
+            return Err(fmt::Error);
+        }
+
+        // Normal write: append the full string.
+        self.buffer[self.used..self.used + s.len()].copy_from_slice(s.as_bytes());
+        self.used += s.len();
```
wmmc88: Bug: after clear+write, `as_cstr()` returns stale bytes from the previous write because there is no NUL at `buffer[used]`. Same for the tail write in `WdkFlushableFormatBuffer::write_str` (line 209). 

Should also add a regression test: write full capacity -> clear -> write shorter -> assert `as_cstr()` matches `as_str()`.

crates/wdk/src/fmt.rs
wmmc88: @gurry the point of the non-flushing type is primarily to construct `CStr` types using typical Rust `fmt` apis, without needing to pull in `alloc`. i think having the two types here makes sense

crates/wdk/src/fmt.rs
wmmc88: this type also has a `as_str` so it must segment by code points at least. otherwise `as_str` could return invalid `str`

wmmc88: and keeping `as_str` is useful so this type can be used to construct `str` via `fmt` without needing to pull in `alloc`

crates/wdk/src/fmt.rs
wmmc88: Possible fix:
Maintain the invariant buffer[used] == 0 after every mutation:
clear(): add self.buffer[0] = 0; (covers clear() → as_cstr() without an intervening write)

WdkFormatBuffer::write_str: add self.buffer[self.used] = 0; after both the normal path (line 119) and the overflow path (line 113)

WdkFlushableFormatBuffer::write_str: add self.format_buffer.buffer[self.format_buffer.used] = 0; after the tail write (line 209)
The index is always safe since used ≤ N-1 (the reserved NUL slot).

crates/wdk/src/fmt.rs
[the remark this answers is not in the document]
gurry: Chars is okay. Not using graphemes will mangle/break up ligatures like characters with diacritics, matras in Indic languages and certain emojis. English will work fine though. We can come back to this problem later.

crates/wdk/src/fmt.rs
gurry: Sure. That's not a bad argument, although I would have introduced something like this only when it is actually required. But let's go with it.

wmmc88: ah. I should've put this context somewhere, but we have some drivers that need something like it :) I pushed @leon-xd  to make it available here since we had something similar in use for print already

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,582 @@
...
+
+            remaining = &remaining[split..];
+        }
+
+        // Remaining bytes fit in the buffer.
+        self.format_buffer.buffer
+            [self.format_buffer.used..self.format_buffer.used + remaining.len()]
+            .copy_from_slice(remaining.as_bytes());
+        self.format_buffer.used += remaining.len();
+        Ok(())
```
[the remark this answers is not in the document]
leon-xd: fixed via `append_bytes` helper function

crates/wdk/src/fmt.rs
leon-xd: fixed this via the `append_bytes` method and added the test

crates/wdk/src/fmt.rs
leon-xd: changed

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,626 @@
...
+        }
+
+        // Remaining bytes fit in the buffer.
+        self.format_buffer.append_bytes(remaining.as_bytes());
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod wdk_format_buffer_tests {
```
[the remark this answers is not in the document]
gurry: The test modules are no longer in sync with the names of the buffer types

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,630 @@
...
+
+        use super::*;
+
+        #[test]
+        fn new_is_zeroed() {
+            let fmt_buffer: FormatBuffer = FormatBuffer::new();
+            assert_eq!(fmt_buffer.used, 0);
+            assert_eq!(fmt_buffer.buffer.len(), 512);
+            assert!(fmt_buffer.buffer.iter().all(|&b| b == 0));
+        }
```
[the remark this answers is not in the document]
leon-xd: the second assert takes care of this

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,630 @@
...
+    pub const fn as_c_str(&self) -> &CStr {
+        match CStr::from_bytes_until_nul(&self.buffer) {
+            Ok(cstr) => cstr,
+            // `unreachable!()` with a message uses `format_args!`, which is
+            // not const-compatible. Use `panic!` with a string literal instead.
+            Err(_) => {
+                panic!("internal error: entered unreachable code: buffer must contain a NUL byte")
+            }
+        }
+    }
```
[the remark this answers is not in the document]
leon-xd: fixed

crates/wdk/src/fmt.rs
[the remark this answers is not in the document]
leon-xd: added constant definition to test

crates/wdk/src/print.rs
```diff
@@ -142,20 +154,37 @@ macro_rules! dbg {
...
 
-            if fmt::write(&mut buffered_writer, args).is_ok() {
-                buffered_writer.flush();
-            } else {
-                unreachable!("DbgPrintBufWriter should never fail to write");
-            }
+            // For N=512, write_str cannot fail: the largest UTF-8 code point
+            // is 4 bytes, which always fits in the 511-byte capacity. Overflow
+            // is handled by flushing. Errors from Display impls are silently
+            // dropped — partial output is acceptable for debug printing.
```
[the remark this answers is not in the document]
leon-xd: rewrote comment

crates/wdk/Cargo.toml
```diff
@@ -17,8 +17,6 @@ readme.workspace = true
 repository.workspace = true
 
 [features]
-alloc = []
-default = ["alloc"]
 nightly = ["wdk-sys/nightly"]
```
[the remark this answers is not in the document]
leon-xd: rewrote PR description to match

crates/wdk/src/fmt.rs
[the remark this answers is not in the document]
leon-xd: fixed this

crates/wdk/src/lib.rs
```diff
@@ -28,14 +26,16 @@ pub use wdk_sys::NT_SUCCESS as nt_success;
...
-    ),
+    driver_model__driver_type = "WDM",
+    driver_model__driver_type = "KMDF",
     driver_model__driver_type = "UMDF",
 ))]
 mod print;
 
+mod fmt;
+
+pub use fmt::{WdkFlushableFormatBuffer, WdkFormatBuffer};
```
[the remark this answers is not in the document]
leon-xd: changed

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,631 @@
...
+        fn write() {
+            let mut fmt_buffer: FormatBuffer = FormatBuffer::new();
+            let world: &str = "world";
+            assert!(write!(&mut fmt_buffer, "Hello {world}!").is_ok());
+
+            let mut cmp_buffer: [u8; 512] = [0; 512];
+            let cmp_str: &str = "Hello world!";
+            cmp_buffer[..cmp_str.len()].copy_from_slice(cmp_str.as_bytes());
+
+            assert_eq!(fmt_buffer.buffer, cmp_buffer);
```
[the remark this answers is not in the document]
leon-xd: The default should always be 512, this test should fail if the default size deviates from that.

crates/wdk/Cargo.toml
```diff
@@ -17,8 +17,6 @@ readme.workspace = true
 repository.workspace = true
 
 [features]
```
wmmc88, line 19: **nit:** Consider keeping `alloc = []` as a deprecated no-op for backward compatibility, but remove it from `default`. Downstream crates that specify `features = ["alloc"]` will get a hard build error after this change. Keeping it as an empty non-default feature avoids the breakage while still removing all functional use.

```toml
[features]
# Deprecated: this feature is no longer required and is not a default
# feature. Kept as a no-op for backward compatibility with crates that
# still reference it.
alloc = []
nightly = ["wdk-sys/nightly"]
```

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,631 @@
...
+    const N: usize = DEFAULT_WDK_FORMAT_BUFFER_SIZE,
+> {
+    format_buffer: FormatBuffer<N>,
+    flush_fn: F,
+}
+
+impl<F: FnMut(&FormatBuffer<N>), const N: usize> FlushableFormatBuffer<F, N> {
+    /// Creates a new flushable writer with the given flush closure.
+    #[must_use]
+    pub const fn new(flush_fn: F) -> Self {
```
wmmc88, line 201: **suggestion:** The flush closure runs from `Drop::drop` (line 191). Consider adding a `# Panics` doc section noting the propagation behavior:

```rust
/// # Panics
///
/// If `flush_fn` panics, this function will also panic. If `flush_fn`
/// panics during drop, the `Drop` impl will also panic.
```

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,631 @@
...
+        f.debug_struct("FormatBuffer")
+            .field("used", &self.used)
+            .field("capacity", &(N - 1))
+            .field("content", &self.as_str())
+            .finish_non_exhaustive()
+    }
+}
+
+impl<const N: usize> fmt::Write for FormatBuffer<N> {
+    fn write_str(&mut self, s: &str) -> fmt::Result {
```
wmmc88, line 154: **suggestion:** This impl does a partial write (up to the last char boundary that fits) and then returns `Err`. That's legal per `fmt::Write` but consider adding an `# Errors` section to the trait impl doc to make this explicit:

```rust
/// # Errors
///
/// Returns [`fmt::Error`] if `s` exceeds the remaining capacity. Bytes
/// that fit (clamped to the last UTF-8 char boundary) **are** written to
/// the buffer before the error is returned.
```

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,631 @@
...
+            buffer: [0; N],
+            used: 0,
+        }
+    }
+
+    /// Clears the buffer, resetting it to its initial empty state.
+    pub const fn clear(&mut self) {
+        self.used = 0;
+        self.buffer[0] = 0;
+    }
```
wmmc88, line 66: **suggestion:** Consider adding basic introspection methods. Right now callers can't query the buffer state without going through `as_str().len()`. These are cheap `const fn`s:

```rust
/// Returns the number of bytes currently written.
#[must_use]
pub const fn len(&self) -> usize { self.used }

/// Returns the usable capacity in bytes (`N - 1`, excluding the
/// reserved NUL terminator).
#[must_use]
pub const fn capacity(&self) -> usize { N - 1 }

/// Returns `true` if no bytes have been written.
#[must_use]
pub const fn is_empty(&self) -> bool { self.used == 0 }
```

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,631 @@
...
+/// let mut buf = FormatBuffer::<16>::new();
+/// write!(&mut buf, "hello {}", 42).unwrap();
+///
+/// let s = buf.as_str();
+/// assert_eq!(s, "hello 42");
+///
+/// let c = buf.as_c_str();
+/// assert_eq!(c.to_bytes(), b"hello 42");
+/// ```
+pub struct FormatBuffer<const N: usize = DEFAULT_WDK_FORMAT_BUFFER_SIZE> {
```
wmmc88, line 32: **nit:** `FormatBuffer` is `Copy`-friendly (all fields are `Copy`), so it can trivially derive `Clone`. This is useful if callers want to snapshot buffer state before a potentially-overflowing write:

```rust
impl<const N: usize> Clone for FormatBuffer<N> {
    fn clone(&self) -> Self { *self }
}
```

(Or derive `Copy` + `Clone` if you want full value semantics.)

crates/wdk/Cargo.toml
wmmc88: This is while still removing it from the default features array

examples/sample-kmdf-driver/src/lib.rs
wmmc88, line 1: what triggered these changes? looks fine on master?

wmmc88 (dismissed):

crates/wdk/Cargo.toml
leon-xd: added this, good catch. didn't know about this for backwards compatibility will keep this in mind for the future

crates/wdk/src/fmt.rs
leon-xd: yadded

crates/wdk/src/fmt.rs
leon-xd: added

crates/wdk/src/fmt.rs
leon-xd: Added all of these

crates/wdk/src/fmt.rs
leon-xd: Added

examples/sample-kmdf-driver/src/lib.rs
leon-xd: Old version of nightly locally -- reverted these changes

crates/wdk/src/fmt.rs
```diff
@@ -33,6 +33,15 @@ pub struct FormatBuffer<const N: usize = DEFAULT_WDK_FORMAT_BUFFER_SIZE> {
...
 
+impl<const N: usize> Clone for FormatBuffer<N> {
+    fn clone(&self) -> Self {
+        Self {
+            buffer: self.buffer,
+            used: self.used,
+        }
+    }
+}
+
```
gurry: We can just derive these traits with `#[derive(Copy, Clone)]` instead of a manual impl

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,669 @@
...
+        // Only scan up to `used + 1` — the NUL is guaranteed at `buffer[used]`.
+        match CStr::from_bytes_until_nul(self.buffer.split_at(self.used + 1).0) {
+            Ok(cstr) => cstr,
+            // `unreachable!()` with a message uses `format_args!`, which is
+            // not const-compatible. Use `panic!` with a string literal instead.
+            Err(_) => {
+                panic!("internal error: entered unreachable code: buffer must contain a NUL byte")
+            }
+        }
+    }
```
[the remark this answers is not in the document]
leon-xd: bumped msrv

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,669 @@
...
+
+            if split == 0 {
+                if self.format_buffer.used == 0 {
+                    // A single character doesn't fit in the entire buffer.
+                    return Err(fmt::Error);
+                }
+                // Buffer has content but no room for the next char — flush and retry.
+                (self.flush_fn)(&self.format_buffer);
+                self.format_buffer.clear();
+                continue;
```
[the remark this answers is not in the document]
leon-xd: replaced

crates/wdk/src/fmt.rs
```diff
@@ -0,0 +1,669 @@
...
+            }
+
+            self.format_buffer
+                .append_bytes(&remaining.as_bytes()[..split]);
+
+            (self.flush_fn)(&self.format_buffer);
+            self.format_buffer.clear();
```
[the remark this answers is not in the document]
leon-xd: replaced

crates/wdk/src/fmt.rs
leon-xd: I purposely don't want to derive Copy -- given the resource constrained environment of drivers I don't think we want any accidental copies of buffers passed through. I got confused between Copy and Clone derivation dependencies, looking it up we can just `#[derive(Clone)]`, just pushed a change

Cargo.toml
```diff
@@ -19,7 +19,7 @@ edition = "2024"
 license = "MIT OR Apache-2.0"
 readme = "README.md"
 repository = "https://github.com/microsoft/windows-drivers-rs"
-rust-version = "1.85.0"
+rust-version = "1.87.0"
```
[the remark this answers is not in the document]
leon-xd: fixed

crates/wdk/Cargo.toml
```diff
@@ -17,8 +17,8 @@ readme.workspace = true
...
+# Deprecated: no longer required. Kept as a no-op for backward compatibility.
 alloc = []
-default = ["alloc"]
 nightly = ["wdk-sys/nightly"]
```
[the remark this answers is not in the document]
leon-xd: added

gurry (approved):

wmmc88 (approved):

PR #621, opened 2026-03-04
docs: add AGENTS.md and CLAUDE.md for Copilot code review

wmmc88:
Adds AGENTS.md (and a CLAUDE.md pointer) so that GitHub Copilot code review has the repo context it needs to give useful feedback: crate layering, unsafe discipline rules, lint configuration, and build/test commands. Also useful for other AI coding assistants that read these files.


AGENTS.md
```diff
@@ -0,0 +1,94 @@
...
+4. **`wdk`** — Safe idiomatic Rust wrappers around `wdk-sys`. Conditionally `no_std` for WDM/KMDF; std-capable for UMDF.
+5. **`wdk-alloc`** — Global allocator (`WdkAllocator`) for kernel-mode crates needing `alloc`.
+6. **`wdk-panic`** — Default panic handlers for WDK binaries.
+7. **`cargo-wdk`** — Cargo subcommand for building and packaging driver projects.
+
+**Key design patterns:**
+
+- Conditional compilation via custom `cfg` flags (not Cargo features) distinguishes driver types. These flags are emitted by `wdk-build` during the build script phase.
+- `wdk-sys` has feature flags for optional WDK subsystem bindings: `gpio`, `hid`, `spb`, `storage`, `usb`, `parallel-ports`.
+- All crates expose a `nightly` feature for functionality gated behind unstable Rust features.
```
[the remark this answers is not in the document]
wmmc88: fixed

AGENTS.md
```diff
@@ -0,0 +1,94 @@
...
+```shell
+cargo make wdk-pre-commit-flow
+```
+
+## Code Conventions
+
+- **Unsafe discipline:** `unsafe_op_in_unsafe_fn` is **forbid** at workspace level. Every unsafe block must have exactly one unsafe operation (`multiple_unsafe_ops_per_block = "forbid"`) and a `// SAFETY:` comment (`undocumented_unsafe_blocks = "forbid"`). Unnecessary safety docs are also forbidden.
+- **FFI naming:** Bindings in `wdk-sys` retain their original C names and style (e.g., PascalCase functions). Safe Rust wrappers in `wdk` follow [Rust naming conventions (RFC 430)](https://rust-lang.github.io/api-guidelines/naming.html).
+- **Formatting:** `rustfmt` uses nightly-only unstable options (see `rustfmt.toml`). Import grouping is `StdExternalCrate`, granularity is `Crate`.
+- **Lints:** Clippy `all` is deny, `pedantic`/`nursery`/`cargo` are warn. All workspace members must include `[lints] workspace = true` (except `wdk-sys` which overrides some to `deny` for generated code opt-out).
```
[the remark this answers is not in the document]
wmmc88: fixed

AGENTS.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Pre-commit (runs everything)
+
+```shell
+cargo make wdk-pre-commit-flow
+```
+
+## Code Conventions
+
+- **Unsafe discipline:** `unsafe_op_in_unsafe_fn` is **forbid** at workspace level. Every unsafe block must have exactly one unsafe operation (`multiple_unsafe_ops_per_block = "forbid"`) and a `// SAFETY:` comment (`undocumented_unsafe_blocks = "forbid"`). Unnecessary safety docs are also forbidden.
```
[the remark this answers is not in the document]
wmmc88: fixed

AGENTS.md
gurry, line 1: A lot of information here is duplicated from `README.md` or `CONTRIBUTING.md`. This file could just point to those files for such info. Will keep agents automatically updated when those files change.

wmmc88: The duplication is intentional. Copilot code review (and other agents) only reads the instruction file itself — it [doesn't necessarily always follow links or read referenced files like README.md or CONTRIBUTING.md](https://docs.github.com/en/copilot/tutorials/use-custom-instructions)... and when it does it may [do it poorly since its not meant for llm ingestion.](https://docs.github.com/en/copilot/tutorials/use-custom-instructions#keep-instructions-short-and-focused). The AGENTS.md spec similarly recommends keeping critical rules self-contained because [many agents won't resolve cross-file references and may miss instructions entirely](https://developers.openai.com/codex/guides/agents-md). If anything drifts, it's straightforward to regenerate the file — but a collection of pointers that agents ignore is worse than slightly stale content they actually consume.

wmmc88: oh also the common guidance these days is to just delete the copilot instructions/agents.md files and regenerate them with `/init` every now and then

gurry: The pages you linked do not say anything about referring to other files or the agents not following links. Maybe I'm missing something.

gurry, line 1: Should we name it `copilot-instructions.md` instead of `AGENTS.md`. As per [this discussion](https://github.com/orgs/community/discussions/164530#discussioncomment-14246235) vscode seems to ignore `AGENTS.md`

wmmc88: that is stale information. it supports agents.md (and we want it .to be more open to other agents as well). AFAIU the only widely used tool that doesn't natively support agents.md is claude code

wmmc88: https://code.visualstudio.com/docs/copilot/customization/custom-instructions#_use-an-agentsmd-file

wmmc88: Sorry, you're right — I linked the wrong sections. Here are the corrected references:

1. The same tutorial page, [Unsupported instruction types](https://docs.github.com/en/copilot/tutorials/use-custom-instructions#unsupported-instruction-types) section — it says Copilot code review can't follow external links and the workaround is to "copy the relevant content directly into your instruction file instead." That's about URLs specifically, but the workaround language is telling: the tool processes what's in the instruction file, not content it has to go fetch.

2. The [AGENTS.md spec](https://agents.md) itself — "README.md files are for humans … AGENTS.md complements this by containing the extra, sometimes detailed context coding agents need." The whole premise is that agents need their own self-contained instruction file rather than relying on human-facing docs.

3. Copilot code review also has a hard [4,000-character limit on instruction files](https://docs.github.com/en/copilot/tutorials/use-custom-instructions#understanding-how-github-copilot-code-review-processes-instructions) and there's no documented mechanism for it to expand that by resolving file references. So even if it could follow a pointer to README.md in theory, those chars would eat into a very tight budget.

I don't have a single page that says "code review ignores cross-file references" in those exact words — the argument is more that the design of these systems (character limits, explicit "copy content in" guidance, separate agent-focused files) all point in the same direction.

PR #622, opened 2026-03-04
ci: use Install-PSResource instead of Install-Module for WinGet

wmmc88:
Replace `Install-Module` (PowerShellGet v2) with `Install-PSResource` (PSResourceGet, built into PowerShell 7.4+) for installing the `Microsoft.WinGet.Client` module. PSResourceGet has its own repository system that is independent of the legacy NuGet pipeline, avoiding transient PSGallery registration failures on freshly provisioned `windows-2025` runners.

Also switches `codeql.yaml` to use the shared `install-winget` composite action instead of inlining its own `Install-Module` step.

Example failure: https://github.com/microsoft/windows-drivers-rs/actions/runs/22659747111/job/65677070595?pr=621

Upstream issue: https://github.com/actions/runner-images/issues/13758

Stress-tested in #626 (expanded LLVM matrix).

### DIFF
--- a/.github/workflows/code-formatting-check.yaml
+++ b/.github/workflows/code-formatting-check.yaml
@@ -41,31 +41,18 @@ jobs:
       - name: Install Rust Toolchain (Stable)
         uses: dtolnay/rust-toolchain@stable
 
-      - name: Install taplo-cli
-        uses: taiki-e/install-action@v2
+      # Using cache-cargo-install-action to cache the compiled binary.
+      # Once taplo ships a release with field-level `keys` matching, this can
+      # be simplified back to install-action with a normal version pin.
+      - name: Install taplo-cli from pinned revision b673b44d
+        uses: taiki-e/cache-cargo-install-action@v3
         with:
           tool: taplo-cli
+          git: https://github.com/tamasfe/taplo
+          rev: b673b44d
+          locked: true
+        env:
+          RUSTFLAGS: ""
 
       - run: taplo fmt --check --diff
-        name: Check TOML files formatting
-
-  cargo-sort:
-    name: Check Dependencies are sorted lexicographically
-    runs-on: windows-2025
-
-    steps:
-      - name: Checkout Repository
-        uses: actions/checkout@v6
-
-      - name: Install Rust Toolchain (Stable)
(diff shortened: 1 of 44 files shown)


.github/workflows/build.yaml
```diff
@@ -38,6 +38,9 @@ jobs:
 
         llvm:
           - 17.0.6
+          - 19.1.7
+          - 20.1.8
+          - 21.1.2
```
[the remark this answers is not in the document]
wmmc88: this is fine for now. if we find it is hindering us, we can reduce it later

.github/actions/install-winget/action.yaml
```diff
@@ -110,26 +110,35 @@ runs:
...
+            Install-PSResource -Name Microsoft.WinGet.Client -Repository PSGallery -Version $best -TrustRepository -Quiet
           } else {
             Write-Warning "No module with major version $cliMajor found on PSGallery. Installing latest version."
-            Install-Module -Name Microsoft.WinGet.Client -Repository PSGallery -Force
+            Install-PSResource -Name Microsoft.WinGet.Client -Repository PSGallery -TrustRepository -Quiet
           }
         }
 
-        $moduleVersion = (Get-InstalledModule Microsoft.WinGet.Client).Version.ToString()
+        $moduleVersion = (Get-Module -ListAvailable Microsoft.WinGet.Client | Select-Object -First 1).Version.ToString()
```
leon-xd: is it possible that there is multiple versions of winget at this point? if so since we're taking `-First` it could log the wrong version.

leon-xd: `$moduleVersion = (Get-InstalledPSResource -Name Microsoft.WinGet.Client | Sort-Object Version -Descending | Select-Object -First 1).Version.ToString()`

wmmc88: nice catch. @copilot add this fix

wmmc88: welp i tried...

```suggestion
        $moduleVersion = (Get-InstalledPSResource -Name Microsoft.WinGet.Client | Sort-Object Version -Descending | Select-Object -First 1).Version.ToString()
```

PR #643, opened 2026-04-16
refactor: fix `collapsable_if` lint, rust 1.95 `assert_matches` destabilization

leon-xd:
Fixes `collapsable_if` lint failing in CI pipelines.

The `assert_matches` stabilization that was expected to ship with Rust 1.95.0 was [reverted from the 1.95 beta](https://github.com/rust-lang/rust/pull/154999) due to an [inconsistency in temporary scoping behavior](https://github.com/rust-lang/rust/issues/154406) between `assert_matches!`/`assert_eq!` and their `debug_` variants.

We gated `assert_matches` usage based on the nightly compiler warning that the feature was "stable since 1.95.0" (see #612). Since the stabilization was reverted before the actual release, this gate now fires on stable 1.95 and tries to use an unstable API, breaking CI pipelines.

This bumps the gate to `1.96.0`. Beta 1.96 currently still has `assert_matches` available, though that may change if the destabilization propagates.

### DIFF
--- a/crates/wdk-build/build.rs
+++ b/crates/wdk-build/build.rs
@@ -14,8 +14,9 @@ fn main() {
 // Custom attributes cannot be applied to expressions yet, so separate functions
 // are required for `rustversion` conditional compilation: https://github.com/rust-lang/rust/issues/15701
 // TODO: Remove the `setup_assert_matches_stabilized_cfg` feature and related
-// code once the minimum supported Rust version includes stable `assert_matches`.
-// Stabilization was reverted from 1.95.0 beta: https://github.com/rust-lang/rust/pull/154999
+// code once the minimum supported Rust version includes stable
+// `assert_matches`. Stabilization was reverted from 1.95.0 beta:
+// https://github.com/rust-lang/rust/pull/154999
 #[rustversion::since(1.96.0)]
 fn setup_assert_matches_stabilized_cfg() {
     println!("cargo::rustc-cfg=assert_matches_stabilized");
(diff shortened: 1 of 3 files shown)


crates/wdk-build/build.rs
```diff
@@ -14,11 +14,13 @@ fn main() {
...
+// `assert_matches`. Stabilization was reverted from 1.95.0 beta:
+// https://github.com/rust-lang/rust/pull/154999
+#[rustversion::since(1.96.0)]
 fn setup_assert_matches_stabilized_cfg() {
     println!("cargo::rustc-cfg=assert_matches_stabilized");
 }
 
-#[rustversion::before(1.95.0)]
+#[rustversion::before(1.96.0)]
 const fn setup_assert_matches_stabilized_cfg() {}
```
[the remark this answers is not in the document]
leon-xd: its ok this is intentional, that was the purpose of the original fix and we want this eventually so we're keeping it

crates/wdk-build/build.rs
```diff
@@ -14,11 +14,13 @@ fn main() {
 // Custom attributes cannot be applied to expressions yet, so separate functions
 // are required for `rustversion` conditional compilation: https://github.com/rust-lang/rust/issues/15701
 // TODO: Remove the `setup_assert_matches_stabilized_cfg` feature and related
-// code once the minimum supported Rust version is 1.95.0 or later.
-#[rustversion::since(1.95.0)]
+// code once the minimum supported Rust version includes stable
+// `assert_matches`. Stabilization was reverted from 1.95.0 beta:
+// https://github.com/rust-lang/rust/pull/154999
```
wmmc88: this is irrelevant. this should probably be a comment pointing to the stabilization issue (which should be reopened now?)

leon-xd: fixed comment

wmmc88 (commented):
the PR description is a little messy. can't quite pin down why it seems off. also i think this should be labelled `refactor: fix xyz lint` based off past lint fix prs

wmmc88 (approved):

Alan632 (approved):

PR #649, opened 2026-04-27
feat: introduce sign mode for driver packaging, allowing skipping of signing steps

### LINKED ISSUE #588: `cargo wdk build`: Skip signtool
Say I want to release a production build. `cargo wdk build` always signs it with the test cert:

https://github.com/microsoft/windows-drivers-rs/blob/f1cf7f9e9d2a416aeed3175e0f044260013f1d21/crates/cargo-wdk/src/actions/build/package_task.rs#L241-L250

But in case of a production build, I don't want the test signature. I want the sys file without it, so that I can go and submit it for MS to sign.

Add an option to skip running signtool with `cargo wdk build`.

svasista-ms:
This PR adds a `--sign-mode` option to the `cargo wdk build` command that lets you turn off test signing which is useful for production/HLK scenarios.

Resolves #588

## Functionality

`--sign-mode` takes two values:

* `test`: generates a self-signed cert and signs artifacts with it. Same behavior as today
* `off`: skips all signing including cert generation

`test` is the default so if `--sign-mode` is omitted the command behaves exactly as today and thus remains backwards compatible. 

An error is returned if you try to use `--verify-signature` with `--sign-mode=off` because verification does not make sense when nothing is signed.

## Future Direction

This PR is part of a broader goal to replicate [Visual Studio's `SignMode` setting](https://learn.microsoft.com/en-us/windows-hardware/drivers/develop/driver-signing-properties) in `cargo-wdk`.  In the future we plan to:

- Introduce a third `--sign-mode` value called `prod` to enable a production signing mode that will not generate a test cert and will instead require a user-specified cert
- Add CLI options for specifying certs, timestamp servers and signing algorithms

## Changes
- New `sign_mode` field in `BuildArgs` [[1]](https://github.com/microsoft/windows-drivers-rs/pull/649/changes#diff-4861850c7366ce9ab803db669bc522b1265ae68a766778ab50c3a9976fde66d9R97)
- CLI-layer logic to reject use of `--verify-signature` with `--sign-mode=off` [[2]](https://github.com/microsoft/windows-drivers-rs/pull/649/changes#diff-4861850c7366ce9ab803db669bc522b1265ae68a766778ab50c3a9976fde66d9R171)
- New `SignMode` enum with `Test` and `Off` variants [[3]](https://github.com/microsoft/windows-drivers-rs/pull/649/changes#diff-a2e03ddb9aa9896d1c25be3314b2bfe2f59129c55be5540d992dd56341c34db6R53-R88).
- Plumbed `SignMode` from `BuildAction` through `PackageTask`
- In `PackageTask::run()`, gated `generate_certificate`, the certificate copy and `signtool` invocations on `SignMode::Test`[[4]](https://github.com/microsoft/windows-drivers-rs/pull/649/changes#diff-a6887c609082ccee97f37483a4395a3d87c8ccfd5d1b4a7ea0d79c3884536037R245-R263)
- Unit tests for the new paths [[5]](https://github.com/microsoft/windows-drivers-rs/pull/649/changes#diff-b30d5879afa7d114c914430fd892a49a824db8040ecc061332d4b3760f278e18R269-R1985)
- Integration tests [[6]](https://github.com/microsoft/windows-drivers-rs/pull/649/changes#diff-7f39e8d21b60832dceaba74ceb186dce58e45d2c660c6ce0674d1be3181a18e9R290-R399)

## Screenshots

`cargo wdk build --help` showing `--sign-mode`:

image

Building with `--sign-mode=off`:

image

Building with `--sign-mode=test`:

image

Building without the `--sign-mode` option -- falls back to the default (`test`) behavior:

image

Error when using `--verify-signature` with `--sign-mode=off`:

image


crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -287,6 +287,140 @@ mod kmdf_driver_with_target_override {
...
+        .args([
+            "-NoProfile",
+            "-NonInteractive",
+            "-Command",
+            &format!("(Get-AuthenticodeSignature -FilePath '{file_path}').Status.ToString()"),
+        ])
+        .output()
+        .expect("failed to invoke powershell to check Authenticode signature");
+    let status = String::from_utf8_lossy(&output.stdout).trim().to_string();
+    !status.is_empty() && status != "NotSigned"
```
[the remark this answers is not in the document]
svasista-ms: Dropped the `is_authenticode_signed` PowerShell check — too environment-fragile (failed locally several times due to a broken `Microsoft.PowerShell.Security` module and other issues; same risk on CI). This behavior is already covered: 
- integration test asserts no `WDRLocalTestCert.cer` in the final package 
- unit test `given_a_driver_project_when_sign_mode_is_off_then_signing_and_verification_steps_are_skipped` verifies signtool/inf2cat/infverif aren't invoked. 

If we want binary-level verification later, `signtool verify` would be a better option.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -236,24 +242,36 @@ impl<'a> PackageTask<'a> {
...
+                    WDR_LOCAL_TEST_CERT,
+                )?;
+                self.run_signtool_sign(
+                    &self.dest_cat_file_path,
+                    WDR_TEST_CERT_STORE,
+                    WDR_LOCAL_TEST_CERT,
+                )?;
+            }
+            SignMode::Off => {
+                info!("Sign mode is 'Off'; skipping certificate generation and signing");
```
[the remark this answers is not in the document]
svasista-ms: Checked how MSBuild/WDK handles this in `WindowsDriver.Common.targets` (10.0.26100.0): all signing targets are gated on `SignMode == TestSign|ProductionSign`, same as this PR. There's no mode-switch cleanup. I think we should let cleanup be handled by `cargo wdk clean` similar to `MSBuild`'s `Clean`. If this UX becomes an issue, I can revisit it and implement something to ensure the final package is not stale.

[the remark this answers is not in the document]
gurry: Why are we adding the cert to the package? Is that something Visual Studio does too?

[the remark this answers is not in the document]
svasista-ms: Good question! `cargo-wdk` actually follows the `cargo-make` based approach in [rust-driver-makefile.toml:452-468](https://github.com/microsoft/windows-drivers-rs/blob/ebc705a1e7a8bc5973625c5d435fe568068fdc6b/crates/wdk-build/rust-driver-makefile.toml#L452), which in turn was derived from WDK's MSBuild targets. Shipping the cert for "Test-Signing" workflows makes sense because the cert needs to be installed on the target (test) machine. I need to re-verify the behavior for C++ drivers in Visual Studio.

crates/cargo-wdk/src/actions/mod.rs
```diff
@@ -50,6 +50,42 @@ impl Display for Profile {
...
 }
 
+/// Driver signing mode used during the package phase.
+///
+/// * `Off` - Skip all signing-related steps (certificate generation, signing,
+///   and signature verification).
+/// * `Test` - Use test-signing, which currently auto-generates a self-signed
+///   certificate (`WDRLocalTestCert` in the `WDRTestCertStore` store) and signs
+///   the driver binary and catalog file with it.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
```
gurry:
```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
#[value(rename_all = "lower")]
```

In help etc. we want to say "test" and "off" instead of "Test" and "Off" as typing the former is easier for users.

crates/cargo-wdk/src/actions/build/package_task.rs
```suggestion
                info!("Sign mode is 'off'; skipping certificate generation and signing");
```

We want to use and promote lower case values of sign mode everywhere since they are easier to type.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -236,24 +242,36 @@ impl<'a> PackageTask<'a> {
...
+        }
         self.run_infverif()?;
         // Verify signatures only when --verify-signature flag = true is passed
+        // and signing was done (sign mode is not 'Off').
         if self.verify_signature {
-            info!("Verifying signatures for driver binary and cat file using signtool");
-            self.run_signtool_verify(&self.dest_driver_binary_path)?;
-            self.run_signtool_verify(&self.dest_cat_file_path)?;
+            if matches!(self.sign_mode, SignMode::Off) {
+                warn!("Skipping signature verification because sign mode is 'Off'");
```
```suggestion
                warn!("Skipping signature verification because sign mode is 'off'");
```

crates/cargo-wdk/src/actions/build/package_task.rs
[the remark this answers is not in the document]
gurry: Okay.

Yeah, let's verify the C++ behaviour.

gurry:
It's odd that we're printing the error twice:
image

Why is that? Is this true for other errors as well? What can we do to fix it?

crates/cargo-wdk/src/actions/build/tests.rs
```diff
@@ -265,6 +266,132 @@ pub fn given_a_driver_project_when_verify_signature_is_true_then_it_builds_succe
     );
 }
 
+// Given: A driver project
+// When: --sign-mode=off is provided
+// Then: It builds successfully and skips all certificate generation and
+// signing steps (no makecert / certmgr / signtool sign calls).
```
gurry: Remove these comments for all the new tests for consistency because we don't use them on existing tests either. The function names themselves are super verbose so no need for comments.

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -287,6 +287,116 @@ mod kmdf_driver_with_target_override {
     }
 }
 
+////////////////////////////////////////////////////////////////////////////////
+/// Sign-mode integration tests
+////////////////////////////////////////////////////////////////////////////////
+
```
gurry: Remove this header comment as we don't use such comments anywhere else in this file

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -287,6 +287,112 @@ mod kmdf_driver_with_target_override {
...
+        "expected validation error mentioning both flags, got: {stderr}"
+    );
+}
+
+/// `cargo wdk build --help` must advertise every supported option. Adding a
+/// new option to the `build` subcommand should also be reflected here so that
+/// future regressions (e.g., a flag accidentally hidden or removed) are caught
+/// at the CLI surface.
+#[test]
+fn build_help_advertises_all_options() {
```
krishnakumar4a4: This is one of the missing tests for both the build and new subcommands. It’s good that you’ve added it for the build subcommand.

We should make this test more comprehensive. We should assert on the entire help output, including the listed options, the help text for each option, grouping, and even indentation, if possible. This test should catch any changes in the help output, as this is what users read most often. Having such a test would be useful to detect changes to help text, options being mangled or incorrectly grouped, or indentation issues.

If we know that Clap always displays these options in a specific order and that this order will not change, we could do a simple snapshot match of the expected output text against the actual stdout.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -236,24 +242,36 @@ impl<'a> PackageTask<'a> {
...
+                    &self.dest_cat_file_path,
+                    WDR_TEST_CERT_STORE,
+                    WDR_LOCAL_TEST_CERT,
+                )?;
+            }
+            SignMode::Off => {
+                info!("Sign mode is 'off'; skipping certificate generation and signing");
+            }
+        }
         self.run_infverif()?;
```
krishnakumar4a4, line 252: I think run_infverif() would read better either before the signing block or after signature verification, so signing and signature verification stay grouped together.

If we move it before signing step, then we could do early return for [SignMode::Off](https://github.com/microsoft/windows-drivers-rs/pull/649/changes#diff-a6887c609082ccee97f37483a4395a3d87c8ccfd5d1b4a7ea0d79c3884536037R260) case above and potentially get rid off [this case](https://github.com/microsoft/windows-drivers-rs/pull/649/changes#diff-a6887c609082ccee97f37483a4395a3d87c8ccfd5d1b4a7ea0d79c3884536037R268) and the if condition?

crates/cargo-wdk/src/actions/build/tests.rs
```diff
@@ -265,6 +266,118 @@ pub fn given_a_driver_project_when_verify_signature_is_true_then_it_builds_succe
...
+        profile,
+        None,
+        verify_signature,
+        sample_class,
+        test_build_action,
+    );
+}
+
+#[test]
+pub fn given_a_sample_class_driver_project_when_sign_mode_is_off_then_signing_is_skipped_and_sample_infverif_still_runs()
```
krishnakumar4a4: If you do this https://github.com/microsoft/windows-drivers-rs/pull/649/changes#r3223695837

This test won't be required, I think.

crates/cargo-wdk/src/actions/build/tests.rs
```diff
@@ -265,6 +266,118 @@ pub fn given_a_driver_project_when_verify_signature_is_true_then_it_builds_succe
...
+        profile,
+        None,
+        verify_signature,
+        sample_class,
+        test_build_action,
+    );
+}
+
+#[test]
+pub fn given_a_driver_project_when_sign_mode_is_off_and_verify_signature_is_true_then_verification_is_skipped()
```
krishnakumar4a4: https://github.com/microsoft/windows-drivers-rs/pull/649/changes#r3223744162

krishnakumar4a4 (changes requested):

crates/cargo-wdk/src/actions/mod.rs
```diff
@@ -50,6 +50,27 @@ impl Display for Profile {
     }
 }
 
+/// Driver signing mode used during the package phase.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, clap::ValueEnum)]
+#[value(rename_all = "lower")]
+pub enum SignMode {
+    /// Skip signing entirely.
+    Off,
+    /// Use test-signing with an auto-generated self-signed certificate.
```
gurry:
```suggestion
    /// Sign with an auto-generated self-signed certificate.
```

Shorter and more direct

crates/cargo-wdk/src/main.rs
```diff
@@ -21,21 +22,20 @@ mod test_utils;
...
-fn main() -> Result<()> {
+/// * [`ExitCode::SUCCESS`] on success,
+/// * [`ExitCode::FAILURE`] on error.
+fn main() -> ExitCode {
     let cli: Cli = Cli::parse();
     trace::init_tracing(cli.verbose);
-    cli.run().inspect_err(|e| error!("{}", e))?;
-    Ok(())
+    if let Err(e) = cli.run() {
+        error!("{e:#}");
```
svasista-ms, line 37: The "duplicate error log" issue reported by @gurry ([comment](https://github.com/microsoft/windows-drivers-rs/pull/649#issuecomment-4394198389)) was because both`inspect_err` closure and the `Err` returned from the `main` function (which uses the `debug` impl) printed the errors. After discussing the issue internally, as a quick fix, we decided to return an `ExitCode` from `main` instead of `std::result::Result` and handling all the error reporting prior to the return.

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -296,6 +296,76 @@ mod kmdf_driver_with_target_override {
...
 }
 
+#[test]
+fn kmdf_driver_builds_successfully_with_sign_mode_off() {
+    let driver = "kmdf-driver";
+    let project_path = format!("tests/{driver}");
+    with_mutex(&project_path, || {
+        run_cargo_clean(&project_path);
+
+        let stderr = run_build_cmd(&project_path, Some(&["--sign-mode", "off"]), None);
```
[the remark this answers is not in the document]
svasista-ms: Fixed

crates/cargo-wdk/src/actions/build/package_task.rs
svasista-ms: Moved `run_infverif` before the signing steps and factored out the signing and verification steps to `sign_and_verify` fn.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -28,10 +28,21 @@ use windows::{
...
 use crate::providers::{exec::CommandExec, fs::Fs, wdk_build::WdkBuild};
-use crate::{
-    actions::{SignMode, build::error::PackageTaskError},
-    providers::error::FileError,
-};
+use crate::{actions::build::error::PackageTaskError, providers::error::FileError};
+
+/// Signing configuration.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SignConfig {
```
gurry: Rename to `SignMode`

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -236,6 +250,20 @@ impl<'a> PackageTask<'a> {
...
+        self.sign_and_verify()?;
+        Ok(())
+    }
+
+    /// Signs the driver binary and catalog file according to `self.sign_mode`
+    /// and optionally verifies the resulting signatures. Returns a variant of
+    /// `PackageTaskError` if any step of the process fails.
+    fn sign_and_verify(&self) -> Result<(), PackageTaskError> {
+        let SignMode::Test { verify_signature } = self.sign_mode else {
+            info!("Sign mode is 'off'; skipping certificate generation, signing, and verification");
```
```suggestion
            info!("Sign mode is off; skipping signing");
```
Certificate generation is only an implementation detail and verification being skipped is implicit if signing is skipped. Therefore they don't need to be mentioned and we can keep it short. "off" without quotes is fine (and cleaner).

crates/cargo-wdk/src/actions/build/package_task.rs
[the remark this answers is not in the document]
gurry: @svasista-ms have you verified this?

crates/cargo-wdk/tests/build_command_test.rs
krishnakumar4a4: @svasista-ms Are you planning to include snapshot tests for the help output in the current PR, or will you open a separate PR?

svasista-ms: I'll open a separate PR for the help tests along with tests for other subcommands.

crates/cargo-wdk/README.md
```diff
@@ -90,9 +91,14 @@ If you have a workspace with a mix of sample and non-sample driver projects, the
 
 #### Signing and Verification
 
-To sign driver artifacts `build` looks for a certificate called `WDRLocalTestCert` in a store called `WDRTestCertStore`. Make sure you place your signing certificate there with that name. If no certificate is found, `build` will automatically generate a new self-signed one and add it for you.
+The `--sign-mode` flag controls how driver artifacts are signed. It accepts the following values:
```
```suggestion
The `build` command has a `--sign-mode` flag that controls how driver artifacts are signed. It accepts the following values:
```

crates/cargo-wdk/README.md
```diff
@@ -90,9 +91,14 @@ If you have a workspace with a mix of sample and non-sample driver projects, the
...
 
-If the `--verify-signature` flag is provided, the signatures are verified after signing. For verification to work, make sure you add a copy of the signing certificate in the `Trusted Root Certification Authorities` store. For security reasons `build` does not automatically do this even when it automatically generates the cert. You will have to always perform this step manually. 
+- `test` (default): Sign with an auto-generated self-signed certificate. `build` looks for a certificate called `WDRLocalTestCert` in a store called `WDRTestCertStore`. Make sure you place your signing certificate there with that name. If no certificate is found, `build` will automatically generate a new self-signed one and add it for you.
```
```suggestion
- `test` (default): Sign with a test certificate. The command looks for a certificate called `WDRLocalTestCert` in a store called `WDRTestCertStore`. If you wish you to use your certificate add it to the same store with the same name. Otherwise a self-signed certifcate will be automatically generated, added and used for signing.  
```

We can't say "sign with an auto-generated self-signed certificate" while also saying "make sure you place your own cert in the store". The more accurate description is "sign with a test certificate". That's all. It may or may not be auto generated.

I have also suggested changes in the rest of the paragraph for clarity.

crates/cargo-wdk/README.md
```diff
@@ -90,9 +91,14 @@ If you have a workspace with a mix of sample and non-sample driver projects, the
...
+- `off`: Skip signing entirely. No signing or signature verification steps are performed. This is useful when you intend to sign the driver later with your own toolchain (for example, HLK certification). Note that an unsigned driver cannot be installed on Windows without first enabling [test signing](https://learn.microsoft.com/en-us/windows-hardware/drivers/install/the-testsigning-boot-configuration-option) or otherwise loading it through a signed mechanism.
```
```suggestion
- `off`: Skip signing entirely. No signing or signature verification is performed. This is useful when you intend to sign the driver later with your own toolchain (e.g. HLK certification).
```

Developers know this already and need not be repeated here:

_Note that an unsigned driver cannot be installed on Windows without first enabling [test signing](https://learn.microsoft.com/en-us/windows-hardware/drivers/install/the-testsigning-boot-configuration-option) or otherwise loading it through a signed mechanism._

Replace "steps are" with "is" and "for example" with "e.g." for brevity.

crates/cargo-wdk/README.md
```diff
@@ -90,9 +91,14 @@ If you have a workspace with a mix of sample and non-sample driver projects, the
...
+
+If the `--verify-signature` flag is provided, the signatures are verified after signing. For verification to work, make sure you add a copy of the signing certificate in the `Trusted Root Certification Authorities` store. For security reasons `build` does not automatically do this even when it automatically generates the cert. You will have to always perform this step manually.
+
+`--verify-signature` cannot be combined with `--sign-mode=off` because there is no signature to verify. Passing both will cause `build` to fail with an error.
```
```suggestion
`--verify-signature` cannot be combined with `--sign-mode=off` because if signing is off there is nothing to verify. Passing both will cause `build` to fail with an error.
```

Makes the sentence clearer

crates/cargo-wdk/README.md
```diff
@@ -113,3 +119,9 @@ If the `--verify-signature` flag is provided, the signatures are verified after
     ```pwsh
     cargo wdk build --target-arch amd64
     ```
+
+- To build a driver project and skip signing (for example, to sign later with a production certificate), run:
```
```suggestion
- To build a driver project with test signing, navigate to the root of the project and run:
```
Changed "and skip signing" to "with test signing" as we should use the testing signing use-case as the example (see below suggestion the command line as well).

Secondly, remove the stuff in parenthesis as it is common knowledge and doesn't add much. Change "run" to "navigate to the root of the project and run" for consistency with other examples here.

crates/cargo-wdk/src/actions/build/package_task.rs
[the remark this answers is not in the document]
gurry: We agreed to fix this issue by deleting the package folder every time we begin the packaging phase

crates/cargo-wdk/README.md
```diff
@@ -113,3 +119,9 @@ If the `--verify-signature` flag is provided, the signatures are verified after
...
+
+    ```pwsh
+    cargo wdk build --sign-mode off
```
```suggestion
    cargo wdk build --sign-mode test
```

Test signing is much more common than disabling off signing so we should use that as the example

svasista-ms: Fixed ✅

crates/cargo-wdk/README.md
svasista-ms: Fixed ✅

crates/cargo-wdk/README.md
svasista-ms: Fixed ✅

crates/cargo-wdk/README.md
svasista-ms: Fixed ✅

crates/cargo-wdk/README.md
svasista-ms: Fixed ✅

crates/cargo-wdk/README.md
svasista-ms: Fixed ✅

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -236,24 +248,33 @@ impl<'a> PackageTask<'a> {
...
-        // Verify signatures only when --verify-signature flag = true is passed
-        if self.verify_signature {
-            info!("Verifying signatures for driver binary and cat file using signtool");
-            self.run_signtool_verify(&self.dest_driver_binary_path)?;
-            self.run_signtool_verify(&self.dest_cat_file_path)?;
+        match self.sign_mode {
+            SignMode::Test => {
+                self.generate_certificate()?;
+                self.copy(&self.src_cert_file_path, &self.dest_cert_file_path)?;
+                self.run_signtool_sign(
```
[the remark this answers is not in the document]
svasista-ms: `run_infverif` now runs before `sinn_and_verify`

crates/cargo-wdk/src/actions/build/tests.rs
```diff
@@ -265,6 +266,118 @@ pub fn given_a_driver_project_when_verify_signature_is_true_then_it_builds_succe
...
+ {
+    // Input CLI args
+    let cwd = PathBuf::from("C:\\tmp");
+    let profile = None;
+    let target_arch = CpuArchitecture::Amd64;
+    let verify_signature = true;
```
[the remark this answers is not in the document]
svasista-ms: This is refactored. `verify_signature` is now part of `SignMode::Test`.

crates/cargo-wdk/src/actions/mod.rs
svasista-ms: Fixed ✅

crates/cargo-wdk/src/actions/build/package_task.rs
svasista-ms: Fixed ✅

crates/cargo-wdk/src/actions/build/package_task.rs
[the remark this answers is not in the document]
gurry: Created an issue for this to fix later: https://github.com/microsoft/windows-drivers-rs/issues/660

crates/cargo-wdk/src/cli.rs
```diff
@@ -254,4 +288,29 @@ mod tests {
...
+                sample: false,
+            }),
+            verbose: clap_verbosity_flag::Verbosity::default(),
+        };
+
+        let result = cli.run();
+        assert!(result.is_err());
+        let err = result.err().unwrap().to_string();
+        assert!(
+            err.contains("--verify-signature") && err.contains("--sign-mode=off"),
```
gurry: Intead of these `err.contains` checks we can check against the full text ` "--verify-signature cannot be used with --sign-mode=off."`, can't we?

crates/cargo-wdk/src/main.rs
```diff
@@ -21,21 +22,20 @@ mod test_utils;
 
 /// Main function for the [`cargo-wdk`][crate] CLI application.
 ///
-/// The main function parses the CLI input, sets up tracing and executes the
-/// command. If an error occurs during execution, it logs the error and exits
-/// with a non-zero status code.
+/// Parses the CLI input, initializes tracing, and runs the chosen subcommand.
+/// On failure, the error is logged through `tracing::error!`, and the process
+/// exits with a non-zero status.
```
gurry: Why have we changed this comment?

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -296,6 +296,76 @@ mod kmdf_driver_with_target_override {
...
+        assert!(
+            !cert_in_package.exists(),
+            "Expected no cert file in package folder when --sign-mode=off, but found {}",
+            cert_in_package.display()
+        );
+    });
+}
+
+#[test]
+fn sign_mode_off_does_not_produce_cert_file_in_target_dir() {
```
gurry: Can't we just move the checks we're doing here into the test `kmdf_driver_builds_successfully_with_sign_mode_off` and delete this test?

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -296,6 +296,76 @@ mod kmdf_driver_with_target_override {
...
+    let project_path = format!("tests/{driver}");
+    with_mutex(&project_path, || {
+        run_clean_cmd(&project_path);
+
+        let stderr = run_build_cmd(&project_path, Some(&["--sign-mode", "off"]), None);
+        assert!(stderr.contains(&format!("Building package {driver}")));
+        assert!(stderr.contains(&format!("Finished building {driver}")));
+
+        let driver_name = driver.replace('-', "_");
+        let target_dir = format!("{project_path}/target/debug");
```
gurry, line 311: Is it safe to assume the output dir will always be `debug` and not `release`?

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -296,6 +296,76 @@ mod kmdf_driver_with_target_override {
...
+    let mut cmd = create_cargo_wdk_cmd(
+        "build",
+        Some(&["--sign-mode", "off", "--verify-signature"]),
+        None,
+        Some(&project_path),
+    );
+    let assertion = cmd.assert().failure();
+    let stderr = String::from_utf8_lossy(&assertion.get_output().stderr).to_string();
+    assert!(
+        stderr.contains("--verify-signature") && stderr.contains("--sign-mode=off"),
```
gurry: Again, can we not check for the full error string instead of contains checks?

crates/cargo-wdk/README.md
```diff
@@ -65,16 +65,17 @@ Usage: cargo wdk build [OPTIONS]
 Options:
       --profile <PROFILE>          Build artifacts with the specified profile
       --target-arch <TARGET_ARCH>  Build for the target architecture
+      --sign-mode <SIGN_MODE>      Driver signing mode [default: test] [possible values: off, test]
       --verify-signature           Verify the signature
       --sample                     Build sample class driver project
-  -h, --help                       Print help
+  -h, --help                       Print help (see more with '--help')
```
gurry: Can you remove the `(see more with '--help')` part? It is a bit confusing in the readme.

crates/cargo-wdk/src/main.rs
svasista-ms: I accepted this Copilot suggestion when we changed the `main` to return `ExitCode`. I feel the original comment reads better. I will revert this. Thank you!

crates/cargo-wdk/tests/build_command_test.rs
svasista-ms: Yes, in this test, `--profile release` is not passed to `run_build_cmd`, so it is safe to hardcode the `target_dir`

crates/cargo-wdk/src/cli.rs
svasista-ms: Changed the `assert!` to verify the complete string ✅

crates/cargo-wdk/src/main.rs
svasista-ms: Fixed ✅

crates/cargo-wdk/tests/build_command_test.rs
svasista-ms: Thanks @gurry, I moved this check inside the `kmdf_driver_builds_successfully_with_sign_mode_off` itself.

crates/cargo-wdk/tests/build_command_test.rs
svasista-ms: Fixed ✅

crates/cargo-wdk/README.md
svasista-ms: Fixed ✅

crates/cargo-wdk/README.md
```diff
@@ -113,3 +119,9 @@ If the `--verify-signature` flag is provided, the signatures are verified after
     ```pwsh
     cargo wdk build --target-arch amd64
     ```
+
+- To build a driver project with test signing, navigate to the root of the project and run:
+
+    ```pwsh
+    cargo wdk build --sign-mode test
```
krishnakumar4a4: nit: Since the default is test signing on. Its better to put the counter example here

```
- To build a driver project with signing off, navigate to the root of the project and run:

    ```pwsh
    cargo wdk build --sign-mode off
    ```
```

krishnakumar4a4 (dismissed):

gurry (dismissed):

crates/cargo-wdk/README.md
gurry: That's a good point. I think I had asked @svasista-ms to change it from `off` to `test` earlier. @svasista-ms apologies, but I think it's better to go with the `off` example.

gurry (approved):

krishnakumar4a4 (approved):

crates/cargo-wdk/tests/build_command_test.rs
gurry: Issue created to track this: https://github.com/microsoft/windows-drivers-rs/issues/663

Issue #516, opened 2025-09-23
Implement mutually exclusive header support in bindgen invocations (constants, types, API-subsets)


wmmc88:
Update all bindgen invocations for constants, types, and API-subset-specific modules to leverage the mutually exclusive header grouping mechanism. Specifically:

- Bindgen runs should first process base types, then run separately on each mutually exclusive header, always including the base headers but only one mutually exclusive header per run.
- Use bindgen's `allowlist_file` and set `allowlist_recursively` to false to prevent duplicate definitions and allow correct parsing.
- Ensure this logic is reusable for all relevant code paths and easily maintainable as WDK header compatibility changes over time.
- Implementation should minimize changes outside of grouping logic and maximize reusability across the codebase.

Reference: See parent issue for detailed background and rationale.

PR #652, opened 2026-04-29
feat: `wdk_build::TraitsMap` for collection of implemented traits for bindgen output

leon-xd:
This PR introduces `wdk_build::TraitsMap`, a struct to collect the traits each `bindgen` emitted type implements. See #654 for the related PR.

### Purpose
These changes are a pre-requisite to solve the current issues with generating bindings for mutually exclusive WDK headers (tracked by #516 ). The solution requires types and constants for subsystems to be generated in independent `bindgen` runs, and will be published in a follow-up PR. In order to prevent duplicate type generation, `bindgen` invocation for subsystem types will utilize a `blocklist_file` for every file traversed in the initial base types run. 

`bindgen` by default assumes very minimal derived traits when encountering a blocklisted type, so this implementation on its own would leave most subsystem types with a sparse set of automatic trait implementations. This is a regression that is unacceptable. In order to prevent this while still keeping our mutual exclusive headers solution, we need to gather the trait implementation for all of the generated base types, and pass it off to subsystem `bindgen` type generators via the callback [`blocklisted_type_implements_trait`](https://docs.rs/bindgen/latest/bindgen/callbacks/trait.ParseCallbacks.html#method.blocklisted_type_implements_trait).

This PR introduces the mechanism to be able to parse a `bindgen` generated file using [`syn`](https://docs.rs/syn/latest/syn/) and extract the trait implementation for each individual type. In practice, the future bindings generation pipeline will generate the types for the `base` layer of the WDK, parse it via `TraitsMap::from_file`, and then insert the created `TraitsMap` into each subsystem type `bindgen` via `BaseTraitsCallback`.

### Changes
- `Cargo.toml`
  - Bump `bindgen` from `0.71.0` to `0.72.1`
- `crates/wdk-build/Cargo.toml`
  - Add `syn` dependency with the `parsing` feature enabled
- `crates/wdk-build/src/lib.rs`
  - Declare new `pub mod Traits`
- `crates/wdk-build/src/bindgen.rs`
  - Pin `size_t_is_usize(true)` on `wdk_default`.
- `crates/wdk-build/src/traits.rs`
  - `TraitsError`: typed error enum for failures.
  - `TraitsSet`: a struct of bools tracking the five traits bindgen tracks (`Copy`, `Clone`, `Debug`, `Default`, `PartialEq`).
  - `TraitsSource`: enum distinguishing whether a type's traits are immediately known (from a `#[derive()]` attribute or from a base type) or come from an aliased type.
  - `TraitsMap`: a map from type name to `TraitsSet`. Builds incrementally from Rust source files via `syn`, and implements the bindgen callback for derive implementation
  - `syn` helpers -- extract and parse functions to navigate `syn` types
  - Tests: unit-test coverage for parsing, alias resolution (including cycle detection and stable iteration order), classification of `#[derive(...)]` attributes, and `syn` helpers.
-  `crates/wdk-build/tests/traits.rs`
  - Integration tests testing full end to end flow

### Validation
Tested this with the full implementation of the mutually exclusive header resolution and ensured that all outputted bindings had full `#[derive]` parity against bindings generated from `microsoft/main`.


crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,738 @@
...
+                curr = next;
+            }
+
+            let target_derive_set = *self
+                .types
+                .get(curr)
+                .expect("`self.types.contains_key(curr)` just returned true");
+
+            for new_derive_key in walked {
+                self.types.insert(new_derive_key.clone(), target_derive_set);
```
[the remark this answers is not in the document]
leon-xd: `.clone` on a `&String` calls on the String because of autoref

crates/wdk-build/src/bindgen.rs
```diff
@@ -138,6 +138,9 @@ impl BuilderExt for Builder {
             // Defaults enums to generate as a set of constants contained in a module (default value
             // is EnumVariation::Consts which generates enums as global constants)
             .default_enum_style(bindgen::EnumVariation::ModuleConsts)
+            // `size_t`/`ssize_t` are pointer-width on every supported Windows
```
wmmc88, line 145: This should not be needed since afaiu this is true by default

crates/wdk-build/Cargo.toml
```diff
@@ -32,6 +33,7 @@ rustversion.workspace = true
 semver.workspace = true
 serde = { features = ["derive"], workspace = true }
 serde_json.workspace = true
+syn = { features = ["parsing"], workspace = true }
```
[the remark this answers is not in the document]
wmmc88: @leon-xd double check this. if its true, then the suggestion is more correct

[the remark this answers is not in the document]
leon-xd: added

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,719 @@
...
+    UnhandledSynCase { node: String },
+
+    #[error("malformed shape: {reason}: {node}")]
+    MalformedShape { reason: String, node: String },
+
+    #[error("alias cycle among: {names:?}")]
+    AliasCycle { names: Vec<String> },
+
+    #[error("alias targets not found: {names:?}")]
+    UnresolvedAlias { names: Vec<String> },
```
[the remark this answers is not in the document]
leon-xd: we are pulling this in via `#[doc(hidden)]` so this doesn't fail the build but more comments is good so i added them

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,719 @@
...
+#[derive(Debug)]
+pub struct BaseDerivesCallback {
+    map: Arc<DerivesMap>,
+}
+
+impl BaseDerivesCallback {
+    #[must_use]
+    pub const fn new(map: Arc<DerivesMap>) -> Self {
+        Self { map }
+    }
```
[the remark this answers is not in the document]
leon-xd: we are pulling this in via #[doc(hidden)] so this doesn't fail the build but more comments is good so i added them

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,955 @@
+// Copyright (c) Microsoft Corporation
+// License: MIT OR Apache-2.0
+
+//! Parses bindgen-emitted Rust source to recover the set of derives bindgen
+//! applied to each generated type. Used by the per-subsystem bindgen pipeline
+//! to answer `blocklisted_type_implements_trait` for base types.
```
wmmc88: answer?

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,955 @@
...
+/// X = Y;` target.
+const PRIMITIVES: &[&str] = &[
+    "bool", "char", "f32", "f64", "i8", "i16", "i32", "i64", "i128", "isize", "u8", "u16", "u32",
+    "u64", "u128", "usize",
+];
+
+/// C stdint names that bindgen lowers to Rust integer primitives internally.
+/// Bindgen never emits these as `pub type` aliases, so they have to be seeded
+/// into the map directly. Mirrors bindgen 0.72.1's `is_stdint_type` allowlist —
+/// re-verify on bindgen upgrades.
```
wmmc88: a comment like this isn't really enforceable imo. at least provide a permalink to where its defined.

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,955 @@
...
+    path::{Path as FsPath, PathBuf},
+    sync::Arc,
+};
+
+use bindgen::callbacks::{DeriveTrait, ImplementsTrait, ParseCallbacks};
+use syn::{Attribute, Item, ItemUse, Path, PathArguments, Type, UseTree};
+use thiserror::Error;
+
+/// Rust language primitives that can appear as a bare identifier in a `pub type
+/// X = Y;` target.
```
wmmc88: comment unclear. is this for the Y? also the type itself is supposed to be "an exhaustive list of all rust primitives". imo its usage isnt relevant here. document that at the callsite etc

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,955 @@
...
+const PRIMITIVES: &[&str] = &[
+    "bool", "char", "f32", "f64", "i8", "i16", "i32", "i64", "i128", "isize", "u8", "u16", "u32",
+    "u64", "u128", "usize",
```
wmmc88: nit: may want to add nightly flagged support for f16/f128: https://github.com/rust-lang/rust/issues/116909. could even add a test that will start to fail once those types are stabilized so we know to update this

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,955 @@
...
+                curr = next;
+            }
+
+            let target_derive_set = *self
+                .types
+                .get(curr)
+                .expect("`self.types.contains_key(curr)` just returned true");
+
+            for new_derive_key in walked {
+                self.types.insert(new_derive_key.clone(), target_derive_set);
```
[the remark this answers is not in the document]
leon-xd: i tried to write this algorithm with the least amount of clones and owned strings as possible, this compiles and runs so going to close this

crates/wdk-build/tests/derives.rs
```diff
@@ -0,0 +1,146 @@
+// Copyright (c) Microsoft Corporation
+// License: MIT OR Apache-2.0
+
+//! Integration tests for [`wdk_build::derives::DerivesMap`] driven through
+//! [`DerivesMap::from_file`]: writes a representative bindgen source snippet
```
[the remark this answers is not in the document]
leon-xd: : is plain text, this parses fine. iirc cargo doc won't even render this because it's an integration test file

gurry:
@leon-xd can parsing base types have any measurable impact on build time?

wmmc88:
> @leon-xd can parsing base types have any measurable impact on build time?

@gurry in the pr description of [Leon's other pr which uses this new feature](https://github.com/microsoft/windows-drivers-rs/pull/654), he says:
> I also timed the build process with this change to ensure there was not a significant build time regression. As it stands I am currently able to build wdk_sys with all_features in ~45 seconds, compared to microsoft/main's 60 seconds. I would not trust this number as the build script is subject to change with future revisions for this PR.

So while technically its not attributable to only this change, as a whole the change reduces build time. I think a theory was that bindgen's parsing of header seems to scale non-linearly

crates/wdk-build/src/lib.rs
```diff
@@ -28,6 +28,8 @@ pub mod metadata;
 mod utils;
 
 mod bindgen;
+#[doc(hidden)]
```
wmmc88: why is this hidden?

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,955 @@
...
+    fn from(derives: Vec<String>) -> Self {
+        let mut set = Self::empty();
+        for derive in &derives {
+            set |= match derive.as_str() {
+                "Copy" => Self::COPY,
+                "Debug" => Self::DEBUG,
+                "Default" => Self::DEFAULT,
+                "Hash" => Self::HASH,
+                "PartialEq" | "PartialOrd" => Self::PARTIAL_EQ_OR_PARTIAL_ORD,
+                _ => Self::empty(),
```
wmmc88: is dropping unknown derives the right call here?

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,955 @@
...
+    #[error("alias cycle among: {names:?}")]
+    AliasCycle {
+        /// Names participating in the detected cycle, in walk order.
+        names: Vec<String>,
+    },
+
+    /// Alias chain terminated at a name that is neither a recorded type nor
+    /// another pending alias.
+    #[error("alias target not found: {target}")]
+    UnresolvedAlias {
```
wmmc88: nit: to be precise, i think all `Alias` should be `TypeAlias`

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,955 @@
...
+        /// The unresolved target name.
+        target: String,
+    },
+}
+
+bitflags::bitflags! {
```
wmmc88: not sure i would pull in bitflags here. we're never serializing it and never need a repr(c) version of it. its only a handful of flags too so my gut here is not to preemptively optimize. unless you have some bench showing this is much faster?

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,955 @@
...
+    },
+
+    /// `syn` failed to parse the source as Rust.
+    #[error("failed to parse source as Rust")]
+    Parse(#[source] syn::Error),
+
+    /// Encountered a top-level [`syn::Item`] variant this parser does not
+    /// handle.
+    #[error("unhandled syn node: {node}")]
+    UnhandledSynCase {
```
gurry:
```suggestion
    UnsupportedSynItem {
```

This is clearer IMO

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,955 @@
...
+    /// Returns whether `name`'s recorded derive set contains `derive_trait`.
+    /// Returns `false` if `name` is not recorded.
+    #[must_use]
+    pub fn satisfies(&self, name: &str, derive_trait: DeriveTrait) -> bool {
+        self.types
+            .get(name)
+            .is_some_and(|&set| set.implements(derive_trait))
+    }
+
+    /// Parses a Rust source file and records the derive set for every
```
gurry:
```suggestion
    /// Parses given Rust source file content and records the derive set for every
```

I feel this is clearer since the method does not operate on physical files

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,955 @@
...
+/// - [`DerivesError::UnhandledSynCase`] if `ty` is a `syn::Type` variant other
+///   than Ptr/Path/Array
+/// - [`DerivesError::MalformedShape`] if the path has no segments
+/// - [`DerivesError::UnhandledSynCase`] if the path has generic arguments
+fn derives_for_type(ty: &Type) -> Result<DerivesSource, DerivesError> {
+    match ty {
+        Type::Ptr(_) => Ok(DerivesSource::Direct(DerivesSet::all())),
+        Type::Array(arr) => derives_for_type(&arr.elem),
+        Type::Path(tp) => {
+            if path_is_option(&tp.path) && inner_is_bare_fn(&tp.path) {
```
gurry: Might wanna add a comment saying that this check is for C callbacks or simply have a function called `is_callback_type()`

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,955 @@
...
+                return Err(DerivesError::UnhandledSynCase {
+                    node: format!("{:?}", last.arguments),
+                });
+            };
+
+            if PRIMITIVES.iter().any(|&p| last.ident == p) || path_is_core_ffi_type(&tp.path) {
+                return Ok(DerivesSource::Direct(DerivesSet::all()));
+            }
+
+            Ok(DerivesSource::Alias(last.ident.to_string()))
```
gurry: We are using only the last segment as key here. Can bindgen output ever have two distinct types with the same last segment and they are both used as alias RHSes?

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,955 @@
...
+    pub fn satisfies(&self, name: &str, derive_trait: DeriveTrait) -> bool {
+        self.types
+            .get(name)
+            .is_some_and(|&set| set.implements(derive_trait))
+    }
+
+    /// Parses a Rust source file and records the derive set for every
+    /// top-level `struct`, `union`, `enum`, and type alias. Unknown derive
+    /// idents are ignored.
+    ///
```
gurry: May be worth calling out in the comment that this method does not work for bindgen output containing function declarations (i.e. output generated with `CodegenConfig::FUNCTIONS`)

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,955 @@
...
+    let segs = &path.segments;
+    segs.len() >= 3 && segs[segs.len() - 3].ident == "core" && segs[segs.len() - 2].ident == "ffi"
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[track_caller]
+    fn assert_direct_full(source: DerivesSource) {
```
gurry:
```suggestion
    fn assert_direct_all(source: DerivesSource) {
```

More accurate name since we assert `DeriveSet::all()`

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,955 @@
...
+            let ty: Type = parse_str("*mut u32").unwrap();
+            assert_direct_full(derives_for_type(&ty).unwrap());
+            let ty: Type = parse_str("*const ::core::ffi::c_void").unwrap();
+            assert_direct_full(derives_for_type(&ty).unwrap());
+        }
+
+        #[test]
+        fn derives_for_type_array_recurses_into_element() {
+            let ty: Type = parse_str("[u32; 4]").unwrap();
+            assert_direct_full(derives_for_type(&ty).unwrap());
```
gurry: Perhaps add an assert for arrays element types that do not implement all traits.

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,955 @@
...
+        #[test]
+        fn unsupported_item_kind_surfaces_unhandled_syn_case() {
+            // Item::Trait is not part of the supported Struct/Union/Enum/Type/Mod/
+            // Use/Impl/Const set, so the catch-all arm fires.
+            assert!(matches!(
+                DerivesMap::from_source("pub trait T {}"),
+                Err(DerivesError::UnhandledSynCase { .. })
+            ));
+        }
+    }
```
gurry: Add a test verifying that foreign mods like this are rejected with error:

```rust
extern "C" {
    pub fn some_ffi_func(...);
    pub static mut some_const: SOME_TYPE;
}
```

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,955 @@
...
+
+            let ty: Type = parse_str("[SomeAlias; 8]").unwrap();
+            assert_alias(derives_for_type(&ty).unwrap(), "SomeAlias");
+        }
+
+        #[test]
+        fn derives_for_type_primitive_path_gets_all() {
+            let ty: Type = parse_str("u32").unwrap();
+            assert_direct_full(derives_for_type(&ty).unwrap());
+        }
```
gurry: Maybe we should check for all primitive types in the `PRIMITIVES` here just like we do for stdints in `stdint_names_all_derive_standard_set()`

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,955 @@
...
+                    );
+                }
+            }
+        }
+
+        /// A cyclic alias pair (`A = B; B = A;`) must surface as `AliasCycle` —
+        /// the chain-walking loop detects it when a step revisits a name
+        /// already in the walked set.
+        #[test]
+        fn alias_cycle_terminates() {
```
gurry: Will be useful to test for a cycle with three types. Tests for two type cycles may be too narrow to exercise the full logic.

crates/wdk-build/tests/traits.rs
```diff
@@ -0,0 +1,146 @@
...
+fn from_file_invalid_rust_returns_parse_error() {
+    let tmp = NamedTempFile::new("bad.rs").expect("create temp file");
+    tmp.write_str("not @ valid @ rust @@@")
+        .expect("write temp file");
+    let err = DerivesMap::from_file(tmp.path()).expect_err("invalid syntax must error");
+    assert!(
+        matches!(err, DerivesError::Parse(_)),
+        "expected Parse, got {err:?}"
+    );
+}
```
gurry, line 149: Not super important, but may want to also test for bindgen output containing foreign mods i.e. it gets rejected.

gurry (commented):
LGTM. Some minor suggestions

wmmc88 (changes requested):

crates/wdk-build/src/derives.rs
leon-xd: I changed it to `UnsupportedSynNode` since that is more clear -- `Item` is a `syn` type and this error has more uses than just unsupported `Item`s

crates/wdk-build/src/derives.rs
leon-xd: Fixed

crates/wdk-build/src/derives.rs
leon-xd: Added a comment

crates/wdk-build/src/derives.rs
leon-xd: I changed it to join the entire list with `::` as separators. But this is a corner case -- right now the RHS is only a single segmented type

crates/wdk-build/src/derives.rs
leon-xd: Added

crates/wdk-build/src/derives.rs
leon-xd: changed

crates/wdk-build/src/derives.rs
leon-xd: Added

crates/wdk-build/src/derives.rs
leon-xd: Added

crates/wdk-build/src/derives.rs
leon-xd: Added

crates/wdk-build/src/derives.rs
leon-xd: Added another type to test

crates/wdk-build/tests/traits.rs
leon-xd: Added

crates/wdk-build/src/bindgen.rs
leon-xd: It is true by default but I am pinning it because in case that default value changes -- is it ok to keep?

crates/wdk-build/src/derives.rs
leon-xd: Changed to implements

crates/wdk-build/src/derives.rs
leon-xd: Added permalink, agree about unenforcable comment but don't knwo a better way to go about this

crates/wdk-build/src/derives.rs
leon-xd: Fixed comment

crates/wdk-build/src/derives.rs
leon-xd: Added f16 and f128 ahead of time, didn't add test -- should we do that?

crates/wdk-build/src/lib.rs
leon-xd: It's hidden because this is functionality specifically for `wdk-sys`. This is not intended to be a public feature but given that it's only for the `wdk-sys` build script I figured that keeping it out of the docs was enough to keep this from being unused by doc. iirc i looked at how other examples of this was handled in the repo -- should I handle this differently?

crates/wdk-build/src/derives.rs
leon-xd: Dropping the unknown derives is just indicative that the bindgen output has a feature that doesn't need to be tracked via callbacks. The set of derives that need to be tracked is smaller than the set of derives that could be output (as an example, some bindgen output has `Clone` but bindgen does not need us to pass that an opaque type derives that trait).

crates/wdk-build/src/derives.rs
leon-xd: Changed the wording throughout

crates/wdk-build/src/derives.rs
leon-xd: fixed based on our offline discussion. now muting the struct of bools lint.

crates/wdk-build/src/bindgen.rs
wmmc88: okay

crates/wdk-build/src/derives.rs
wmmc88: its probably fine. we can address this later if we find issues with this

crates/wdk-build/src/lib.rs
wmmc88: if its part of our public api, it should just be public. It's already a public module, and also intended to be used by external crates. The only time where doc(hidden) is really typically used is for providing functions to proc-macros, where there is a language limitation on keeping something like that contained

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,1206 @@
+// Copyright (c) Microsoft Corporation
```
wmmc88: This file has multiple references to a "parser". To be clear, this is just referring to `DerivesMap::from_file` (since there's no actual parser object or module)?

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,1206 @@
...
+/// Map storing Rust source type names to the set of derives the type
+/// implements.
+#[derive(Debug)]
+pub struct DerivesMap {
+    types: HashMap<String, DerivesSet>,
+}
+
+impl DerivesMap {
+    /// Reads a Rust source file from disk and parses its derive
+    /// information. See [`DerivesMap::from_source`] for the parsing behavior.
```
wmmc88: `DerivesMap::from_source` is a private method, so I would assume this rustdoc linking should fail. Is it passing?

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,1206 @@
...
+            path: path.to_path_buf(),
+            source,
+        })?;
+        Self::from_source(&source)
+    }
+
+    /// Returns whether `name`'s recorded derive set contains `derive_trait`.
+    /// Returns `false` if `name` is not recorded.
+    #[must_use]
+    pub fn satisfies(&self, name: &str, derive_trait: DeriveTrait) -> bool {
```
wmmc88: 1. Does this actually need to be public? I only see this used internally and not in #654. I'm also trying to be mindful of new extra public api's were exposing (since they carry back-compat guarantees that can needlessly bump versions)

2. nit: naming is awkward. `map.satisfies('MyStruct', DeriveTrait::Copy)` makes it seem like the map satisfies 'MyStruct' or is otherwise confusing. I think this is mostly a smell that leads me to point 3.

3. Is there a reason to not just expose the underlying `types` `HashMap`? Or implement the same traits and accessor apis from `HashMap` on the `DerivesMap` type (ex. https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.get, https://doc.rust-lang.org/std/collections/struct.HashMap.html#impl-Index%3C%26Q%3E-for-HashMap%3CK,+V,+S,+A%3E)?

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,1206 @@
...
+            DeriveTrait::Default => self.default,
+            DeriveTrait::Hash => self.hash,
+            DeriveTrait::PartialEqOrPartialOrd => self.partial_eq_or_partial_ord,
+        }
+    }
+}
+
+/// Map a bindgen `#[derive(...)]` ident to the tracked [`DeriveTrait`] it
+/// represents, or `None` if the parser does not track it.
+fn derive_trait_from_name(name: &str) -> Option<DeriveTrait> {
```
wmmc88: is there any reason this isn't just a `TryFrom` on `DeriveTrait`?

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,1206 @@
...
+                }
+            }
+        }
+
+        derives_map.resolve_type_aliases(&type_aliases)?;
+
+        Ok(derives_map)
+    }
+
+    fn with_std_types() -> Self {
```
wmmc88: Constructors are typically public

crates/wdk-build/src/derives.rs
wmmc88: Another point on 3 is that `DerivesMap`'s api doesn't actually act as a map-like type right now- it just **contains** a map-like type but it doesn't have the APIs to act like a map itself

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,1206 @@
...
+/// Impl / Const: bindgen helper blocks and anonymous layout assertions.
+/// Neither contributes derive information; both are ignored.
+///
+/// # Errors
+///
+/// Returns:
+/// - [`DerivesError::UnsupportedSynNode`] for `Item` variants other than
+///   Struct/Union/Enum/Type/Mod/Use/Impl/Const
+/// - any error propagated from the per-shape classifiers
+fn idents_and_derives_for_items(
```
wmmc88:
```suggestion
fn extract_idents_and_derives(
```

nit: for free functions, it makes it a lot more clear to lead with a verb since there's no `self` to provide context

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,1206 @@
...
+    items: &[Item],
+) -> Result<Vec<(String, DerivesSource)>, DerivesError> {
```
wmmc88: It might be worth adding some verbose tracing calls to some of this parsing, so that if there is ever some user problem we need to debug, we can get enough parsing information to debug the issue without making source code changes

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,1206 @@
...
+                    node: format!("{other:?}"),
+                });
+            }
+        }
+    }
+    Ok(derives)
+}
+
+/// Collects the derive trait names from a `#[derive(...)]` attribute list.
+fn derives_from_attrs(attrs: &[Attribute]) -> Vec<String> {
```
wmmc88: https://github.com/microsoft/windows-drivers-rs/pull/652/changes#r3616537521

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,1206 @@
...
+                syn::punctuated::Punctuated::<syn::Path, syn::Token![,]>::parse_terminated,
+            )
+            .ok()
+        })
+        .flatten()
+        .filter_map(|path| {
+            path.segments
+                .into_iter()
+                .next_back()
+                .map(|seg| seg.ident.to_string())
```
wmmc88: Its worth calling out that this will break for custom derives or derives with ambiguous trailing path segments, but that it doesn't matter since it later gets discarded

crates/wdk-build/src/derives.rs
wmmc88: This applies to all the free functions you created

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,1206 @@
...
+                    });
+                }
+                walked.push(next);
+                curr = next;
+            }
+
+            let target_derive_set = *self
+                .types
+                .get(curr)
+                .expect("`self.types.contains_key(curr)` just returned true");
```
wmmc88: https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,1206 @@
...
+/// pub type WCHAR = u16;                                    // Type::Path (primitive)
+/// ```
+///
+/// # Errors
+///
+/// Returns:
+/// - [`DerivesError::UnsupportedSynNode`] if `ty` is a `syn::Type` variant
+///   other than Ptr/Path/Array, or if the path has generic arguments
+/// - [`DerivesError::MalformedShape`] if the path has no segments
+fn derives_for_type(ty: &Type) -> Result<DerivesSource, DerivesError> {
```
wmmc88: This function has a bunch of implementation details whose intent should be more explicitly documented (i.e. why each set of derives is being added for each case)

crates/wdk-build/src/derives.rs
```diff
@@ -0,0 +1,1206 @@
...
+}
+
+/// True when `path` ends in `core::ffi::*`.
+fn path_is_core_ffi_type(path: &Path) -> bool {
+    let segs = &path.segments;
+    segs.len() >= 3 && segs[segs.len() - 3].ident == "core" && segs[segs.len() - 2].ident == "ffi"
+}
+
+#[cfg(test)]
+mod tests {
```
wmmc88: please organize tests so that each function being tested is a sub module, and each test is a named scenario expectation. 

toy example:

```rust
fn add(a: u32, b: u32) -> u32 { ... }

#[cfg(test)]
mod tests {
    mod add {
        #[test]
        fn two_negative_values_return_correct_sum() { ... }
    }
}
```

This makes it clear for when the test harness prints out test names with full paths. For example, `tests::add::two_negative_values_return_correct_sum ... FAIL`

wmmc88 (changes requested):
Overall shape is good. Just some cleanup needed

crates/wdk-build/src/lib.rs
leon-xd: changed this to be a public mod

crates/wdk-build/src/derives.rs
leon-xd: Removed language of "parser" -- that was from an old revision.

crates/wdk-build/src/derives.rs
leon-xd: Based method implementation on `std::collections::HashSet` including implementing the `Index` trait

crates/wdk-build/src/derives.rs
leon-xd: Orphan rule. But this did end up being a bad smell -- construction is now a TryFrom with two different impls: a vec of `String`s and a single `String`

crates/wdk-build/src/derives.rs
leon-xd: This was only used internally and we want this to not be an access point -- in my ideal world the only way to construct a `TraitMap` would be through pointing it at a Rust file (e.g. `from_file`, which is currently `pub`). But because it was only used once I just removed it and moved the logic to the callsite (TraitMap::from_source)

crates/wdk-build/src/derives.rs
leon-xd: changed wording of functions -- anything that should have output i have prefixed with `extract` and returns `Result`. anything that is possible to contain output I have prefixed with `parse` and returns `Option`

crates/wdk-build/src/derives.rs
leon-xd: added tracing throughout

crates/wdk-build/src/derives.rs
leon-xd: organized all tests

crates/wdk-build/src/traits.rs
```diff
@@ -0,0 +1,2083 @@
...
+fn extract_traits_from_type(ty: &Type) -> Result<TraitsSource, TraitsError> {
+    match ty {
+        Type::Ptr(_) => {
+            trace!("Type recognized as a `Ptr`");
+            Ok(TraitsSource::Direct(TraitsSet::all()))
+        }
+        Type::Array(arr) => {
+            trace!("Type recognized as a `Array`");
+            extract_traits_from_type(&arr.elem)
+        }
```
Alan632, line 690: Not sure how realistic this situation would be but would this allow `TraitsSet::default = true` for array definitions with length > 32?

The docs mention that trait implementations are generated up to size 32. https://doc.rust-lang.org/std/primitive.array.html

And it looks like @wmmc88 ran into issues b/c of it: https://github.com/rust-lang/rust-bindgen/issues/2803

Alan632 (commented):
Great changes and documentation!

crates/wdk-build/src/traits.rs
```diff
@@ -0,0 +1,2192 @@
...
+    "c_long",
+    "c_longlong",
+    "c_ptrdiff_t",
+    "c_schar",
+    "c_short",
+    "c_size_t",
+    "c_ssize_t",
+    "uchar",
+    "c_uint",
+    "c_ulong",
```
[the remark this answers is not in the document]
leon-xd: fixed

crates/wdk-build/src/derives.rs
leon-xd: Added

crates/wdk-build/src/derives.rs
leon-xd: fixed expect messages to be recommended style

crates/wdk-build/src/derives.rs
leon-xd: this got changed/outdated -- specify a bit more that this is only collecting tracked trait names. before this was returning a string I believe but we are now doing the conversion into TraitsSet in this method which makes it a bit clearer overall

leon-xd: (see current `extract_derived_traits_from_attrs` for details)

crates/wdk-build/src/derives.rs
leon-xd: fixed function names

crates/wdk-build/src/derives.rs
leon-xd: fixed rustdoc linking

crates/wdk-build/src/traits.rs
```diff
@@ -0,0 +1,2083 @@
...
+    ///
+    /// Returns:
+    /// - [`TraitsError::Parse`] if `source` is not valid Rust
+    /// - [`TraitsError::UnsupportedSynNodeVariant`] or
+    ///   [`TraitsError::UnsupportedNodeShape`] if a classified construct does
+    ///   not match any recognized bindgen output shape
+    /// - [`TraitsError::UnresolvedTypeAlias`] or
+    ///   [`TraitsError::TypeAliasCycle`] if a type alias cannot be resolved to
+    ///   a recorded type
+    #[tracing::instrument(level = "trace", err(level = Level::TRACE ))]
```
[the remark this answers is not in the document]
leon-xd: this is not correct

crates/wdk-build/src/traits.rs
leon-xd: Pushed a change where this is detected and taken care of for arrays where the element type is `TraitsSource::Direct`. Currently am throwing an error if the element type is an alias, this case is never hit in the current commented configurations located in `~/Cargo.toml` (tested locally). Left a comment for future resolution if we ever hit it.

crates/wdk-build/src/traits.rs
```diff
@@ -0,0 +1,2083 @@
...
+        name: &str,
+        derive_trait: DeriveTrait,
+    ) -> Option<ImplementsTrait> {
+        Some(if self.map[name].contains(derive_trait) {
+            ImplementsTrait::Yes
+        } else {
+            ImplementsTrait::No
+        })
+    }
+}
```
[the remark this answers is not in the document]
leon-xd: after testing this this was a good callout -- realized that we could avoid index shenanigans by removing it and having `TraitsMap` implement the `ParseCallbacks` trait directly. Pushed a change accordingly and changed the callback to return `None` if the key is not found in the map.

crates/wdk-build/src/traits.rs
```diff
@@ -0,0 +1,2295 @@
...
+        self.partial_eq_or_partial_ord |= other.partial_eq_or_partial_ord;
+    }
+}
+
+impl TryFrom<Vec<String>> for TraitsSet {
+    type Error = TraitsError;
+
+    /// Builds a `TraitsSet` from a list of implemented traits.
+    ///
+    /// Returns [`TraitsError::UntrackedTraits`] if all traits are not tracked.
```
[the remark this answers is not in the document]
leon-xd: reworded

crates/wdk-build/src/traits.rs
```diff
@@ -0,0 +1,2295 @@
...
+                    TraitsSource::Direct(set) => assert_eq!(set, expected, "{name}"),
+                    TraitsSource::TypeAlias(t) => {
+                        panic!("{name}: expected Direct, got TypeAlias({t:?})")
+                    }
+                }
+            }
+        }
+
+        #[test]
+        fn str_implements_all_except_copy_and_default() {
```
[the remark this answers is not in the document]
leon-xd: renamed

Issue #631, opened 2026-03-13
wdk-build: The `signtool-sign` build process requires network connection for a little value


tandasat:
The `signtool-sign` build process executes `signtool` with the `/t` flag. This requires network connection every time, which is inconvenient for offline work. 

https://github.com/microsoft/windows-drivers-rs/blob/547b792ba962065fd13795135352c5b752d9d0a5/crates/wdk-build/rust-driver-makefile.toml#L475-L484

I propose to remove this flag. 

While specifying `/t` is not inherently a bad idea, the benefit of that with wdk-build is effectively none given signing with this project is done with merely `WDRLocalTestCert` and not meant for production releases. In other words, while adding a timestamp with `/t` keeps the signature valid even after the certificate has expired ([ref](https://learn.microsoft.com/en-us/windows-hardware/drivers/install/release-signing)), but practically no one would care about this situation, put aside that it has a quite long validity (15 years on my env). If they do care, they would be signing with their own certs and commands like requested in #588. So, overall, invoking `signtool` with `/t` only makes the use of this project difficult with no practical value.

svasista-ms:
This is a valuable change, especially for offline builds. This change has to be made in `cargo-wdk`. With `signtool-args`, the `build` command already supports supplying custom arguments to `signtool sign`. However, the current default test signing process timestamps for a very little value.

PR #656, opened 2026-05-12
feat: add `--locked` option to the `build` command

### LINKED ISSUE #648: Support --locked passthrough in cargo wdk
## Summary

`cargo wdk` should support passing `--locked` through to the underlying `cargo build` / `cargo check` invocations it spawns.

## Motivation

The `--locked` flag is critical for reproducible builds. It ensures that `Cargo.lock` is respected exactly as committed, preventing any dependency resolution changes at build time. This matters for:

- **CI/CD reproducibility** — builds must be deterministic; without `--locked`, a CI run can silently pick up a different dependency version than what was tested locally, leading to hard-to-diagnose regressions.
- **Security & supply-chain integrity** — `--locked` guarantees that only audited, committed dependency versions are used. Without it, a compromised or yanked crate version could be silently substituted.
- **Compliance with Cargo best practices** — the Rust ecosystem strongly recommends `--locked` for production and release builds. Driver development, where correctness is safety-critical, should be no exception.
- **Parity with `cargo make` workflows** — existing CI pipelines (e.g., `codeql.yml`) already pass `--locked` via `cargo make`. As `cargo wdk` replaces `cargo make`, it must support the same flags to avoid workflow regressions.

## Expected Behavior

When a user runs:

```sh
cargo wdk build --locked
```

The `--locked` flag is forwarded to the underlying `cargo build` invocation, so that `Cargo.lock` is enforced and the build fails if the lockfile is out of date.

## Current Behavior

`cargo wdk` does not recognize or forward `--locked`, so there is no way to enforce lockfile-pinned builds through the `cargo wdk` interface.

## Proposed Solution

Forward `--locked` (and ideally the related `--frozen` and `--offline` flags) from `cargo wdk` to the inner `cargo` commands it invokes. This is a passthrough — `cargo wdk` does not need to interpret the flag itself, only ensure it reaches `cargo build` / `cargo check`.

## Impact

**High** — Blocks adoption of `cargo wdk` in CI pipelines that require deterministic, lockfile-enforced builds. This is a prerequisite for fully deprecating `cargo make` in WDR workflows.

svasista-ms:
This PR adds support for forwarding Cargo's `--locked` option from `cargo wdk build` down to the underlying Cargo invocations to ensure the `Cargo.lock` file is not altered during the build process.

Resolves #648 

## Changes

1. Introduces a `--locked` CLI flag [[1]](https://github.com/microsoft/windows-drivers-rs/pull/656/changes#diff-4861850c7366ce9ab803db669bc522b1265ae68a766778ab50c3a9976fde66d9R110-R113)
2. Threads the flag through BuildAction/BuildTask and forwards flags to `cargo  metadata`, `cargo build` and `cargo rustc` invocations in the build action. [[2]](https://github.com/microsoft/windows-drivers-rs/pull/656/changes#diff-90872d1fa59d353a99dbb2817d79ca143d8ae9050fdfa81558537d88667a0e29R117-R119) [[3]](https://github.com/microsoft/windows-drivers-rs/pull/656/changes#diff-5f6257c5c36ab5e36f0f1343cff20b245fbb898d3ed044f81b2a0d1ef57c064aR322-R325) [[4]](https://github.com/microsoft/windows-drivers-rs/pull/656/changes#diff-5f6257c5c36ab5e36f0f1343cff20b245fbb898d3ed044f81b2a0d1ef57c064aR515-R519)
3. Adds a parameter to the `metadata` provider to accept other options to forward to the `cargo metadata` command. [[5]](https://github.com/microsoft/windows-drivers-rs/pull/656/changes#diff-c4798e678306e6e4faa4c2c590518de8c72989c06c1243709bc5b1d2b67935e2R47-R54)
4. Adds/updates unit tests and system level tests. [[6]](https://github.com/microsoft/windows-drivers-rs/pull/656/changes#diff-b30d5879afa7d114c914430fd892a49a824db8040ecc061332d4b3760f278e18R306) [[7]](https://github.com/microsoft/windows-drivers-rs/pull/656/changes#diff-7f39e8d21b60832dceaba74ceb186dce58e45d2c660c6ce0674d1be3181a18e9R68)

## Screenshots

`cargo wdk build --help`:

image

### DIFF
--- a/crates/cargo-wdk/README.md
+++ b/crates/cargo-wdk/README.md
@@ -68,19 +68,15 @@ Options:
       --sign-mode <SIGN_MODE>      Driver signing mode [default: test] [possible values: off, test]
       --verify-signature           Verify the signature
       --sample                     Build sample class driver project
+      --locked                     Assert that `Cargo.lock` will remain unchanged
   -h, --help                       Print help
-
-Manifest Options:
-      --locked   Require Cargo.lock is up to date
-      --offline  Run without accessing the network
-      --frozen   Require Cargo.lock and cache are up to date
 
 Verbosity:
   -v, --verbose...  Increase logging verbosity
   -q, --quiet...    Decrease logging verbosity
 ```
 
-`build` takes a number of inputs specifying build profile (`dev` or `release`), target architecture (`amd64` or `arm64`), the driver signing mode, a flag enabling signature verification and a flag indicating a sample driver along with verbosity flags. It also accepts cargo manifest flags `--locked`, `--offline` and `--frozen`, which are forwarded to `cargo`.
+`build` takes a number of inputs specifying build profile (`dev` or `release`), target architecture (`amd64` or `arm64`), the driver signing mode, a flag enabling signature verification and a flag indicating a sample driver, a flag enabling the `--locked` flag passthrough to `cargo`, along with verbosity flags.
 
 When the command completes the packaged driver artifacts are emitted at the path `target\<profile>\<project-name>-package`.
 
--- a/crates/cargo-wdk/README.md
+++ b/crates/cargo-wdk/README.md
@@ -80,7 +80,7 @@ Verbosity:
   -q, --quiet...    Decrease logging verbosity
 ```
 
-`build` takes a number of inputs specifying build profile (`dev` or `release`), target architecture (`amd64` or `arm64`), the driver signing mode, a flag enabling signature verification and a flag indicating a sample driver along with verbosity flags.
+`build` takes a number of inputs specifying build profile (`dev` or `release`), target architecture (`amd64` or `arm64`), the driver signing mode, a flag enabling signature verification and a flag indicating a sample driver along with verbosity flags. It also accepts cargo manifest flags `--locked`, `--offline` and `--frozen`, which are forwarded to `cargo`.
 
 When the command completes the packaged driver artifacts are emitted at the path `target\<profile>\<project-name>-package`.
 
--- a/crates/cargo-wdk/README.md
+++ b/crates/cargo-wdk/README.md
@@ -69,12 +69,17 @@ Options:
       --sample                     Build sample class driver project
   -h, --help                       Print help
 
+Manifest Options:
+      --locked   Require Cargo.lock is up to date
+      --offline  Run without accessing the network
+      --frozen   Require Cargo.lock and cache are up to date
+
 Verbosity:
   -v, --verbose...  Increase logging verbosity
   -q, --quiet...    Decrease logging verbosity
 ```
 
-`build` takes a number of inputs specifying build profile (`dev` or `release`), target architecture (`amd64` or `arm64`), a flag enabling signature verification and a flag indicating a sample driver along with verbosity flags.
+`build` takes a number of inputs specifying build profile (`dev` or `release`), target architecture (`amd64` or `arm64`), a flag enabling signature verification and a flag indicating a sample driver along with verbosity flags. It also accepts the cargo manifest flags `--locked`, `--offline` and `--frozen`, which are forwarded to the underlying `cargo` invocations (e.g. `cargo build`, `cargo metadata`).
 
 When the command completes the packaged driver artifacts are emitted at the path `target\<profile>\<project-name>-package`.
 
@@ -112,4 +117,10 @@ If the `--verify-signature` flag is provided, the signatures are verified after 
 
     ```pwsh
     cargo wdk build --target-arch amd64
+    ```
+
+- To build a driver project with a strict, reproducible dependency lockfile (e.g. for CI), navigate to the root of the project and run:
+
+    ```pwsh
+    cargo wdk build --locked
     ```
--- a/crates/cargo-wdk/README.md
+++ b/crates/cargo-wdk/README.md
@@ -76,7 +76,7 @@ Verbosity:
   -q, --quiet...    Decrease logging verbosity
 ```
 
-`build` takes a number of inputs specifying build profile (`dev` or `release`), target architecture (`amd64` or `arm64`), the driver signing mode, a flag enabling signature verification and a flag indicating a sample driver, a flag enabling the `--locked` flag passthrough to `cargo`, along with verbosity flags.
+`build` takes a number of inputs specifying build profile (`dev` or `release`), target architecture (`amd64` or `arm64`), the driver signing mode, a flag enabling signature verification and a flag indicating a sample driver along with verbosity flags.
 
 When the command completes the packaged driver artifacts are emitted at the path `target\<profile>\<project-name>-package`.
 
--- a/crates/cargo-wdk/src/actions/build/build_task.rs
+++ b/crates/cargo-wdk/src/actions/build/build_task.rs
@@ -16,7 +16,7 @@ use wdk_build::CpuArchitecture;
 #[double]
 use crate::providers::exec::CommandExec;
 use crate::{
-    actions::{Profile, build::error::BuildTaskError, to_target_triple},
+    actions::{ManifestOptions, Profile, build::error::BuildTaskError, to_target_triple},
     providers::error::CommandError,
     trace,
 };
@@ -27,6 +27,7 @@ pub struct BuildTask<'a> {
     profile: Option<&'a Profile>,
     target_arch: Option<CpuArchitecture>,
     verbosity_level: clap_verbosity_flag::Verbosity,
+    manifest_options: ManifestOptions,
     manifest_path: PathBuf,
     command_exec: &'a CommandExec,
     working_dir: &'a Path,
@@ -41,6 +42,8 @@ impl<'a> BuildTask<'a> {
     /// * `profile` - An optional profile for the build
     /// * `target_arch` - The target architecture for the build
     /// * `verbosity_level` - The verbosity level for logging
+    /// * `manifest_options` - Cargo manifest options (`--locked`, `--frozen`,
+    ///   `--offline`) to forward to the underlying `cargo build` invocation
     /// * `command_exec` - The command execution provider
     ///
     /// # Returns
@@ -54,6 +57,7 @@ impl<'a> BuildTask<'a> {
         profile: Option<&'a Profile>,
         target_arch: Option<CpuArchitecture>,
         verbosity_level: clap_verbosity_flag::Verbosity,
+        manifest_options: ManifestOptions,
         command_exec: &'a CommandExec,
     ) -> Self {
         assert!(
@@ -66,6 +70,7 @@ impl<'a> BuildTask<'a> {
             profile,
             target_arch,
             verbosity_level,
+            manifest_options,
             manifest_path: working_dir.join("Cargo.toml"),
             command_exec,
             working_dir,
@@ -109,6 +114,7 @@ impl<'a> BuildTask<'a> {
             args.push("--target".to_string());
             args.push(to_target_triple(target_arch));
         }
+        args.extend(self.manifest_options.cargo_args().map(String::from));
         if let Some(flag) = trace::get_cargo_verbose_flags(self.verbosity_level) {
             args.push(flag.to_string());
         }
@@ -149,7 +155,7 @@ mod tests {
 
     use super::*;
     use crate::{
-        actions::Profile,
+        actions::{ManifestOptions, Profile},
         providers::{error::CommandError, exec::MockCommandExec},
     };
 
@@ -168,6 +174,7 @@ mod tests {
             Some(&profile),
             target_arch,
             verbosity_level,
+            ManifestOptions::default(),
             &command_exec,
         );
 
@@ -201,6 +208,7 @@ mod tests {
             profile.as_ref(),
             target_arch,
             verbosity_level,
+            ManifestOptions::default(),
             &command_exec,
         );
     }
@@ -257,6 +265,7 @@ mod tests {
             Some(&profile),
             Some(target_arch),
             verbosity,
+            ManifestOptions::default(),
             &mock,
         );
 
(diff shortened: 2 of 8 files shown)


crates/cargo-wdk/src/actions/mod.rs
```diff
@@ -90,3 +90,32 @@ impl Display for DriverType {
...
 }
+
+/// Cargo manifest options
+#[derive(clap::Args, Clone, Copy, Debug, Default, PartialEq, Eq)]
+#[command(next_help_heading = "Manifest Options")]
+pub struct ManifestOptions {
+    /// Require Cargo.lock is up to date
+    #[arg(long)]
+    pub locked: bool,
+    /// Run without accessing the network
```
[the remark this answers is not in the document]
svasista-ms: This is a valid point. I will take a look at this

crates/cargo-wdk/src/actions/build/tests.rs
```diff
@@ -265,6 +266,49 @@ pub fn given_a_driver_project_when_verify_signature_is_true_then_it_builds_succe
...
 
+#[test]
+pub fn given_a_driver_project_when_manifest_options_are_set_then_they_are_forwarded_to_cargo_invocations()
+ {
+    // Input CLI args
+    let cwd = PathBuf::from("C:\\tmp");
+    let profile = None;
+    let target_arch = CpuArchitecture::Amd64;
+    let verify_signature = false;
+    let sample_class = false;
```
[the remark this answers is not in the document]
svasista-ms: `cargo make wdk-pre-commit-flow` already fixes it locally

crates/cargo-wdk/src/actions/mod.rs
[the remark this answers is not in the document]
svasista-ms: Fixed ✅, the manifest options are now being forwarded to all `cargo` invocations in the `build` command

crates/cargo-wdk/src/providers/metadata.rs
```diff
@@ -42,12 +44,20 @@ impl Metadata {
     ///
     /// This function will return an error if the `cargo metadata` command fails
     /// to execute or if the specified path is not a valid Cargo project.
-    pub fn get_cargo_metadata_at_path(
+    pub fn get_cargo_metadata_at_path<'a>(
         &self,
-        working_dir: &Path,
+        working_dir: &'a Path,
+        other_options: &'a [&'a str],
```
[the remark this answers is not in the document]
svasista-ms: The explicit lifetime is needed here because this provider is `mockall::automock`-backed and the generated mock API doesn’t type-check with the elided form. I checked it by trying the change; it failed to compile, so we have to keep the lifetime annotation

crates/cargo-wdk/src/actions/build/build_task.rs
```diff
@@ -356,4 +367,98 @@ mod tests {
...
+            .withf(|command, args, _env, _wd| {
+                command == "cargo"
+                    && args.contains(&"--frozen")
+                    && args.contains(&"--locked")
+                    && args.contains(&"--offline")
+                    && {
+                        // Confirm fixed ordering: locked before offline before frozen
+                        let locked_idx = args.iter().position(|a| *a == "--locked").unwrap();
+                        let offline_idx = args.iter().position(|a| *a == "--offline").unwrap();
+                        let frozen_idx = args.iter().position(|a| *a == "--frozen").unwrap();
```
gurry: Why does order matter?

crates/cargo-wdk/src/actions/mod.rs
```diff
@@ -92,3 +92,32 @@ impl Display for DriverType {
...
+    pub offline: bool,
+    /// Require Cargo.lock and cache are up to date
+    #[arg(long)]
+    pub frozen: bool,
+}
+
+impl ManifestOptions {
+    /// Returns an iterator over the cargo CLI flags equivalent to this set of
+    /// options.
+    pub fn cargo_args(&self) -> impl Iterator<Item = &'static str> {
```
gurry:
```suggestion
    pub fn as_cargo_args(&self) -> impl Iterator<Item = &'static str> {
```
Slightly more idiomatic

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -64,14 +65,30 @@ fn kmdf_driver_builds_successfully() {
         assert!(output.status.success());
     }
 
-    clean_build_and_verify_project("kmdf", driver, None, None, None, None, None, None);
+    clean_build_and_verify_project("kmdf", driver, None, None, None, None, None, None, None);
+}
+
+#[test]
+fn kmdf_driver_builds_successfully_with_offline_flag() {
```
gurry: Add a test for `--locked` as well

crates/cargo-wdk/README.md
```diff
@@ -113,3 +118,9 @@ If the `--verify-signature` flag is provided, the signatures are verified after
     ```pwsh
     cargo wdk build --target-arch amd64
     ```
+
+- To build a driver project with a strict, reproducible dependency lockfile (e.g. for CI), navigate to the root of the project and run:
+
+    ```pwsh
+    cargo wdk build --locked
+    ```
```
gurry, line 128: This is example is not needed. We don't need to add examples for every commandline line arg

crates/cargo-wdk/README.md
```diff
@@ -69,12 +69,17 @@ Options:
...
+      --offline  Run without accessing the network
+      --frozen   Require Cargo.lock and cache are up to date
+
 Verbosity:
   -v, --verbose...  Increase logging verbosity
   -q, --quiet...    Decrease logging verbosity
 ```
 
-`build` takes a number of inputs specifying build profile (`dev` or `release`), target architecture (`amd64` or `arm64`), a flag enabling signature verification and a flag indicating a sample driver along with verbosity flags.
+`build` takes a number of inputs specifying build profile (`dev` or `release`), target architecture (`amd64` or `arm64`), a flag enabling signature verification and a flag indicating a sample driver along with verbosity flags. It also accepts the cargo manifest flags `--locked`, `--offline` and `--frozen`, which are forwarded to the underlying `cargo` invocations (e.g. `cargo build`, `cargo metadata`).
```
```suggestion
`build` takes a number of inputs specifying build profile (`dev` or `release`), target architecture (`amd64` or `arm64`), a flag enabling signature verification and a flag indicating a sample driver along with verbosity flags. It also accepts cargo manifest options `--locked`, `--offline` and `--frozen`, which are forwarded to `cargo`.
```
Slightly shorter

crates/cargo-wdk/README.md
```diff
@@ -65,17 +65,21 @@ Usage: cargo wdk build [OPTIONS]
...
       --target-arch <TARGET_ARCH>  Build for the target architecture
-      --sign-mode <SIGN_MODE>      Driver signing mode [default: test] [possible values: off, test]
       --verify-signature           Verify the signature
       --sample                     Build sample class driver project
   -h, --help                       Print help
 
+Manifest Options:
+      --locked   Require Cargo.lock is up to date
+      --offline  Run without accessing the network
+      --frozen   Require Cargo.lock and cache are up to date
```
gurry: This is from `cargo build`'s help:

```text
      --locked               Assert that `Cargo.lock` will remain unchanged
      --offline               Run without accessing the network
      --frozen               Equivalent to specifying both --locked and --offline
```

We should use the exact same text as far as possible. Doing so reinforces the idea that these flags are the same as `cargo` flags. So our help should be:

```text
      --locked               Assert that `Cargo.lock` will remain unchanged
      --offline               Build code without accessing the network
      --frozen               Equivalent to specifying both --locked and --offline
```

Notice that I'm using the word "Build" in the `--offline` help instead of "Run" used in `cargo build`'s `--offline` help. That is because in our case even if we pass `--offline` we will still access the network in our packaging phase.

Maybe this should cause a rethink on whether we should use the word `--offline` at all.

crates/cargo-wdk/src/actions/build/build_task.rs
svasista-ms: Resolving conversation as `--offline` and `--frozen` options have been removed.

crates/cargo-wdk/src/actions/mod.rs
svasista-ms: Resolving conversation as this function is no longer required

crates/cargo-wdk/tests/build_command_test.rs
svasista-ms: Added `kmdf_driver_builds_successfully_with_locked_flag` [test](https://github.com/microsoft/windows-drivers-rs/pull/656/changes#diff-7f39e8d21b60832dceaba74ceb186dce58e45d2c660c6ce0674d1be3181a18e9R71-R85)

crates/cargo-wdk/README.md
svasista-ms: Removed example from README

crates/cargo-wdk/README.md
svasista-ms: Changed to following:

> 
`build` takes a number of inputs specifying build profile (`dev` or `release`), target architecture (`amd64` or `arm64`), the driver signing mode, a flag enabling signature verification and a flag indicating a sample driver, a flag enabling the `--locked` flag passthrough to `cargo`, along with verbosity flags.

crates/cargo-wdk/README.md
svasista-ms: In an internal discussion we decided to add support for `locked` flag at the moment to ensure reproducible CI builds. Supporting `--offline` and `--frozen` (in the same sense as `cargo` does) means the `cargo wdk build` command must support/define behavior in air-gapped environments. Though `cargo` accepts these options and guarantees execution without accessing the network, the packaging phase, specifically time stamping using `signtool`, requires access to the network. [Issue 631](https://github.com/microsoft/windows-drivers-rs/issues/631) talks about the same. So, we decided to resolve offline support in a subsequent PR.

crates/cargo-wdk/src/actions/mod.rs
```diff
@@ -92,3 +92,6 @@ impl Display for DriverType {
         write!(f, "{s}")
     }
 }
+
+/// Cargo's `--locked` flag.
+pub const LOCKED_FLAG: &str = "--locked";
```
krishnakumar4a4: nit: Can we call it `LOCKED_PARAM` or `CARGO_LOCKED_PARAM` to make it clear that this represents a CLI parameter rather than a binary value? This would make the intent of the variable clearer at the call site.

crates/cargo-wdk/src/actions/build/build_task.rs
krishnakumar4a4, line 110: can we move these (--profile, --target) also to constants for consistency?

crates/cargo-wdk/src/cli.rs
```diff
@@ -208,6 +212,7 @@ impl Cli {
                         target_arch: cli_args.target_arch,
                         sign_mode,
                         is_sample_class: cli_args.sample,
+                        locked: cli_args.locked,
```
krishnakumar4a4, line 215: nit: The variable name `locked` sounds like it represents the driver's current state. For example, `is_sample_class` indicates to Build Action that the driver is a sample class driver. 

In locked option case, we are specifying what Build Action should do, so names like `use_locked` or `should_lock` would be more explicit. 

That said, I’m not strongly opinionated—since we only have one such parameter, it’s easy for someone to navigate the code and understand how it’s used. I just wanted to share this so we keep it in mind for naming in the future when we have multiple such options. 

It would be even better to create a struct called CargoOptions to pass all cargo options together to the action, but that's for later

crates/cargo-wdk/src/actions/mod.rs
svasista-ms: Changed to `CARGO_LOCKED_FLAG_NAME`

crates/cargo-wdk/src/actions/build/build_task.rs
svasista-ms: Done ✅

crates/cargo-wdk/src/cli.rs
svasista-ms: We decided to leave it as is for now and revisit when other options are added.

crates/cargo-wdk/README.md
```diff
@@ -68,14 +68,15 @@ Options:
...
+      --locked                     Assert that `Cargo.lock` will remain unchanged
   -h, --help                       Print help
 
 Verbosity:
   -v, --verbose...  Increase logging verbosity
   -q, --quiet...    Decrease logging verbosity
 ```
 
-`build` takes a number of inputs specifying build profile (`dev` or `release`), target architecture (`amd64` or `arm64`), the driver signing mode, a flag enabling signature verification and a flag indicating a sample driver along with verbosity flags.
+`build` takes a number of inputs specifying build profile (`dev` or `release`), target architecture (`amd64` or `arm64`), the driver signing mode, a flag enabling signature verification and a flag indicating a sample driver, a flag enabling the `--locked` flag passthrough to `cargo`, along with verbosity flags.
```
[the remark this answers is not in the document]
svasista-ms: Reverted

krishnakumar4a4 (approved):

crates/cargo-wdk/src/providers/metadata.rs
```diff
@@ -30,6 +30,8 @@ impl Metadata {
     ///
     /// * `working_dir` - A reference to a `Path` that specifies the path to the
     ///   working directory.
+    /// * `other_options` - Additional command-line flags to forward to `cargo
+    ///   metadata` (e.g. `--locked`).
```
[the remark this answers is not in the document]
svasista-ms: Fixed ✅

gurry (approved):

PR #657, opened 2026-05-13
ci: add explicit least-privilege permissions to read-only workflows

arpitjain099:
## What

Adds a workflow-level `permissions: contents: read` block to seven workflows that currently declare no permissions at all:

- `build.yaml`
- `code-formatting-check.yaml`
- `docs.yaml`
- `local-development-makefile.yaml`
- `test.yaml`
- `typos.yaml`
- `version-checks.yaml`

## Why

All seven are pure CI workflows — build, format check, docs build, makefile compile, tests, `typos` spell check, and version consistency check. None push commits, create releases, comment on PRs, or call any GitHub API endpoint that needs write scopes. The `GITHUB_TOKEN` for these workflows only needs read access for the checkout step, so declaring `contents: read` at the workflow level documents the minimum scope and removes reliance on the repository default token permissions.

This matches the existing project pattern: workflows that need elevated scopes (`cargo-audit.yaml`, `codeql.yaml`, `github-dependency-review.yaml`, `lint.yaml`) already declare them at the job level. This PR fills the gap for the read-only workflows that were leaving scope implicit.

## Verification

- YAML validity: `python3 -c "import yaml; yaml.safe_load(open(...))"` passes on each of the seven files.
- No change to jobs, steps, or matrix; only the workflow-level `permissions:` block is added.

## Reference

- [GitHub docs — Modifying the permissions for the GITHUB_TOKEN](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token)
- [OpenSSF Scorecard — Token-Permissions check](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions)

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -9,6 +9,9 @@ on:
   schedule: # Trigger a job on default branch at 4AM PST everyday
     - cron: 0 11 * * *
 
+permissions:
+  contents: read
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.compare || github.head_ref || github.ref }}
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
--- a/.github/workflows/code-formatting-check.yaml
+++ b/.github/workflows/code-formatting-check.yaml
@@ -9,6 +9,9 @@ on:
   schedule: # Trigger a job on default branch at 4AM PST everyday
     - cron: 0 11 * * *
 
+permissions:
+  contents: read
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.compare || github.head_ref || github.ref }}
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
--- a/.github/workflows/code-formatting-check.yaml
+++ b/.github/workflows/code-formatting-check.yaml
@@ -36,6 +36,9 @@ jobs:
   taplo-fmt:
     name: .toml Formatting Check
     runs-on: windows-2025
+    permissions:
+      contents: read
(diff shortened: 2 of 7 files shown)


arpitjain099:
Hi @svasista-ms, gentle ping on this. PR has been open for 4 days without review. I noticed you've been on the recent-merger side of recent merges in this repo. When you have a moment, would you mind giving it a quick look? No urgency. Happy to address any feedback.

svasista-ms:
Thanks for the PR @arpitjain099. We verified that the default `GITHUB_TOKEN` permissions are already set to **read-only at the org level**, so there's no security gap. However, it doesn't hurt to make this explicit at the workflow level, so we're happy to take this in.

A couple of small asks before we merge:
1. Could you update the PR title to follow Conventional Commits? Something like:
```ci: add explicit least-privilege permissions to read-only workflows```
2. Could you please address [Copilot's review note](https://github.com/microsoft/windows-drivers-rs/pull/657#discussion_r3231906790)?

I have approved the workflows. Once these issues are addressed and workflows pass, this is GTG. Thanks again!

arpitjain099:
Thanks @svasista-ms. Both addressed:

1. Title updated to conventional commits format.
2. For Copilot's note on `code-formatting-check.yaml` - added a per-job `permissions` block on `taplo-fmt` granting `contents: read` + `actions: write` so the `taiki-e/cache-cargo-install-action` step keeps working under the workflow-level read-only default. Other jobs in the workflow only need `contents: read`, so the broader scope stays scoped to the one job that uses the Actions cache.

Issue #502, opened 2025-09-09
Running tests on `cdylib` driver crates cause failure in Github Actions Runner


wmmc88:
Example failure:

```
     Running `D:\a\windows-drivers-rs\windows-drivers-rs\tests\mixed-package-kmdf-workspace\target\debug\deps\driver-3c34d004f375bef0.exe`
error: test failed, to rerun pass `-p driver --lib`

Caused by:
  process didn't exit successfully: `D:\a\windows-drivers-rs\windows-drivers-rs\tests\mixed-package-kmdf-workspace\target\debug\deps\driver-3c34d004f375bef0.exe` (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
```

This is odd since it only seems to manifest in github actions runners. Even on a local machine that is running a windows-server 2025 image does not repro this issue.

wmmc88:
One possible approach to this is to have the empty extern blocks with the link attribute be generated in `wdk-sys` and tie it to `cfg(not(feature=test-stubs))`. This should essentially be the same thing as what we currently do (ie. emit `cargo::rustc-link-lib=static` prints), but now we can prevent linking when compiling for tests

wmmc88:
@alan632 this should be solved by #685

PR #685, opened 2026-06-18
feat: add linker directive for VHF lib + move static linker directives from wdk-build to wdk-sys

Alan632:
## Summary

Adds conditionally compiled linker directives for the VHF library tied to the "hid" feature.

Moves build script emitted static linker directives (`cargo::rustc-link-lib=static=*`) from wdk-build/src/lib.rs into bindgen generated files as conditionally compiled attributes. The Rust source insertion happens in wdk-sys/build.rs while the backend logic and Rust source string build lives in wdk-build/src/lib.rs (closely following the pattern the bindgen generated headers use). This allows conditional compilation of linker directives (and eventually link args) without having to rely on cross crate feature signaling, and guards against issues from version drift (in case multiple versions of wdk-build are used in one project).

This PR is a redesign of and supersedes [PR!653](https://github.com/microsoft/windows-drivers-rs/pull/653).

## Verification

Verified both kmdf and umdf drivers built with "hid" and Vhf functions linked against VhfKm.lib and VhfUm.lib respectively. Additionally, inspected each driver's linker `.map` file for evidence of the respective Vhf symbols.

Bindgen generated files are also inspected post build for the presence of the link attributes.

### DIFF
--- a/crates/wdk-build/src/lib.rs
+++ b/crates/wdk-build/src/lib.rs
@@ -1075,10 +1075,13 @@ impl Config {
     /// [`ApiSubset`] rather than a set like
     /// [`Config::bindgen_header_contents`].
     ///
-    /// Each emitted directive is gated behind `#[cfg(not(feature =
-    /// "test-stubs"))]`, evaluated in the crate that compiles the generated
-    /// bindings (`wdk-sys`), so the directives are suppressed when `wdk-sys` is
-    /// built with its `test-stubs` feature.
+    /// Each emitted directive is gated behind
+    /// `#[cfg(not(any(test, feature = "test-stubs")))]`, evaluated in the crate
+    /// that compiles the generated bindings (`wdk-sys`). The directives are
+    /// therefore suppressed when `wdk-sys` is built with its `test-stubs`
+    /// feature, or when `wdk-sys` is itself compiled as a test target (e.g.
+    /// `cargo test -p wdk-sys`); see the `NOT_TEST_CFG` constant for the
+    /// rationale.
     #[must_use]
     pub fn bindgen_library_link_raw_lines(&self, api_subset: ApiSubset) -> String {
         self.libraries(api_subset)
@@ -1592,7 +1595,7 @@ pub fn configure_wdk_binary_build() -> Result<(), ConfigError> {
 }
 
 /// `cfg` predicate that gates the WDK link directives so that they are only
-/// emitted when the `wdk-sys` `test-stubs` feature is *not* enabled.
+/// emitted when the bindings are compiled for a real driver build.
 ///
 /// Downstream test executables that depend on `wdk-sys` enable its `test-stubs`
 /// feature (which provides stubbed symbols *and*, via this gate, suppresses
@@ -1600,7 +1603,13 @@ pub fn configure_wdk_binary_build() -> Result<(), ConfigError> {
 /// not driver binaries, so they never receive the WDK `rustc-link-search` paths
 /// that `Config::configure_binary_build` emits, and must therefore not attempt
 /// to link the real WDK libraries.
-const NOT_TEST_STUBS_CFG: &str = r#"not(feature = "test-stubs")"#;
+///
+/// The `test` predicate covers the analogous case where `wdk-sys` *itself* is
+/// compiled as a test target (e.g. `cargo test -p wdk-sys`): that build goes
+/// through `configure_library_build` rather than `configure_binary_build`, so
+/// it likewise lacks the `rustc-link-search` paths and must not link the real
+/// WDK libraries.
+const NOT_TEST_CFG: &str = r#"not(any(test, feature = "test-stubs"))"#;
 
 /// The `kind` of a `#[link]` attribute (i.e. how the library is linked).
 ///
@@ -1608,7 +1617,7 @@ const NOT_TEST_STUBS_CFG: &str = r#"not(feature = "test-stubs")"#;
 /// [rustc reference][1] for the full set.
 ///
 /// [1]: https://doc.rust-lang.org/reference/items/external-blocks.html#the-link-attribute
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum LinkKind {
     /// `kind = "static"`
     Static,
@@ -1629,7 +1638,7 @@ impl LinkKind {
 
 /// A linking modifier for a `#[link]` attribute (the `modifiers = "..."`
 /// field). Only the modifiers actually emitted by [`Config`] are modelled.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum LinkModifier {
     /// `-bundle`
     NoBundle,
@@ -1649,9 +1658,9 @@ impl LinkModifier {
 ///
 /// Rendered by [`LinkDirective::render`] as a `#[link(...)]` attribute on an
 /// (empty) `extern` block and emitted into the bindgen output via bindgen's
-/// `raw_line`. Every directive is gated behind `NOT_TEST_STUBS_CFG` so that
+/// `raw_line`. Every directive is gated behind [`NOT_TEST_CFG`] so that
 /// non-driver test consumers don't link the real WDK libraries.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 struct LinkDirective {
     /// `name = "..."` — the library to link.
     name: &'static str,
@@ -1700,7 +1709,7 @@ impl LinkDirective {
 #[cfg({cfg})]
 #[link(name = "{name}", kind = "{kind}"{modifiers})]
 unsafe extern "C" {{}}"#,
-            cfg = NOT_TEST_STUBS_CFG,
+            cfg = NOT_TEST_CFG,
             name = self.name,
             kind = self.kind.as_str(),
         )
@@ -2259,6 +2268,269 @@ mod tests {
             let result = config.compute_wdffunctions_symbol_name();
 
             assert_eq!(result, None);
+        }
+    }
+
+    mod link_directive_contents {
+        use super::*;
+        use crate::{KmdfConfig, UmdfConfig};
+
+        /// Collects the linked library names from a slice of
+        /// [`LinkDirective`]s, in order.
+        fn lib_names(directives: &[LinkDirective]) -> Vec<&str> {
+            directives.iter().map(|directive| directive.name).collect()
+        }
+
+        /// Builds a [`Config`] for the given target architecture and driver
+        /// model.
+        ///
+        /// Constructs [`Config`] directly (rather than via
+        /// `..Config::default()`) so these pure link-directive tests
+        /// don't require a real WDK/eWDK to be
+        /// present; `wdk_content_root` is irrelevant to the logic under test.
+        fn config_for(target_arch: &str, driver_config: DriverConfig) -> Config {
+            Config {
+                wdk_content_root: std::path::PathBuf::new(),
+                driver_config,
+                cpu_architecture: CpuArchitecture::try_from_cargo_str(target_arch)
+                    .expect("test target_arch should be a recognized CpuArchitecture"),
+            }
+        }
+
+        #[test]
+        fn render_minimal_directive() {
+            assert_eq!(
+                LinkDirective::new("Foo", LinkKind::Dylib).render(),
+                format!(
+                    "\n#[cfg({NOT_TEST_CFG})]\n#[link(name = \"Foo\", kind = \"dylib\")]\nunsafe \
+                     extern \"C\" {{}}"
+                )
+            );
+        }
+
+        #[test]
+        fn render_static_without_modifiers_omits_modifiers_key() {
+            assert_eq!(
+                LinkDirective::new("Foo", LinkKind::Static).render(),
+                format!(
+                    "\n#[cfg({NOT_TEST_CFG})]\n#[link(name = \"Foo\", kind = \"static\")]\nunsafe \
+                     extern \"C\" {{}}"
+                )
+            );
+        }
+
+        #[test]
+        fn render_with_no_bundle_modifier() {
+            assert_eq!(
+                LinkDirective::new("Foo", LinkKind::Static)
+                    .with_modifiers(&[LinkModifier::NoBundle])
+                    .render(),
+                format!(
+                    "\n#[cfg({NOT_TEST_CFG})]\n#[link(name = \"Foo\", kind = \"static\", \
+                     modifiers = \"-bundle\")]\nunsafe extern \"C\" {{}}"
+                )
+            );
+        }
+
+        #[test]
+        fn base_wdm_amd64() {
+            let config = config_for("x86_64", DriverConfig::Wdm);
+
+            assert_eq!(
+                lib_names(&config.base_libraries()),
+                ["BufferOverflowFastFailK", "ntoskrnl", "hal", "wmilib"]
+            );
+        }
+
+        #[test]
+        fn base_wdm_arm64_appends_arm64rt() {
+            let config = config_for("aarch64", DriverConfig::Wdm);
+
+            assert_eq!(
+                lib_names(&config.base_libraries()),
+                [
+                    "BufferOverflowFastFailK",
+                    "ntoskrnl",
+                    "hal",
+                    "wmilib",
+                    "arm64rt"
+                ]
+            );
+        }
+
+        #[test]
+        fn base_kmdf_amd64() {
+            let config = config_for("x86_64", DriverConfig::Kmdf(KmdfConfig::new()));
+
+            assert_eq!(
+                lib_names(&config.base_libraries()),
+                [
+                    "BufferOverflowFastFailK",
+                    "ntoskrnl",
+                    "hal",
+                    "wmilib",
+                    "WdfLdr",
+                    "WdfDriverEntry"
+                ]
+            );
+        }
+
+        #[test]
+        fn base_kmdf_arm64_appends_arm64rt() {
+            let config = config_for("aarch64", DriverConfig::Kmdf(KmdfConfig::new()));
+
+            assert_eq!(
+                lib_names(&config.base_libraries()),
+                [
+                    "BufferOverflowFastFailK",
+                    "ntoskrnl",
+                    "hal",
+                    "wmilib",
+                    "WdfLdr",
+                    "WdfDriverEntry",
+                    "arm64rt"
+                ]
+            );
+        }
+
+        #[test]
+        fn base_umdf_v2_includes_wdf_stub_libs() {
+            let config = config_for("x86_64", DriverConfig::Umdf(UmdfConfig::new()));
+
+            assert_eq!(
+                lib_names(&config.base_libraries()),
+                ["WdfDriverStubUm", "ntdll", "OneCoreUAP"]
+            );
+        }
+
+        #[test]
+        fn base_umdf_v1_omits_wdf_stub_libs() {
+            let config = config_for(
+                "x86_64",
+                DriverConfig::Umdf(UmdfConfig {
+                    umdf_version_major: 1,
+                    target_umdf_version_minor: 15,
+                    minimum_umdf_version_minor: None,
+                }),
+            );
+
+            assert_eq!(lib_names(&config.base_libraries()), ["OneCoreUAP"]);
+        }
+
+        #[test]
+        fn base_umdf_v2_arm64_omits_arm64rt() {
+            let config = config_for("aarch64", DriverConfig::Umdf(UmdfConfig::new()));
+
+            assert_eq!(
+                lib_names(&config.base_libraries()),
+                ["WdfDriverStubUm", "ntdll", "OneCoreUAP"]
+            );
+        }
+
+        #[test]
+        fn base_umdf_v1_arm64_omits_arm64rt_and_stub_libs() {
+            let config = config_for(
+                "aarch64",
+                DriverConfig::Umdf(UmdfConfig {
+                    umdf_version_major: 1,
+                    target_umdf_version_minor: 15,
+                    minimum_umdf_version_minor: None,
+                }),
+            );
+
+            assert_eq!(lib_names(&config.base_libraries()), ["OneCoreUAP"]);
+        }
+
+        #[test]
+        fn base_directives_are_all_static_no_bundle() {
+            let config = config_for("aarch64", DriverConfig::Kmdf(KmdfConfig::new()));
+
+            let directives = config.base_libraries();
+            assert!(
+                !directives.is_empty(),
+                "expected base libraries to assert on"
+            );
+            for directive in &directives {
+                assert_eq!(directive.kind, LinkKind::Static);
+                assert_eq!(directive.modifiers, [LinkModifier::NoBundle].as_slice());
+            }
+        }
+
+        #[test]
+        fn hid_wdm_and_kmdf_link_vhfkm_static() {
+            for driver_config in [DriverConfig::Wdm, DriverConfig::Kmdf(KmdfConfig::new())] {
+                let config = config_for("x86_64", driver_config);
+
+                assert_eq!(
+                    config.hid_libraries(),
+                    vec![
+                        LinkDirective::new("VhfKm", LinkKind::Static)
+                            .with_modifiers(&[LinkModifier::NoBundle])
+                    ]
+                );
+            }
+        }
+
+        #[test]
+        fn hid_umdf_links_vhfum_dylib_without_modifiers() {
+            let config = config_for("x86_64", DriverConfig::Umdf(UmdfConfig::new()));
+
+            assert_eq!(
+                config.hid_libraries(),
+                vec![LinkDirective::new("VhfUm", LinkKind::Dylib)]
+            );
+        }
+
+        #[test]
+        fn libraries_non_contributing_subsets_are_empty() {
+            let config = config_for("x86_64", DriverConfig::Kmdf(KmdfConfig::new()));
+
+            for subset in [
+                ApiSubset::Wdf,
+                ApiSubset::Gpio,
+                ApiSubset::ParallelPorts,
+                ApiSubset::Spb,
+                ApiSubset::Storage,
+                ApiSubset::Usb,
+            ] {
+                assert!(
+                    config.libraries(subset).is_empty(),
+                    "{subset:?} should contribute no libraries"
+                );
+            }
+        }
+
+        #[test]
+        fn libraries_base_subset_matches_base() {
+            let config = config_for("aarch64", DriverConfig::Kmdf(KmdfConfig::new()));
+
+            assert_eq!(config.libraries(ApiSubset::Base), config.base_libraries());
+        }
+
+        #[test]
+        fn libraries_hid_subset_matches_hid() {
+            let config = config_for("x86_64", DriverConfig::Umdf(UmdfConfig::new()));
+
+            assert_eq!(config.libraries(ApiSubset::Hid), config.hid_libraries());
+        }
+
+        #[test]
+        fn raw_lines_concatenate_all_libraries_in_subset() {
+            let config = config_for("x86_64", DriverConfig::Wdm);
+
+            // `bindgen_library_link_raw_lines` is the plain concatenation of each
+            // library's rendered `#[link]` block (no separators or wrapping), so
+            // it stays correct regardless of which libraries a subset contributes.
+            // The exact rendered format is locked by the `render_*` tests, and the
+            // per-subset library set by the `base_*`/`hid_*` tests.
+            assert_eq!(
+                config.bindgen_library_link_raw_lines(ApiSubset::Base),
+                config
+                    .base_libraries()
+                    .iter()
+                    .map(LinkDirective::render)
+                    .collect::<String>()
+            );
         }
     }
 
--- a/crates/wdk-build/src/lib.rs
+++ b/crates/wdk-build/src/lib.rs
@@ -1643,7 +1643,8 @@ impl LinkDirective {
         format!(
             r#"#[cfg({cfg})]
 #[link(name = "{name}", kind = "{kind}"{modifiers})]
-unsafe extern "C" {{}}"#,
+unsafe extern "C" {{}}
+"#,
             cfg = NOT_TEST_CFG,
             name = self.name,
             kind = self.kind.as_str(),
--- a/crates/wdk-build/src/lib.rs
+++ b/crates/wdk-build/src/lib.rs
@@ -1095,7 +1095,12 @@ impl Config {
     #[tracing::instrument(level = "trace")]
     fn base_link_directives(&self) -> Vec<LinkDirective<'static>> {
         const fn static_lib(name: &'static str) -> LinkDirective<'static> {
-            LinkDirective::new(name, "static").with_modifiers(&["-bundle"])
+            // Gated on `NOT_TEST_STUBS_CFG` so non-driver test consumers don't
+            // link the real WDK libraries (see the const's docs for the full
+            // rationale).
+            LinkDirective::new(name, "static")
+                .with_modifiers(&["-bundle"])
+                .with_cfg(NOT_TEST_STUBS_CFG)
         }
 
         let mut directives = Vec::new();
@@ -1143,9 +1148,15 @@ impl Config {
     fn hid_link_directives(&self) -> Vec<LinkDirective<'static>> {
         match self.driver_config {
             DriverConfig::Wdm | DriverConfig::Kmdf(_) => {
-                vec![LinkDirective::new("VhfKm", "static").with_modifiers(&["-bundle"])]
+                vec![
+                    LinkDirective::new("VhfKm", "static")
+                        .with_modifiers(&["-bundle"])
+                        .with_cfg(NOT_TEST_STUBS_CFG),
+                ]
             }
-            DriverConfig::Umdf(_) => vec![LinkDirective::new("VhfUm", "dylib")],
+            DriverConfig::Umdf(_) => {
+                vec![LinkDirective::new("VhfUm", "dylib").with_cfg(NOT_TEST_STUBS_CFG)]
+            }
         }
     }
 
@@ -1558,6 +1569,17 @@ where
 pub fn configure_wdk_binary_build() -> Result<(), ConfigError> {
     Config::from_env_auto()?.configure_binary_build()
 }
+
+/// `cfg` predicate that gates the WDK link directives so that they are only
+/// emitted when the `wdk-sys` `test-stubs` feature is *not* enabled.
+///
+/// Downstream test executables that depend on `wdk-sys` enable its `test-stubs`
+/// feature (which provides stubbed symbols *and*, via this gate, suppresses
+/// real WDK library linkage; see `wdk_sys::test_stubs`). Such executables are
+/// not driver binaries, so they never receive the WDK `rustc-link-search` paths
+/// that `Config::configure_binary_build` emits, and must therefore not attempt
+/// to link the real WDK libraries.
+const NOT_TEST_STUBS_CFG: &str = r#"not(feature = "test-stubs")"#;
 
 /// A native library to link into the generated bindings.
 ///
--- a/crates/wdk-build/src/lib.rs
+++ b/crates/wdk-build/src/lib.rs
@@ -1060,10 +1060,6 @@ impl Config {
     /// Returns the formatted `#[link]` raw lines for the given [`ApiSubset`],
     /// ready to be passed to bindgen's `raw_line`.
     ///
-    /// Each generated bindings file requests only the native libraries
-    /// introduced by its own [`ApiSubset`]. This avoids duplicate `#[link]`
-    /// directives across generated files.
-    ///
     /// Each emitted directive is gated behind
     /// `#[cfg(not(any(test, feature = "test-stubs")))]`, evaluated in the crate
     /// that compiles the generated bindings (`wdk-sys`). The directives are
@@ -2250,7 +2246,7 @@ mod tests {
         }
 
         #[test]
-        fn render_static_directive_includes_no_bundle_modifier() {
+        fn render_static_directive_disables_bundle() {
             assert_eq!(
                 LinkDirective::new("Foo", LinkKind::Static).render(),
                 format!(
@@ -2370,7 +2366,7 @@ mod tests {
         }
 
         #[test]
-        fn base_directives_are_all_static_no_bundle() {
+        fn base_directives_are_all_static_disables_bundle() {
             let config = config_for("aarch64", DriverConfig::Kmdf(KmdfConfig::new()));
 
             let directives = config.base_libraries();
--- a/crates/wdk-build/src/lib.rs
+++ b/crates/wdk-build/src/lib.rs
@@ -1125,20 +1125,12 @@ impl Config {
             println!("cargo::rustc-link-search={}", path.display());
         }
 
+        // TODO: Once [link-arg-attribute](https://doc.rust-lang.org/unstable-book/language-features/link-arg-attribute.html)
+        // stabilizes, the `cargo::rustc-cdylib-link-arg=*` directives will be moved to 
+        // the wdk-sys build script
         match &self.driver_config {
             DriverConfig::Wdm => {
                 // Emit WDM-specific libraries to link to
-                println!("cargo::rustc-link-lib=static=BufferOverflowFastFailK");
-                println!("cargo::rustc-link-lib=static=ntoskrnl");
-                println!("cargo::rustc-link-lib=static=hal");
-                println!("cargo::rustc-link-lib=static=wmilib");
-
-                // Emit ARM64-specific libraries to link to derived from
-                // WindowsDriver.arm64.props
-                if self.cpu_architecture == CpuArchitecture::Arm64 {
-                    println!("cargo::rustc-link-lib=static=arm64rt");
-                }
-
                 // Linker arguments derived from WindowsDriver.KernelMode.props in Ni(22H2) WDK
                 println!("cargo::rustc-cdylib-link-arg=/DRIVER");
                 println!("cargo::rustc-cdylib-link-arg=/NODEFAULTLIB");
@@ -1160,19 +1152,6 @@ impl Config {
             }
             DriverConfig::Kmdf(_) => {
                 // Emit KMDF-specific libraries to link to
-                println!("cargo::rustc-link-lib=static=BufferOverflowFastFailK");
-                println!("cargo::rustc-link-lib=static=ntoskrnl");
-                println!("cargo::rustc-link-lib=static=hal");
-                println!("cargo::rustc-link-lib=static=wmilib");
-                println!("cargo::rustc-link-lib=static=WdfLdr");
-                println!("cargo::rustc-link-lib=static=WdfDriverEntry");
-
-                // Emit ARM64-specific libraries to link to derived from
-                // WindowsDriver.arm64.props
-                if self.cpu_architecture == CpuArchitecture::Arm64 {
-                    println!("cargo::rustc-link-lib=static=arm64rt");
-                }
-
                 // Linker arguments derived from WindowsDriver.KernelMode.props in Ni(22H2) WDK
                 println!("cargo::rustc-cdylib-link-arg=/DRIVER");
                 println!("cargo::rustc-cdylib-link-arg=/NODEFAULTLIB");
@@ -1187,16 +1166,9 @@ impl Config {
                 // might not run` since `rustc` has no support for `/KERNEL`
                 println!("cargo::rustc-cdylib-link-arg=/IGNORE:4257");
             }
-            DriverConfig::Umdf(umdf_config) => {
-                // Emit UMDF-specific libraries to link to
-                if umdf_config.umdf_version_major >= 2 {
-                    println!("cargo::rustc-link-lib=static=WdfDriverStubUm");
-                    println!("cargo::rustc-link-lib=static=ntdll");
-                }
-
+            DriverConfig::Umdf(_) => {
                 println!("cargo::rustc-cdylib-link-arg=/NODEFAULTLIB:kernel32.lib");
                 println!("cargo::rustc-cdylib-link-arg=/NODEFAULTLIB:user32.lib");
-                println!("cargo::rustc-link-lib=static=OneCoreUAP");
 
                 // Linker arguments derived from WindowsDriver.UserMode.props in Ni(22H2) WDK
                 println!("cargo::rustc-cdylib-link-arg=/SUBSYSTEM:WINDOWS");
--- a/crates/wdk-build/src/lib.rs
+++ b/crates/wdk-build/src/lib.rs
@@ -1058,7 +1058,7 @@ impl Config {
     }
 
     /// Returns the formatted `#[link]` raw Strings for the given [`ApiSubset`].
-    /// Returns [`None`] if the [`ApiSubset`] does not define any libraries to
+    /// Returns `None` if the [`ApiSubset`] does not define any libraries to
     /// link.
     ///
     /// Each emitted directive is gated behind the `NOT_TEST_CFG` logic.
--- a/crates/wdk-build/src/lib.rs
+++ b/crates/wdk-build/src/lib.rs
@@ -18,6 +18,7 @@ use std::{
     sync::LazyLock,
 };
 
+use ::bindgen::Builder;
 pub use bindgen::BuilderExt;
 use metadata::TryFromCargoMetadataError;
 use tracing::debug;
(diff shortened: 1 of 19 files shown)


crates/wdk-build/src/lib.rs
```diff
@@ -1495,6 +1559,123 @@ pub fn configure_wdk_binary_build() -> Result<(), ConfigError> {
...
+/// Rendered as a `#[link(...)]` attribute on an (empty) `extern` block and
+/// emitted into the bindgen output via [`Builder::raw_line`]. An optional
+/// `cfg` predicate gates the directive (e.g. by driver type).
+#[derive(Debug, Clone)]
+pub struct LinkDirective<'a> {
+    /// `#[link(name = "...")]` — the library to link.
+    name: &'a str,
+    /// `kind = "..."` — one of `static`, `dylib`, `raw-dylib`, `framework`.
+    /// `#[link(name = "VhfKm", kind = "...")]`
+    kind: &'a str,
```
wmmc88: if its a closed set, it should be an enum

crates/wdk-build/src/lib.rs
```diff
@@ -1495,6 +1559,123 @@ pub fn configure_wdk_binary_build() -> Result<(), ConfigError> {
...
+    /// `modifiers = "..."` values (e.g. `-bundle`, `+whole-archive`). Joined
+    /// with commas; an empty slice omits the `modifiers` key entirely.
+    /// `#[link(name = "VhfKm", kind = "static", modifiers = "...")]`
+    modifiers: &'a [&'a str],
+    /// ABI of the emitted `extern` block (e.g. `C`, `system`).
+    abi: &'a str,
+    /// Inner predicate of `#[cfg(...)]` gating the directive. `None` emits no
+    /// cfg.
+    cfg: Option<&'a str>,
```
wmmc88: YAGNI. Less code is better. If all our directives are static and gated by test, we don't need this configuration nob

crates/wdk-sys/build.rs
```diff
@@ -269,9 +270,13 @@ fn generate_base(out_path: &Path, config: &Config) -> Result<(), ConfigError> {
...
 
-    let bindgen_builder = bindgen::Builder::wdk_default(config)?
-        .with_codegen_config((CodegenConfig::TYPES | CodegenConfig::VARS).complement())
-        .header_contents(&format!("{outfile_name}-input.h"), &header_contents);
+    let bindgen_builder = {
+        let builder = bindgen::Builder::wdk_default(config)?
+            .with_codegen_config((CodegenConfig::TYPES | CodegenConfig::VARS).complement())
+            .header_contents(&format!("{outfile_name}-input.h"), &header_contents);
+
+        add_link_directives(builder, &config.link_directives(ApiSubset::Base))
```
wmmc88: If you mirror the `header_contents` implementation, you would just have a `bindgen_library_link_raw_lines` (or similar name) to get the formatted rawlines and just pass that directly to [raw_line](https://docs.rs/bindgen/latest/bindgen/struct.Builder.html#method.raw_line). That would make it so that you don't need this free function and can continue to chain the builder construction like with `header_contents`

crates/wdk-build/src/lib.rs
```diff
@@ -1495,6 +1559,123 @@ pub fn configure_wdk_binary_build() -> Result<(), ConfigError> {
...
+    /// Gates the directive behind a `#[cfg(...)]` predicate.
+    #[must_use]
+    pub const fn with_cfg(mut self, cfg: &'a str) -> Self {
+        self.cfg = Some(cfg);
+        self
+    }
+
+    /// Renders this directive into a self-contained block of Rust source.
+    fn render(&self) -> String {
+        const VALID_KINDS: [&str; 4] = ["static", "dylib", "raw-dylib", "framework"];
```
wmmc88: you don't need validation, if the inputs are types instead of freeform strings

crates/wdk-build/src/lib.rs
```diff
@@ -1495,6 +1559,123 @@ pub fn configure_wdk_binary_build() -> Result<(), ConfigError> {
     Config::from_env_auto()?.configure_binary_build()
 }
 
+/// A native library to link into the generated bindings.
+///
+/// Rendered as a `#[link(...)]` attribute on an (empty) `extern` block and
+/// emitted into the bindgen output via [`Builder::raw_line`]. An optional
+/// `cfg` predicate gates the directive (e.g. by driver type).
+#[derive(Debug, Clone)]
+pub struct LinkDirective<'a> {
```
wmmc88: I don't think this needs to be `pub`: https://github.com/microsoft/windows-drivers-rs/pull/685/changes#r3439485652

crates/wdk-build/src/lib.rs
```diff
@@ -1495,6 +1559,123 @@ pub fn configure_wdk_binary_build() -> Result<(), ConfigError> {
...
+            .map(|cfg| format!("#[cfg({cfg})]"))
+            .unwrap_or_default();
+
+        // Emitted as one block so the attributes always stay attached to the
+        // following `extern` block, regardless of where bindgen places raw
+        // lines.
+        format!(
+            "\n{cfg_attr}\n{link_attr}\nunsafe extern \"{}\" {{}}",
+            self.abi
+        )
```
wmmc88: This should really just be one `format` call to just insert the libraries to link against so that its more readable. 

You can also see some attempts elsewhere in this repo where there's some string templating to make things more comprehensible when generating rust code: https://github.com/Alan632/windows-drivers-rs/blob/e8756f10dc4dcab932223a4e7b4827c7f3f4d7ff/crates/wdk-sys/build.rs#L54

crates/wdk-sys/build.rs
wmmc88: That `bindgen_library_link_raw_lines` would call a `config.libraries(api_subsets)` just the same as `header_contents` calls `headers`

wmmc88 (changes requested):

crates/wdk-build/src/lib.rs
Alan632: Done! With the exception of leaving a `format!` in there due to needing to build the modifiers string, although this could be condensed if we're not expecting to use more than one (eg. currently only applying `-bundle`).

crates/wdk-build/src/lib.rs
Alan632: Fixed!

crates/wdk-build/src/lib.rs
Alan632: Done! I created types for them instead of just using raw strings.

crates/wdk-sys/build.rs
Alan632: Redesigned the implementation to closely mirror the `header_contents` pattern.

crates/wdk-build/src/lib.rs
Alan632: Valid points, removed everything that I don't expect we would want to configure.

crates/wdk-build/src/lib.rs
Alan632: Good point, done! 

see also: https://github.com/microsoft/windows-drivers-rs/pull/685#discussion_r3444955827

crates/wdk-sys/build.rs
```diff
@@ -282,6 +409,62 @@ fn generate_base(out_path: &Path, config: &Config) -> Result<(), ConfigError> {
...
+/// The `cargo::rustc-cdylib-link-arg=*` linker arguments it emits will
+/// be moved over once .
+///
+/// `-bundle` is used so the libraries are resolved during the final driver link
+/// (where `configure_binary_build` emits the `rustc-link-search` paths) instead
+/// of when `wdk-sys` itself is compiled.
+///
+/// TODO: Once [link-arg-attribute](https://doc.rust-lang.org/unstable-book/language-features/link-arg-attribute.html)
+/// stabilizes, the `cargo::rustc-cdylib-link-arg=*` directives from
+/// [`wdk_build::Config::configure_binary_build`] will be added
```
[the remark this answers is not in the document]
Alan632: out of date

crates/wdk-build/src/lib.rs
```diff
@@ -1125,20 +1125,12 @@ impl Config {
             println!("cargo::rustc-link-search={}", path.display());
         }
 
+        // TODO: Once [link-arg-attribute](https://doc.rust-lang.org/unstable-book/language-features/link-arg-attribute.html)
+        // stabilizes, the `cargo::rustc-cdylib-link-arg=*` directives will be moved to 
+        // the wdk-sys build script
```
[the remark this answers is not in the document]
Alan632: fixed

crates/wdk-sys/build.rs
```diff
@@ -360,7 +543,19 @@ fn generate_hid(out_path: &Path, config: &Config) -> Result<(), ConfigError> {
         for header_file in config.headers(ApiSubset::Hid)? {
             builder = builder.allowlist_file(format!("(?i).*{header_file}.*"));
         }
-        builder
+
+        // Add a link directive for VhfKm/VhfUm when feature "hid" enabled
+        // cfg's provide conditional compilation logic
```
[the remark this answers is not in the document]
Alan632: outdated

crates/wdk-build/src/lib.rs
```diff
@@ -1057,6 +1058,97 @@ impl Config {
...
+            | ApiSubset::Usb => Vec::new(),
+        }
+    }
+
+    /// Builds the static library link directives for the base bindings.
+    ///
+    /// TODO: Once [link-arg-attribute](https://doc.rust-lang.org/unstable-book/language-features/link-arg-attribute.html)
+    /// stabilizes, the `cargo::rustc-cdylib-link-arg=*` directives from
+    /// [`wdk_build::Config::configure_binary_build`] will be added
+    #[tracing::instrument(level = "trace")]
```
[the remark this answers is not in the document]
Alan632: see https://github.com/microsoft/windows-drivers-rs/pull/685#discussion_r3438201361

crates/wdk-build/src/lib.rs
```diff
@@ -1125,20 +1217,12 @@ impl Config {
...
-                println!("cargo::rustc-link-lib=static=hal");
-                println!("cargo::rustc-link-lib=static=wmilib");
-
-                // Emit ARM64-specific libraries to link to derived from
-                // WindowsDriver.arm64.props
-                if self.cpu_architecture == CpuArchitecture::Arm64 {
-                    println!("cargo::rustc-link-lib=static=arm64rt");
-                }
-
                 // Linker arguments derived from WindowsDriver.KernelMode.props in Ni(22H2) WDK
```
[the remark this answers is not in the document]
Alan632: fixed

crates/wdk-build/src/lib.rs
```diff
@@ -1160,19 +1244,6 @@ impl Config {
...
-                println!("cargo::rustc-link-lib=static=WdfDriverEntry");
-
-                // Emit ARM64-specific libraries to link to derived from
-                // WindowsDriver.arm64.props
-                if self.cpu_architecture == CpuArchitecture::Arm64 {
-                    println!("cargo::rustc-link-lib=static=arm64rt");
-                }
-
                 // Linker arguments derived from WindowsDriver.KernelMode.props in Ni(22H2) WDK
                 println!("cargo::rustc-cdylib-link-arg=/DRIVER");
```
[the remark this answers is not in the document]
Alan632: fixed

tests/wdk-sys-tests/Cargo.toml
```diff
@@ -12,4 +12,4 @@ driver-type = "WDM"
 [lib]
 
 [dependencies]
-wdk-sys = { path = "../../crates/wdk-sys" }
+wdk-sys = { features = ["test-stubs"], path = "../../crates/wdk-sys" }
```
leon-xd, line 15: feel like this addition is indicative that this PR is a breaking change. if i'm understanding this correctly the generated link directives are gated behind `test-stubs` not being a feature. unintended consequence is any downstream crate that has tests but never relies on a stub symbol that we generate technically never needed this feature (like this file). very very narrow case but they will need documentation that they need this and is technically a change of the contract of our API.

Read up on [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#:~:text=Learn%20how%20to%20use%20Conventional%20Commits,%20a) to see how to handle this both in the PR title and in the PR description, as in the release process these will be used to generate the release notes

leon-xd: oh way bigger than the situation i just put out -- if a driver has a mixed crate of both tests and code and they enable test stubs (i believe some of our drivers may do this) they need to separate the tests out into a separate crate or else the driver won't be able to link at all

`not(any(test, feature = "test-stubs"))` means that either running `cargo test` or having the test stubs feature enabled blocks linking from happening. `test-stubs` is unconditionally on even if you're not testing if you've included it unless I am really out of whack with how rust features work

leon-xd: this may be cause to rethink the gate on the link. maybe instead of `any` we say `all`? but not sure about how this plays around

crates/wdk-build/src/lib.rs
```diff
@@ -1057,6 +1057,133 @@ impl Config {
             .collect())
     }
 
+    /// Returns the formatted `#[link]` raw lines for the given [`ApiSubset`],
+    /// ready to be passed to bindgen's `raw_line`.
+    ///
+    /// This is the link-directive analogue of
+    /// [`Config::bindgen_header_contents`]: it maps the [`ApiSubset`] (and the
+    /// active [`DriverConfig`] and [`CpuArchitecture`]) to the native libraries
```
leon-xd: this shouldn't be in parentheses

crates/wdk-build/src/lib.rs
```diff
@@ -1057,6 +1057,133 @@ impl Config {
...
+    /// that must be linked, rendering them as `#[link(...)]` attributes emitted
+    /// into the generated bindings.
+    ///
+    /// Unlike header selection, link subsets are selected *non-cumulatively*:
+    /// each generated bindings file requests only its own subset's libraries
+    /// (e.g. `hid.rs` requests only [`ApiSubset::Hid`], not
+    /// [`ApiSubset::Base`]), since the base libraries are already emitted
+    /// into the base bindings file. This avoids duplicate `#[link]`
```
leon-xd: i feel like having such a direct comparison to another method is a bit confusing -- i get timeline wise this comes after the other one but this should be self contained

leon-xd: esp if there are changes to the other method. so i would remove all references to the headers

crates/wdk-build/src/lib.rs
```diff
@@ -1057,6 +1057,133 @@ impl Config {
...
+    /// directives across generated files, and is why this takes a single
+    /// [`ApiSubset`] rather than a set like
+    /// [`Config::bindgen_header_contents`].
+    ///
+    /// Each emitted directive is gated behind
+    /// `#[cfg(not(any(test, feature = "test-stubs")))]`, evaluated in the crate
+    /// that compiles the generated bindings (`wdk-sys`). The directives are
+    /// therefore suppressed when `wdk-sys` is built with its `test-stubs`
+    /// feature, or when `wdk-sys` is itself compiled as a test target (e.g.
+    /// `cargo test -p wdk-sys`); see the `NOT_TEST_CFG` constant for the
```
leon-xd: `NOT_TEST_CFG` is private so any documentation there won't be generated. move that rationale over here

crates/wdk-build/src/lib.rs
```diff
@@ -2050,6 +2271,269 @@ mod tests {
...
+                    .expect("test target_arch should be a recognized CpuArchitecture"),
+            }
+        }
+
+        #[test]
+        fn render_minimal_directive() {
+            assert_eq!(
+                LinkDirective::new("Foo", LinkKind::Dylib).render(),
+                format!(
+                    "\n#[cfg({NOT_TEST_CFG})]\n#[link(name = \"Foo\", kind = \"dylib\")]\nunsafe \
```
leon-xd: may want to consider writing out the `NOT_TEST_CFG` for all of these. or even better just have a test that asserts the `NOT_TEST_CFG` is what we expect it to be

crates/wdk-build/src/lib.rs
```diff
@@ -1057,6 +1057,133 @@ impl Config {
...
+            // emitted (by `Config::configure_binary_build`). Without it, building
+            // wdk-sys's rlib would fail since those search paths aren't available
+            // then.
+            LinkDirective::new(name, LinkKind::Static).with_modifiers(&[LinkModifier::NoBundle])
+        }
+
+        let mut directives = Vec::new();
+
+        // Base libraries derived from WindowsDriver.KernelMode.props (WDM/KMDF)
+        // and WindowsDriver.UserMode.props (UMDF) in the Ni(22H2) WDK.
```
leon-xd: is there a rationale for this being a code comment and not in the rustdoc?

leon-xd: same comment for the `ARM64` comment below in this function

crates/wdk-build/src/lib.rs
```diff
@@ -2050,6 +2271,269 @@ mod tests {
...
+            for subset in [
+                ApiSubset::Wdf,
+                ApiSubset::Gpio,
+                ApiSubset::ParallelPorts,
+                ApiSubset::Spb,
+                ApiSubset::Storage,
+                ApiSubset::Usb,
+            ] {
+                assert!(
+                    config.libraries(subset).is_empty(),
```
leon-xd, line 2365: empty is defined by the private library -- if we want empty to be `""` we should pin it here with an `assert_eq`. this can be done in this test or another test

crates/wdk-sys/build.rs
```diff
@@ -360,7 +361,8 @@ fn generate_hid(out_path: &Path, config: &Config) -> Result<(), ConfigError> {
         for header_file in config.headers(ApiSubset::Hid)? {
             builder = builder.allowlist_file(format!("(?i).*{header_file}.*"));
         }
-        builder
+
+        builder.raw_line(config.bindgen_library_link_raw_lines(ApiSubset::Hid))
```
leon-xd: should this logic be duplicated for all subsets? even if it's an empty string

tests/wdk-sys-tests/Cargo.toml
leon-xd: also take a look at config-kmdf and config-umdf tests to see if they need test stubs

crates/wdk-build/src/lib.rs
```diff
@@ -1057,6 +1057,133 @@ impl Config {
...
+            }
+            DriverConfig::Umdf(umdf_config) => {
+                if umdf_config.umdf_version_major >= 2 {
+                    directives.extend([static_lib("WdfDriverStubUm"), static_lib("ntdll")]);
+                }
+                directives.push(static_lib("OneCoreUAP"));
+            }
+        }
+
+        // ARM64-specific libraries to link to derived from
```
leon-xd: this block might be better duplicated in the exhaustive match above. more of a code style thing but having one match and having logic explicitly duplicated across the kmdf and wdm arms makes more sense to me given that's already there

crates/wdk-build/src/lib.rs
```diff
@@ -1057,6 +1057,133 @@ impl Config {
...
+    #[tracing::instrument(level = "trace")]
+    fn base_libraries(&self) -> Vec<LinkDirective> {
+        const fn static_lib(name: &'static str) -> LinkDirective {
+            // `LinkModifier::NoBundle` (rustc `-bundle`): don't pack the static
+            // lib into wdk-sys's rlib; defer its linking to the final driver
+            // binary, which is where the WDK `rustc-link-search` paths are
+            // emitted (by `Config::configure_binary_build`). Without it, building
+            // wdk-sys's rlib would fail since those search paths aren't available
+            // then.
+            LinkDirective::new(name, LinkKind::Static).with_modifiers(&[LinkModifier::NoBundle])
```
leon-xd: functionally since NoBundle is an invariant should this be encoded directly into the `LinkDirective`? We don't want bundling no matter what and the state we never want to encode should be unrepresentable. I like what you were thinking with LinkModifier but it's too much flexibility given what we want, if we want that kind of flexibility down the road that can be added but to me it's redundant

crates/wdk-build/src/lib.rs
```diff
@@ -1057,6 +1057,133 @@ impl Config {
...
+    ///
+    /// Each emitted directive is gated behind
+    /// `#[cfg(not(any(test, feature = "test-stubs")))]`, evaluated in the crate
+    /// that compiles the generated bindings (`wdk-sys`). The directives are
+    /// therefore suppressed when `wdk-sys` is built with its `test-stubs`
+    /// feature, or when `wdk-sys` is itself compiled as a test target (e.g.
+    /// `cargo test -p wdk-sys`); see the `NOT_TEST_CFG` constant for the
+    /// rationale.
+    #[must_use]
+    pub fn bindgen_library_link_raw_lines(&self, api_subset: ApiSubset) -> String {
```
leon-xd: interesting tension here -- this couples `wdk-sys` and `wdk-build` given that we have a hard dependency on the `test-stubs` feature to exist. but this is a public API. more of a philosophy question for @wmmc88 but how do we resolve the fact that this needs to be public for `wdk-sys` to use this but is also so specific to `wdk-sys` that it is likely not useful to be used for other functions? do we genericize this and have `wdk-sys` determine the specific shape of the line?

leon-xd: we can shrug our shoulders and say wdk-build is meant to be for wdk-sys. or maybe this is a sign that all of this is meant to go into wdk-sys?

leon-xd: or keep this `#[doc(hidden)]` -- that's what i did for the `syn` parsing pr that is out. but iirc melvin left a comment asking why i did that so maybe that's not right

leon-xd: i have a few comments about blank lines and calling this on all api subsets -- it'd be good to have this be an `Option<String>` in case there is no string that needs to be generated for the given API subset

crates/wdk-build/src/lib.rs
```diff
@@ -1057,6 +1057,133 @@ impl Config {
...
+            | ApiSubset::Usb => Vec::new(),
+        }
+    }
+
+    /// Builds the static library link directives for the base bindings.
+    ///
+    /// TODO: Once [link-arg-attribute](https://doc.rust-lang.org/unstable-book/language-features/link-arg-attribute.html)
+    /// stabilizes, the `cargo::rustc-cdylib-link-arg=*` directives emitted by
+    /// [`wdk_build::Config::configure_binary_build`] will be moved here too.
+    #[tracing::instrument(level = "trace")]
```
leon-xd: does this mean that functions that call methods with tracing need to have this annotation too?

crates/wdk-build/src/lib.rs
```diff
@@ -1057,6 +1057,133 @@ impl Config {
...
+                    directives.extend([static_lib("WdfDriverStubUm"), static_lib("ntdll")]);
+                }
+                directives.push(static_lib("OneCoreUAP"));
+            }
+        }
+
+        // ARM64-specific libraries to link to derived from
+        // WindowsDriver.arm64.props.
+        if matches!(
+            self.driver_config,
```
leon-xd: if you don't change this we are not matching on `&self` like we did above

crates/wdk-build/src/lib.rs
```diff
@@ -1057,6 +1057,133 @@ impl Config {
...
+        ) && self.cpu_architecture == CpuArchitecture::Arm64
+        {
+            directives.push(static_lib("arm64rt"));
+        }
+
+        directives
+    }
+
+    #[tracing::instrument(level = "trace")]
+    fn hid_libraries(&self) -> Vec<LinkDirective> {
```
leon-xd, line 1152: doc comment

crates/wdk-build/src/lib.rs
```diff
@@ -1495,6 +1594,128 @@ pub fn configure_wdk_binary_build() -> Result<(), ConfigError> {
...
+/// field). Only the modifiers actually emitted by [`Config`] are modelled.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum LinkModifier {
+    /// `-bundle`
+    NoBundle,
+}
+
+impl LinkModifier {
+    /// Returns the string used in the `modifiers = "..."` field of a `#[link]`
+    /// attribute, including its `+`/`-` prefix.
```
leon-xd: I think we may get rid of this enum but in the future encode the `+` and `-` as enum variants

crates/wdk-build/src/lib.rs
```diff
@@ -1495,6 +1594,128 @@ pub fn configure_wdk_binary_build() -> Result<(), ConfigError> {
...
+                .modifiers
+                .iter()
+                .map(|modifier| modifier.as_str())
+                .collect::<Vec<_>>()
+                .join(",");
+            format!(r#", modifiers = "{modifiers}""#)
+        };
+
+        format!(
+            r#"
```
leon-xd: is there a reason we have a new line to start this

tests/wdk-sys-tests/Cargo.toml
leon-xd, line 15: we should write a comment in `wdk-sys/src/test_stubs.rs` that mentions this change?

leon-xd (changes requested):
Great work. Few changes here and there but the overall shape is very solid!

tests/wdk-sys-tests/Cargo.toml
wmmc88: image

This is what i suggested offline to Alan

Alan632: My understanding is that if `test-stubs` is enabled then any tests available should not use any real libraries that would require linking (opting instead to stub out/mock whatever is needed, similar to what `wdk-sys/src/test_stubs.rs` does), and we should only be turning that feature on in `dev-dependencies` (only run with `cargo test`). 

I think there isn't adequate documentation in the root level README for the intent around the `test-stubs` and how testing should be done. I can add a section on testing in the README unless there's a specific reason it's not there in the first place @wmmc88?

wmmc88: > not(any(test, feature = "test-stubs")) means that either running cargo test or having the test stubs feature enabled blocks linking from happening. test-stubs is unconditionally on even if you're not testing if you've included it unless I am really out of whack with how rust features work

@leon-xd that situation is fine since you can only turn on the `test-stubs` feature in tests by enabling it under `dev-dependencies`

> My understanding is that if test-stubs is enabled then any tests available should not use any real libraries that would require linking (opting instead to stub out/mock whatever is needed, similar to what wdk-sys/src/test_stubs.rs does), and we should only be turning that feature on in dev-dependencies (only run with cargo test).

@Alan632 tests are still free to call APIs from WDK libraries, but the expectation is that they are not linked so the test writer would write a mock for it. 

> I think there isn't adequate documentation in the root level README for the intent around the test-stubs and how testing should be done. I can add a section on testing in the README unless there's a specific reason it's not there in the first place

It's doesn't feature prominent documentation on the top level README because there isn't a full end-to-end unit testing scenario in this repo showcasing it. 

Before this PR's change, tests in a `cdylib` crate wouldn't work because the cdylib crate forced linking against (possibly) km apis always via the cdylib's build.rs call to `configure_binary_build`. See https://github.com/microsoft/windows-drivers-rs/issues/502. But it was possible to have unit-tests in library crates since they don't call `configure_binary_build`. The problem became that when compiling the binary test targets, you would get linking issues due to missing global symbols that would normally be provided by the WDK. These global symbols missing is why the `test-stubs` feature exists: https://github.com/microsoft/windows-drivers-rs/blob/0e3499d68daee51a22dabf2965b7a25159621234/crates/wdk-sys/src/test_stubs.rs#L4-L8. 

After this PR's change, we can also have tests in the `cdylib` crate itself since we no longer force the WDK linking to happen when compiling the test target. 

`wdk-sys-tests` is a special case right now because wdk-config-reliant tests are in our CI matrix rn and are a little finicky to get right. So we just tossed those WDK reliant tests into a separate crate and made it compile with WDM WDK metadata.

wmmc88: Thinking about this more actually, I think there's a more problematic issue here.  With this pr's change, libraries that depend on `wdk-sys` won't be able to build their normal `lib` targets when there isn't a WDK configuration in the environment. There needs to be a solution for that

crates/wdk-build/src/lib.rs
Alan632: Fixed

crates/wdk-build/src/lib.rs
Alan632: Fixed

crates/wdk-build/src/lib.rs
Alan632: Fixed

crates/wdk-build/src/lib.rs
Alan632: Changed test to a more directed assert on `NOT_TEST_CFG`

crates/wdk-build/src/lib.rs
Alan632: Originally just mirroring the in function comments from the section they were pulled from.

https://github.com/microsoft/windows-drivers-rs/blob/21ec8f86ea16766ec5cebc33341de744f3119bbf/crates/wdk-build/src/lib.rs#L1136

Moved into doc comment

crates/wdk-build/src/lib.rs
Alan632: Added a test for the empty subsets asserting `None` since we changed `bindgen_library_link_raw_lines` return to an `Option<String>`

crates/wdk-sys/build.rs
Alan632: Added to the other generate_* functions

crates/wdk-build/src/lib.rs
Alan632: Fixed

crates/wdk-build/src/lib.rs
Alan632: Removed LinkModifier and moved intent into LinkDirective

crates/wdk-build/src/lib.rs
Alan632: Changed to an `Option<String>`

crates/wdk-build/src/lib.rs
Alan632: Removed the tracing attribute

crates/wdk-build/src/lib.rs
Alan632: Fixed

crates/wdk-build/src/lib.rs
Alan632: Done

crates/wdk-build/src/lib.rs
Alan632: Removed in lieu of hardcoding it into LinkDirective, similar to above comment https://github.com/microsoft/windows-drivers-rs/pull/685#discussion_r3494419932

crates/wdk-build/src/lib.rs
Alan632: Removed unnecessary new line

tests/wdk-sys-tests/Cargo.toml
Alan632: Added comment in `test_stubs.rs`

leon-xd: @alan632 do you have a solution for this?

crates/wdk-build/src/lib.rs
```diff
@@ -1060,34 +1060,31 @@ impl Config {
...
-    /// Unlike header selection, link subsets are selected *non-cumulatively*:
-    /// each generated bindings file requests only its own subset's libraries
-    /// (e.g. `hid.rs` requests only [`ApiSubset::Hid`], not
-    /// [`ApiSubset::Base`]), since the base libraries are already emitted
-    /// into the base bindings file. This avoids duplicate `#[link]`
-    /// directives across generated files, and is why this takes a single
-    /// [`ApiSubset`] rather than a set like
-    /// [`Config::bindgen_header_contents`].
+    /// Each generated bindings file requests only the native libraries
+    /// introduced by its own [`ApiSubset`]. This avoids duplicate `#[link]`
```
leon-xd: nit: i think we can delete this entire paragraph, the first sentence is pretty much saying the same thing as the paragraph above and not sure how necessary the second sentence is but i'm fine to keep it if you feel it's necessary

leon-xd (commented):
good changes

tests/wdk-sys-tests/Cargo.toml
Alan632: @wmmc88 and I chatted about this offline and I think we are good in this scenario. `wdk-sys` dependent crates (with a WDK config present) will build to `lib` targets because of our use of the modifier `-bundle` with the link directive which defers including object files until final linking.
If there is no WDK config present in the metadata it will still build, as long as usages of the WDK api's are cfg-gated appropriately like in the wdk crate (since `wdk-sys` would not generate the function declarations w/o a config present anyway).

crates/wdk-build/src/lib.rs
```diff
@@ -2050,6 +2207,264 @@ mod tests {
...
+                LinkDirective::new("Foo", LinkKind::Dylib).render(),
+                format!(
+                    "#[cfg({NOT_TEST_CFG})]\n#[link(name = \"Foo\", kind = \"dylib\")]\nunsafe \
+                     extern \"C\" {{}}\n"
+                )
+            );
+        }
+
+        #[test]
+        fn render_static_directive_includes_no_bundle_modifier() {
```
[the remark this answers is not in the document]
wmmc88: nit: based on [the docs](https://doc.rust-lang.org/rustc/command-line-arguments.html?highlight=bundle#-l-link-the-generated-crate-to-a-native-library), the verbiage here is that we expect the bundle modifier is disabled:

> Modifiers are specified as a comma-delimited string with each modifier prefixed with either a + or - to indicate that the modifier is enabled or disabled, respectively. 

So really, this test is "disables bundle modifier"

crates/wdk-build/src/lib.rs
Alan632: I can see how it's redundant, removed that paragraph.

tests/mixed-package-kmdf-workspace/crates/driver/src/lib.rs
```diff
@@ -5,23 +5,35 @@
 //!
 //! This is a sample KMDF driver that demonstrates how to use the crates in
 //! windows-driver-rs to create a skeleton of a kmdf driver.
+//!
```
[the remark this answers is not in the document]
Alan632: Removed the trailing whitespace.

crates/wdk-build/src/lib.rs
```diff
@@ -2050,6 +2207,264 @@ mod tests {
...
+                    target_umdf_version_minor: 15,
+                    minimum_umdf_version_minor: None,
+                }),
+            );
+
+            assert_eq!(lib_names(&config.base_libraries()), ["OneCoreUAP"]);
+        }
+
+        #[test]
+        fn base_directives_are_all_static_no_bundle() {
```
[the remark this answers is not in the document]
Alan632: Renamed test.

crates/wdk-build/src/lib.rs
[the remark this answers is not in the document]
Alan632: Renamed the test to `render_static_directive_disables_bundle` dropping the `modifier` notation as I think that is what caused the initial confusion. This one explicitly states that the `bundle` action is disabled.

crates/wdk-build/src/lib.rs
wmmc88: `bindgen_library_link_raw_lines` is no different than the existing `bindgen_header_contents` in this respect. It is also more useful than for just `wdk-sys` as people can use `wdk-build` for their own binding generation crates (ex. if wdk-sys doesnt currently support their scenario)

crates/wdk-build/src/lib.rs
```diff
@@ -1057,6 +1057,116 @@ impl Config {
...
 
+    /// Returns the formatted `#[link]` raw lines for the given [`ApiSubset`],
+    /// ready to be passed to bindgen's `raw_line`.
+    ///
+    /// Each emitted directive is gated behind
+    /// `#[cfg(not(any(test, feature = "test-stubs")))]`, evaluated in the crate
+    /// that compiles the generated bindings (`wdk-sys`). The directives are
+    /// therefore suppressed when `wdk-sys` is built with its `test-stubs`
+    /// feature, or when `wdk-sys` is itself compiled as a test target (e.g.
+    /// `cargo test -p wdk-sys`).
```
wmmc88: Please audit all your comments. This is an implementation detail and only one possible use of this function

Alan632: Reviewed all comments and documentation, I made them more succinct and removed implementation details!

tests/wdk-sys-tests/src/lib.rs
```diff
@@ -1,2 +1,25 @@
...
+//! `cargo test` compiles and links each doctest as its own user-mode binary. It
+//! links against `wdk-sys`, which is pulled in here with the `test-stubs`
+//! feature (see this crate's dependencies). That feature cfg-suppresses the
+//! generated WDK `#[link]` directives, so the doctest links and runs without
+//! pulling kernel-mode libraries into it.
+//!
+//! Like the driver unit test in `tests/mixed-package-kmdf-workspace`, it
+//! references a bindgen generated WDM type (`DRIVER_OBJECT`) via a const
+//! `size_of` — so the doctest actually links `wdk-sys`'s generated bindings
+//!  while touching only a type (never a KM function
```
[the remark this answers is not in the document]
Alan632: fixed and outdated

crates/wdk-build/src/lib.rs
```diff
@@ -2050,6 +2206,267 @@ mod tests {
...
+                LinkDirective::new("Foo", LinkKind::Dylib).render(),
+                format!(
+                    "#[cfg({NOT_TEST_CFG})]\n#[link(name = \"Foo\", kind = \"dylib\")]\nunsafe \
+                     extern \"C\" {{}}"
+                )
+            );
+        }
+
+        #[test]
+        fn render_static_directive_includes_no_bundle_modifier() {
```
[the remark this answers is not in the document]
Alan632: test has been renamed and removed `modifier` wording

crates/wdk-build/src/lib.rs
```diff
@@ -2050,6 +2206,267 @@ mod tests {
...
+                    target_umdf_version_minor: 15,
+                    minimum_umdf_version_minor: None,
+                }),
+            );
+
+            assert_eq!(lib_names(&config.base_libraries()), ["OneCoreUAP"]);
+        }
+
+        #[test]
+        fn base_directives_are_all_static_no_bundle() {
```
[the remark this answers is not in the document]
Alan632: test has been renamed

crates/wdk-build/src/lib.rs
Alan632: Reviewed all documentation and have significantly cut down on extraneous implementation details so this shouldn't be an issue anymore.

ref: https://github.com/microsoft/windows-drivers-rs/pull/685#discussion_r3617844044

crates/wdk-build/src/lib.rs
wmmc88: @leon-xd The original comment is not a doc comment because it is an implementation detail. Implementation details typically shouldn't be in the doc comment since they're not relevant to people reading the rustdoc. We should view rustdoc comments on public apis as documenting what it is with the target audience being people who would use that API (without knowledge of internals). 

There's really good docs on this here: https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components

IMO the NI WDK stuff and specific file names is not relevant to the consumer of this API and is also likely to become out of sync, but if you wanted to include that in the rustdoc, it should be rewritten with the audience in mind.

Alan632: I went ahead and moved that portion of the doc comment back into the function. I also refined the doc comment, trying to focus on behavior of the function.

crates/wdk-build/src/lib.rs
```diff
@@ -1108,27 +1091,28 @@ impl Config {
...
     ///
     /// TODO: Once [link-arg-attribute](https://doc.rust-lang.org/unstable-book/language-features/link-arg-attribute.html)
     /// stabilizes, the `cargo::rustc-cdylib-link-arg=*` directives emitted by
     /// [`wdk_build::Config::configure_binary_build`] will be moved here too.
-    #[tracing::instrument(level = "trace")]
+    ///
+    /// # Returns
+    ///
+    /// A [`Vec`] of [`LinkDirective`]s for the [`ApiSubset`]'s variant
+    /// [`ApiSubset::Base`].
```
wmmc88: Why is it saying Returns in two different section? this is not consistent with the other sections. also, i dont think # Returns is a standard rustdoc section.

tests/mixed-package-kmdf-workspace/crates/driver/src/lib.rs
```diff
@@ -53,6 +57,7 @@ static GLOBAL_ALLOCATOR: WdkAllocator = WdkAllocator;
 /// Function is unsafe since it dereferences raw pointers passed to it from WDF
 // SAFETY: "DriverEntry" is the required symbol name for Windows driver entry points.
 // No other function in this compilation unit exports this name, preventing symbol conflicts.
+#[cfg(not(test))]
```
wmmc88: Why did this file add a bunch of these? This test package should work under test

crates/wdk-build/src/lib.rs
```diff
@@ -1057,6 +1057,111 @@ impl Config {
...
+    /// Returns a [`Vec`] of [`LinkDirective`]s for the [`ApiSubset`]'s variant
+    /// [`ApiSubset::Hid`].
+    ///
+    /// WDM/KMDF drivers link `VhfKm`, while UMDF drivers link `VhfUm`.
+    fn hid_libraries(&self) -> Vec<LinkDirective> {
+        match &self.driver_config {
+            DriverConfig::Wdm | DriverConfig::Kmdf(_) => {
+                vec![LinkDirective::new("VhfKm", LinkKind::Static)]
+            }
+            DriverConfig::Umdf(_) => vec![LinkDirective::new("VhfUm", LinkKind::Dylib)],
```
wmmc88: Why is this not statically linked?

crates/wdk-build/src/lib.rs
```diff
@@ -1057,6 +1057,111 @@ impl Config {
             .collect())
     }
 
+    /// Returns the formatted `#[link]` directives (as raw Rust source) for the
+    /// given [`ApiSubset`], or `None` if the [`ApiSubset`] does not define any
+    /// libraries to link.
+    ///
+    /// Each emitted directive is gated behind `#[cfg(not(any(test, feature =
+    /// "test-stubs")))]`.
```
wmmc88: per our discussion, change this to a separate `no-link` feature (and have `test-stubs` automatically enable `no-link`)

tests/mixed-package-kmdf-workspace/crates/driver/src/lib.rs
```diff
@@ -156,6 +161,7 @@ pub unsafe extern "system" fn driver_entry(
     wdf_driver_create_ntstatus
 }
 
+#[cfg(not(test))]
```
wmmc88: this should not need a `#[cfg(not(test))]`

tests/mixed-package-kmdf-workspace/crates/driver/src/lib.rs
```diff
@@ -53,6 +57,7 @@ static GLOBAL_ALLOCATOR: WdkAllocator = WdkAllocator;
 /// Function is unsafe since it dereferences raw pointers passed to it from WDF
 // SAFETY: "DriverEntry" is the required symbol name for Windows driver entry points.
 // No other function in this compilation unit exports this name, preventing symbol conflicts.
+#[cfg(not(test))]
 #[unsafe(export_name = "DriverEntry")] // WDF expects a symbol with the name DriverEntry
 pub unsafe extern "system" fn driver_entry(
```
wmmc88: You should just `cfg` the `export_name` instead of the whole function if you need to enable `test-stubs`

crates/wdk-sys/Cargo.toml
```diff
@@ -27,7 +27,9 @@ storage = []
 usb = []
 
 nightly = ["wdk-build/nightly", "wdk-macros/nightly"]
-test-stubs = []
+test-stubs = ["no-link"]
+
+no-link = []
```
wmmc88: when you add the new test coverage for tests in the example crates, make sure it gets run in CI

crates/wdk-build/src/lib.rs
Alan632: Refactored comment and removed the extra '# Returns' section!

tests/mixed-package-kmdf-workspace/crates/driver/src/lib.rs
Alan632: Removed extra unnecessary cfg gates

crates/wdk-build/src/lib.rs
Alan632: I removed the LinkKind type and all LinkDirectives now should be static

crates/wdk-build/src/lib.rs
Alan632: Changed to `omit-wdk-libs` (b/c I was hitting issues with the following lint `-D clippy::negative-feature-names`) and enabling `test-stubs` also enables the `omit-wdk-libs` feature!

tests/mixed-package-kmdf-workspace/crates/driver/src/lib.rs
Alan632: Removed the cfg gate

tests/mixed-package-kmdf-workspace/crates/driver/src/lib.rs
Alan632: Applied a `cfg_attr` to the export_name instead of blanket cfg gating the entire function

crates/wdk-sys/Cargo.toml
Alan632: Added another step to test.yaml to enable running the test coverage in the top level `examples/` folder

examples/sample-kmdf-driver/src/lib.rs
```diff
@@ -53,7 +49,7 @@ static GLOBAL_ALLOCATOR: WdkAllocator = WdkAllocator;
 /// Function is unsafe since it dereferences raw pointers passed to it from WDF
 // SAFETY: "DriverEntry" is the required symbol name for Windows driver entry points.
 // No other function in this compilation unit exports this name, preventing symbol conflicts.
-#[unsafe(export_name = "DriverEntry")] // WDF expects a symbol with the name DriverEntry
+#[cfg_attr(not(test), unsafe(export_name = "DriverEntry"))] // WDF expects a symbol with the name DriverEntry
```
wmmc88, line 54: Add a comment saying this is temporary to avoid clashing with `test-stubs` definition of `DriverEntry`

examples/sample-kmdf-driver/src/lib.rs
```diff
@@ -189,3 +184,14 @@ extern "C" fn driver_exit(_driver: *mut DRIVER_OBJECT) {
     println!("Goodbye World!");
     println!("Driver Exit Complete!");
 }
+
+#[cfg(test)]
+mod tests {
+
+    #[test]
+    fn test_driver_exit() {
```
wmmc88, line 196: leave a comment explaining this dummy test is to ensure test targets compile correctly without linking in the WDK libraries

examples/sample-umdf-driver/src/lib.rs
wmmc88, line 1: same comments as on KMDF driver about cfg_attr(not(test)) and the dummy test

examples/sample-wdm-driver/src/lib.rs
wmmc88, line 1: same comments as on KMDF driver about cfg_attr(not(test)) and the dummy test

crates/wdk-sys/Cargo.toml
```diff
@@ -27,7 +27,9 @@ storage = []
 usb = []
 
 nightly = ["wdk-build/nightly", "wdk-macros/nightly"]
-test-stubs = []
+test-stubs = ["omit-wdk-libs"]
+
+omit-wdk-libs = []
```
wmmc88: change this to something like `suppress-wdk-linking`. `omit-wdk-libs` isn't very clear (from the perspective of someone looking at the feature list and trying to figure out what to enable)

wmmc88 (commented):
minor nits, but otherwise good to go. Also create a bug documenting the `all-targets` of cdylib defect, and assign it to me

examples/sample-kmdf-driver/src/lib.rs
Alan632: Done!

examples/sample-kmdf-driver/src/lib.rs
Alan632: Added a comment for the test

examples/sample-umdf-driver/src/lib.rs
Alan632: Done!

examples/sample-wdm-driver/src/lib.rs
Alan632: Done!

crates/wdk-sys/Cargo.toml
Alan632: Changed the feature name to `suppress-wdk-linking`.

wmmc88 (dismissed):

crates/wdk-build/src/lib.rs
```diff
@@ -1057,26 +1057,12 @@ impl Config {
             .collect())
     }
 
-    /// Returns the formatted `#[link]` raw lines for the given [`ApiSubset`],
-    /// ready to be passed to bindgen's `raw_line`.
+    /// Returns the formatted `#[link]` directives (as raw Rust source) for the
```
leon-xd: nit get rid of parentheses here

leon-xd (dismissed):
great work :)

leon-xd (approved):

crates/wdk-build/src/lib.rs
Alan632: fixed!

wmmc88 (approved):

crates/wdk-build/src/lib.rs
```diff
@@ -1495,6 +1572,43 @@ pub fn configure_wdk_binary_build() -> Result<(), ConfigError> {
...
+        format!(
+            r#"#[cfg({cfg})]
+#[link(name = "{name}", kind = "static", modifiers = "-bundle")]
+unsafe extern "C" {{}}
+"#,
+            cfg = NOT_TEST_CFG,
+            name = self.name,
+        )
+    }
+}
```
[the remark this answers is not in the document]
Alan632: There are no leading spaces, it is a raw string literal.

crates/wdk-build/src/lib.rs
```diff
@@ -1495,6 +1572,43 @@ pub fn configure_wdk_binary_build() -> Result<(), ConfigError> {
...
+    fn render(&self) -> String {
+        format!(
+            r#"#[cfg({cfg})]
+#[link(name = "{name}", kind = "static", modifiers = "-bundle")]
+unsafe extern "C" {{}}
+"#,
+            cfg = NOT_TEST_CFG,
+            name = self.name,
+        )
+    }
```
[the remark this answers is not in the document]
Alan632: test passes

examples/sample-wdm-driver/src/lib.rs
```diff
@@ -6,7 +6,8 @@
 //! This is a sample WDM driver that demonstrates how to use the crates in
 //! windows-driver-rs to create a skeleton of a WDM driver.
 
-#![no_std]
+#![cfg_attr(not(test), no_std)]
+
 extern crate alloc;
```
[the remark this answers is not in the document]
Alan632: test_stubs.rs provides the necessary stubs, and this driver's `DriverEntry` export_name is cfg gated

examples/sample-kmdf-driver/src/lib.rs
```diff
@@ -6,7 +6,7 @@
 //! This is a sample KMDF driver that demonstrates how to use the crates in
 //! windows-driver-rs to create a skeleton of a kmdf driver.
 
-#![no_std]
+#![cfg_attr(not(test), no_std)]
 
 extern crate alloc;
```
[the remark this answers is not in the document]
Alan632: see comment: https://github.com/microsoft/windows-drivers-rs/pull/685#discussion_r3737865414

PR #691, opened 2026-06-25
ci: set CARGO_NET_RETRY=10 for Package Examples step

gurry:
Raises `CARGO_NET_RETRY` to `10` for the **Package Examples (via Cargo Make)** step only, to mitigate intermittent crates.io HTTP/2 connection-reset failures (libcurl `[16]`/`[55]`/`[56]`) observed during that step. Root cause: transport-layer TCP reset of the multiplexed HTTP/2 connection to crates.io's Fastly CDN; more retries lets cargo recover from a mid-transfer reset.

No other step or workflow is modified.

Opened as draft to monitor several CI runs; will mark ready once a few reruns pass without failure.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -100,6 +100,8 @@ jobs:
         run: cargo make --cwd ./examples build +${{ matrix.rust_toolchain }} --locked --profile ${{ matrix.cargo_profile }} --target ${{ matrix.target_triple.name }}
 
       - name: Package Examples (via Cargo Make)
+        env:
+          CARGO_NET_RETRY: "10"
         run: cargo make --cwd ./examples package-driver-flow +${{ matrix.rust_toolchain }} --locked --profile ${{ matrix.cargo_profile }} --target ${{ matrix.target_triple.name }}
 
       - name: Build Tests (via Cargo Make)


gurry:
Update from diagnostics (PR #688 branch): **raising CARGO_NET_RETRY is not a reliable fix.** This very PR (retry=10) still failed 4/48 legs at *Package Examples*, with [16] Error in the HTTP2 framing layer raised during a rust-script-spawned index resolve — and cargo logged **zero** `spurious network error` retries, i.e. the index/resolve-phase [16] does not enter cargo's retry wrapper.

Frame-level root cause (curl `lib/http2.c`): the crates.io Fastly CDN sends a graceful **HTTP/2 GOAWAY (error=0 / NO_ERROR)** to drain a reused connection that cargo has loaded with a large burst of multiplexed streams. Streams past the GOAWAY `last_stream_id` with 0 bytes received surface as `CURLE_HTTP2 [16]`; the send/recv face is `[55]/56`. One connection death cascades to ~80-140 in-flight streams.

A/B on the 688 branch (`CARGO_HTTP_MULTIPLEXING=false`) eliminated the failures: multiplexing ON = 1 fail / 16 legs, OFF = 0 fail / 16 legs at 2x exposure. Recommended mitigation is **`CARGO_HTTP_MULTIPLEXING=false`** (optionally plus `CARGO_NET_RETRY>=3`) rather than retries alone. Keeping this PR as draft.

PR #699, opened 2026-07-07
feat(cargo-wdk): add `--signtool-args` passthrough to customize driver signing

### LINKED ISSUE #605: Add support to pass custom certificate details for signing through cargo-wdk
At present, the default behaviour of cargo‑wdk is to automatically generate a self‑signed test certificate for signing operations, which can be used during the local development and test cycle. If a user wishes to supply their own certificate for signing, cargo‑wdk is currently restricted to accepting certificates with the subject name WDRLocalTestCert and located under WDRTestCertStore, and it throws the following error even when the user’s certificate is installed under WDRTestCertStore:

```
 INFO Signing test_driver using signtool
SignTool Error: No certificates were found that met all the given criteria
```

To address this limitation, we need to provide a way for cargo‑wdk to accept a custom certificate from a file path or to pick a certificate with a user‑supplied subject name from a chosen store, and allow that certificate to be used for signing.

**Related Reference**
There is already a separate issue to support customization for rust-driver-makefile.toml https://github.com/microsoft/windows-drivers-rs/issues/92

svasista-ms:
Adds `--signtool-args` passthrough so driver signing can be customized (certificate selection, digest algorithm, timestamping, extra file operands).

When `--signtool-args` is **omitted**, `cargo-wdk` signs with the auto-generated WDR test certificate and default switches.
When `--signtool-args` is **provided**, the caller owns the full `signtool sign` option set except the `sign` verb and the trailing file operand (spplied by `cargo-wdk`). Supplying it with `--sign-mode=off` is rejected.

### ⚠️ Behavior change
Packaging removes any existing `<target>/<profile>/<name>_package` folder at the start of each build and recreates it, so stale signing artifacts (e.g. a `WDRLocalTestCert.cer` from a previous `--sign-mode=test` build) don't persist across rebuilds.

### Screenshots

image

Resolves #605


crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -231,55 +230,105 @@ impl<'a> PackageTask<'a> {
...
-        self.run_signtool_sign(
-            &self.dest_cat_file_path,
-            WDR_TEST_CERT_STORE,
-            WDR_LOCAL_TEST_CERT,
-        )?;
+        // Tokenize the raw passthrough string once (this layer owns the split).
+        // `None` selects the auto-generated WDR test-cert flow; `Some` means the
+        // caller owns certificate selection, so cargo-wdk does not generate a
+        // test certificate.
+        let custom_args = signtool_args.as_deref().map(split_signtool_args);
```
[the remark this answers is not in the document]
svasista-ms: Fixed ✅

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -231,55 +230,105 @@ impl<'a> PackageTask<'a> {
...
+        debug!("Creating staging directory");
+        self.fs.create_dir(&self.stage_root_folder)?;
+        Ok(())
+    }
+
+    /// Assembles the final package folder by atomically replacing any existing
+    /// one with the fully-built staging directory. Because this runs only after
+    /// every packaging step succeeds, the package folder never contains partial
+    /// or unsigned output, and stale artifacts (e.g. a `.cer` from a previous
+    /// `--sign-mode=test` build) are removed when the mode changes.
```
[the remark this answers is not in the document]
svasista-ms: Fixed docs ✅

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -379,6 +420,289 @@ fn sign_mode_off_with_verify_signature_is_rejected() {
...
+        let out = Command::new("powershell")
+            // Clear PSModulePath so Windows PowerShell uses its default module path.
+            // When the test runner is launched from PowerShell 7 (pwsh), the child
+            // Windows PowerShell inherits pwsh's PSModulePath and fails to load
+            // `Microsoft.PowerShell.Security` (which provides Get-AuthenticodeSignature).
+            .env_remove("PSModulePath")
+            .args(["-NoProfile", "-NonInteractive", "-Command", &script])
+            .output()
+            .expect("failed to run Get-AuthenticodeSignature");
+        let subject = String::from_utf8_lossy(&out.stdout).trim().to_string();
```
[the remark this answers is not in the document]
svasista-ms: Fixed ✅

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -839,6 +946,181 @@ mod tests {
...
+        PackageTask::new(params, wdk_build, command_exec, fs)
+    }
+
+    // Returns a mocked `CommandExec` that expects a single `signtool` invocation
+    // with exactly the provided argument vector.
+    fn expect_signtool_args(expected: Vec<String>) -> CommandExec {
+        let mut command_exec = CommandExec::default();
+        command_exec
+            .expect_run()
+            .withf(move |command, args, _env, _cwd| command == "signtool" && args == expected)
```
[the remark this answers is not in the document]
svasista-ms: Fixed ✅

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -231,57 +224,68 @@ impl<'a> PackageTask<'a> {
...
-            WDR_LOCAL_TEST_CERT,
-        )?;
-        self.run_signtool_sign(
-            &self.dest_cat_file_path,
-            WDR_TEST_CERT_STORE,
-            WDR_LOCAL_TEST_CERT,
-        )?;
-
-        if verify_signature {
+        let custom_args = signtool_args.as_deref().map(split_signtool_args);
```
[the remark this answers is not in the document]
svasista-ms: Fixed ✅

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -503,21 +519,32 @@ impl<'a> PackageTask<'a> {
...
+                // Caller owns the full option set.
+                args.extend(extra.iter().cloned());
+            }
+        }
+
+        // File operand (must be last).
+        args.push(file_path);
+
+        let arg_refs: Vec<&str> = args.iter().map(String::as_str).collect();
+        if let Err(e) = self.command_exec.run("signtool", &arg_refs, None, None) {
```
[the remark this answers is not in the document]
svasista-ms: Fixed by adding an optional `hide: bool` paramter to the `CommandExec::run` method. This hides args supplied to `signtool` invocation. This may hamper troubleshooting in case of `signtool` failures because the args are hidden. However, if the actual command line args that were passed to `cargo-wdk` via `--signtool-args` is known debugging should be easier.

crates/cargo-wdk/README.md
```diff
@@ -131,3 +151,15 @@ If the `--verify-signature` flag is provided, the signatures are verified after
...
+
+    ```pwsh
+    cargo wdk build --signtool-args "/s MyStore /n MyCert /fd SHA256"
+    ```
+
+- To test-sign with a PFX file whose password is supplied inline, run:
+
+    ```pwsh
+    cargo wdk build --signtool-args "/f C:\certs\my.pfx /p <password> /fd SHA256"
+    ```
```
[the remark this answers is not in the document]
svasista-ms: Added a note regarding the same

crates/cargo-wdk/src/providers/exec.rs
```diff
@@ -59,16 +61,16 @@ impl CommandExec {
             .stdout(Stdio::piped())
             .spawn()
             .and_then(std::process::Child::wait_with_output)
-            .map_err(|e| CommandError::from_io_error(command, args, e))?;
+            .map_err(|e| CommandError::from_io_error(command, log_args, e))?;
 
         if !output.status.success() {
-            return Err(CommandError::from_output(command, args, &output));
+            return Err(CommandError::from_output(command, log_args, &output));
         }
```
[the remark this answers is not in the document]
svasista-ms: `CommandExec::run` only pipes `stdout`; `stderr` is left inherited, so the child process (signtool/cargo/etc.) writes its failure diagnostics directly to cargo-wdk's `stderr`. So, they're visible in the shell/captured by CI.

crates/cargo-wdk/src/cli.rs
```diff
@@ -118,23 +130,98 @@ pub struct BuildArgs {
...
+                    .map(|parsed| parsed.0)
+                    .unwrap_or_default(),
+            }),
+        }
+    }
+}
+
+/// Builds a `clap::Error` with the given message, rendered with the standard
+/// `cargo wdk build` usage for a consistent CLI experience.
+fn build_error(message: impl std::fmt::Display) -> clap::Error {
```
krishnakumar4a4: nit: we could inline this at call site, no need of a separate function

crates/cargo-wdk/src/cli.rs
```diff
@@ -118,23 +130,98 @@ pub struct BuildArgs {
...
+    Cli::command().error(ErrorKind::ArgumentConflict, message)
+}
+
+/// Arguments to `signtool sign` for signing the driver binary and catalog file.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SigntoolArgs(pub Vec<String>);
```
krishnakumar4a4: Move this declaration to the top to keep all arg type declarations at one place.

crates/cargo-wdk/src/cli.rs
```diff
@@ -304,25 +402,119 @@ mod tests {
...
 
-        let result = cli.run();
-        assert!(result.is_err());
+    #[test]
+    fn build_rejects_signtool_args_with_sign_mode_off() {
+        let args = parse_build_args(&["--sign-mode", "off", "--signtool-args", "/fd SHA256"])
+            .expect("args parse");
+        let err = SignMode::try_from(&args).expect_err("should be rejected");
+        assert!(
+            err.to_string().contains("`--sign-mode=off`"),
```
krishnakumar4a4: 1. use the exact message for assertion

```
`--signtool-args` cannot be used with `--sign-mode=off`.
```

2. Can we not match exactly and avoid using contains?

crates/cargo-wdk/src/cli.rs
```diff
@@ -304,25 +402,119 @@ mod tests {
...
+            "unexpected error: {err}"
+        );
+    }
+
+    #[test]
+    fn build_off_mode_maps_to_off() {
```
krishnakumar4a4:
```suggestion
    fn build_sign_mode_off_maps_correctly() {
```

crates/cargo-wdk/src/cli.rs
```diff
@@ -304,25 +402,119 @@ mod tests {
...
+            }
+        );
+    }
+
+    #[test]
+    fn build_rejects_unterminated_quote_in_signtool_args() {
+        let err = parse_build_args(&["--signtool-args", "/n \"CN=Contoso"])
+            .expect_err("unterminated quote should be rejected");
+        assert!(
+            err.to_string().contains("unterminated"),
```
krishnakumar4a4: Try and use exact message assertion

Similar suggestion as this: https://github.com/microsoft/windows-drivers-rs/pull/699/changes#r3579259207

crates/cargo-wdk/src/cli.rs
```diff
@@ -304,25 +402,119 @@ mod tests {
...
+                    verify_signature: false,
+                    signtool_args: Vec::new(),
+                },
+                "value {value:?} should map to no signtool args"
+            );
+        }
+    }
+
+    #[test]
+    fn build_signtool_args_with_verify_signature_maps_both() {
```
krishnakumar4a4:
```suggestion
    fn build_verify_signature_works_with_signtool_args() {
```

crates/cargo-wdk/src/cli.rs
```diff
@@ -304,25 +402,119 @@ mod tests {
 
     #[test]
     fn build_rejects_verify_signature_when_sign_mode_is_off() {
```
krishnakumar4a4: As we have multiple tests for build sub command, We should move all the build tests to a submodule here and avoid using `build` as prefix for each test.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -567,21 +603,82 @@ impl<'a> PackageTask<'a> {
...
+                in_arg = true;
+            }
+        }
+    }
+
+    if in_arg {
+        args.push(current);
+    }
+
+    args
```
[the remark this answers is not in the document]
svasista-ms: Fixed ✅

krishnakumar4a4 (changes requested):

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -359,6 +359,47 @@ fn kmdf_driver_builds_successfully_with_sign_mode_off() {
     });
 }
 
+/// Regression test for issue #660: because the package folder is assembled
+/// fresh (the tools run in a staging dir and the folder is replaced last),
+/// switching from `--sign-mode=test` to `--sign-mode=off` on a subsequent build
+/// (without cleaning) must drop the stale `WDRLocalTestCert.cer` from the
+/// package folder.
```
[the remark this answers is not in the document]
svasista-ms: Removed unnecessary comment

crates/cargo-wdk/src/cli.rs
```diff
@@ -118,23 +130,98 @@ pub struct BuildArgs {
...
+                if args.verify_signature {
+                    return Err(build_error(
+                        "`--verify-signature` cannot be used with `--sign-mode=off`.",
+                    ));
+                }
+                if args.signtool_args.is_some() {
+                    return Err(build_error(
+                        "`--signtool-args` cannot be used with `--sign-mode=off`.",
+                    ));
+                }
```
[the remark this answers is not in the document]
svasista-ms: Using `signtool-args` with `sign-mode = off` is an invalid combination that we want to catch at the CLI layer. This is intentional. To ensure empty `signtool-args` and `sign-mode = off` is still rejected, I added another test `build_rejects_empty_signtool_args_with_sign_mode_off`

crates/cargo-wdk/src/cli.rs
svasista-ms: Yes, it was inline initially. But I added the function because `ArgumentConflict` error can be constructed from multiple places (2 in `try_from` for `SignMode`). Once args for all tools are added we could use `anyhow` to construct CLI errors.

crates/cargo-wdk/src/cli.rs
svasista-ms: Done ✅ , moved to a separate mod

crates/cargo-wdk/src/cli.rs
svasista-ms: Fixed ✅

crates/cargo-wdk/src/cli.rs
svasista-ms: Fixed ✅

crates/cargo-wdk/src/cli.rs
svasista-ms: Fixed ✅

crates/cargo-wdk/src/cli.rs
svasista-ms: Fixed ✅, using the full message for assertion 👍

crates/cargo-wdk/src/cli.rs
svasista-ms: Done✅

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -839,6 +874,160 @@ mod tests {
...
+        task.run_signtool_sign(Path::new("C:/pkg/driver.sys"), &signtool_args)
+            .expect("signing should succeed");
+    }
+
+    #[test]
+    fn sign_and_verify_fails_when_signtool_fails() {
+        // Custom signtool args avoid the makecert/certmgr path. When signtool
+        // fails, sign_and_verify propagates the error. There is no cleanup: the
+        // package folder is only assembled (in `run`) after signing succeeds,
+        // so a failure never leaves unsigned output behind.
```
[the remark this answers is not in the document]
svasista-ms: Removed the unnecessary comment

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -839,6 +874,160 @@ mod tests {
...
+
+    // Returns a mocked `CommandExec` that expects a single `signtool` invocation
+    // with exactly the provided argument vector.
+    fn expect_signtool_args(expected: Vec<String>) -> CommandExec {
+        let mut command_exec = CommandExec::default();
+        command_exec
+            .expect_run()
+            .withf(move |command, args, _hide_args, _env, _cwd| {
+                command == "signtool" && args == expected
+            })
```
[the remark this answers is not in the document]
svasista-ms: This compiles and passes locally and on [CI](https://github.com/microsoft/windows-drivers-rs/actions/runs/29393990097). `expected` is already `Vec<&str>` and `Vec<&str> vs &[&str]` comparison works without extra allocation.

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -322,61 +322,368 @@ mod kmdf_driver_with_target_override {
...
+                .join("extra_to_sign.dll");
+            fs::copy(&built, &extra).expect("copy extra file");
+            strip_signature(&extra);
+            assert!(
+                authenticode_signer_subject(&extra).is_none(),
+                "precondition: extra file should be unsigned after strip"
+            );
+
+            // Pass the extra file as an additional operand inside --signtool-args.
+            let args_with_extra = format!("{common_args} {}", extra.to_string_lossy());
```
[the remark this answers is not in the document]
svasista-ms: Fixed ✅

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -503,21 +502,32 @@ impl<'a> PackageTask<'a> {
...
+            args.extend(signtool_args.iter().cloned());
+        }
+
+        // File operand (must be last).
+        args.push(file_path);
+
+        let arg_refs: Vec<&str> = args.iter().map(String::as_str).collect();
+        if let Err(e) = self
+            .command_exec
+            .run("signtool", &arg_refs, true, None, None)
```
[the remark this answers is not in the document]
svasista-ms: Fixed ✅

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -231,12 +218,14 @@ impl<'a> PackageTask<'a> {
...
         self.check_inx_exists()?;
-        debug!("Creating final package directory if it doesn't exist");
-        if !self.fs.exists(&self.dest_root_package_folder) {
-            self.fs.create_dir(&self.dest_root_package_folder)?;
+        if self.fs.exists(&self.dest_root_package_folder) {
+            debug!("Removing existing package folder");
+            self.fs.remove_dir_all(&self.dest_root_package_folder)?;
         }
+        debug!("Creating package folder");
+        self.fs.create_dir(&self.dest_root_package_folder)?;
```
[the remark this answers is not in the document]
svasista-ms: Updated PR description to match the behavior

crates/cargo-wdk/src/cli.rs
```diff
@@ -105,6 +109,18 @@ pub struct BuildArgs {
     /// Verify the signature
     #[arg(long)]
     pub verify_signature: bool,
+
+    /// Additional arguments to pass to `signtool sign` when signing the driver
+    /// binary and the catalog file, e.g.
+    /// `--signtool-args '/fd SHA512 /n "CN=WDRLocalTestCert, O=Foo"'`.
+    #[arg(
```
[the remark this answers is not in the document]
svasista-ms: This is intentional, to demonstrate the use of quotes within the `--signtool-args` argument

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -839,6 +874,156 @@ mod tests {
...
+
+    // Returns a mocked `CommandExec` that expects a single `signtool` invocation
+    // with exactly the provided argument vector.
+    fn expect_signtool_args(expected: Vec<String>) -> CommandExec {
+        let mut command_exec = CommandExec::default();
+        command_exec
+            .expect_run()
+            .withf(move |command, args, _hide_args, _env, _cwd| {
+                command == "signtool" && args == expected
+            })
```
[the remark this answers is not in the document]
svasista-ms: Not a concern

crates/cargo-wdk/README.md
```diff
@@ -106,6 +111,15 @@ If the `--verify-signature` flag is provided, the signatures are verified after
...
 `--verify-signature` cannot be combined with `--sign-mode=off` because if signing is off there is nothing to verify. Passing both will cause `build` to fail with an error.
 
+##### Customizing signtool arguments
+
+To sign with your own certificate or tweak any signing option, pass `--signtool-args` with a string of the arguments to forward to `signtool sign`.
+
+- When `--signtool-args` is **omitted**, cargo-wdk signs with the auto-generated WDR test certificate as described above.
+- When `--signtool-args` is **provided**, you own the full `signtool sign` option set (certificate selection, digest algorithm, etc.). cargo-wdk only wraps your arguments with the `sign` verb and the trailing file operand. If your arguments don't include a certificate selector (e.g. `/n`, `/s`, `/f`, `/sha1`), `signtool` auto-selects a code-signing certificate from your personal store and fails if none — or more than one — is found.
+
+`--signtool-args` applies only when signing is enabled; supplying it with `--sign-mode=off` is an error.
```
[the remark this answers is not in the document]
svasista-ms: Ideally additional files must not be passed via `signtool-args`. No need to add additional details since the packaging flow is known.

crates/cargo-wdk/src/cli.rs
```diff
@@ -118,25 +141,96 @@ pub struct BuildArgs {
...
+    }
+
+    if let Some(q) = quote {
+        return Err(format!(
+            "unterminated `{q}` quote in `--signtool-args`; make sure every quote is closed"
+        ));
+    }
+    if in_arg {
+        args.push(current);
+    }
```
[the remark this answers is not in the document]
svasista-ms: Fixed ✅

crates/cargo-wdk/src/cli.rs
```diff
@@ -118,25 +141,96 @@ pub struct BuildArgs {
...
+            None if c == '"' || c == '\'' => {
+                quote = Some(c);
+                in_arg = true;
+            }
+            None if c.is_whitespace() => {
+                if in_arg {
+                    args.push(std::mem::take(&mut current));
+                    in_arg = false;
+                }
+            }
```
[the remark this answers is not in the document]
svasista-ms: Fixed ✅

crates/cargo-wdk/src/actions/build/package_task.rs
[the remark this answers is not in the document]
svasista-ms: Not an issue. Verified locally and on CI.

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -322,61 +322,368 @@ mod kmdf_driver_with_target_override {
...
+                    "SHA256",
+                    "-eku",
+                    "1.3.6.1.5.5.7.3.3",
+                    "-ss",
+                    store,
+                    "-n",
+                    &subject,
+                ])
+                .output()
+                .expect("failed to run makecert");
```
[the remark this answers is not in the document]
svasista-ms: `makecert` with `-ss` options works fine without the positional parameter (for file operand). Verified locally by running `makecert` in WDK v10.0.28000 and v10.0.26100.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -475,26 +459,41 @@ impl<'a> PackageTask<'a> {
...
-    /// certificate store.
+    /// Signs the specified file using the `signtool` command.
+    ///
+    /// When `signtool_args` is empty, cargo-wdk signs with the auto-generated
+    /// WDR test certificate and its default switches:
+    ///
+    /// `sign /v /s WDRTestCertStore /n WDRLocalTestCert /fd SHA256 <file>`
+    ///
+    /// When `signtool_args` is non-empty, the caller owns the full signtool
+    /// option set; cargo-wdk only wraps their (already tokenized) arguments
```
```suggestion
    /// command line; cargo-wdk only wraps their (already tokenized) arguments
```

Better to say "command line" instead of "option set"

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -475,26 +459,41 @@ impl<'a> PackageTask<'a> {
...
+    /// with the `sign` verb and the trailing file operand:
```
```suggestion
    /// option set; cargo-wdk only prepends the `sign` verb to those arguments
    /// and adds the trailing file operand:
```
"Prepends" is clearer than "wraps"

Also this layer should is not meant to know that the args are originating from a CLI. Therefore it should not make any mention of arg parsing or tokenizing.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -574,14 +584,34 @@ impl<'a> PackageTask<'a> {
...
+        /// Tokenized `signtool sign` arguments.
+        ///
+        /// When empty, run `signtool sign` with the auto-generated WDR test
+        /// certificate and default switches. When non-empty, auto generation
+        /// is skipped and the caller owns the full signtool option set
+        /// (certificate selection, digest, etc.).
+        signtool_args: Vec<String>,
+    },
+}
+
```
gurry: This struct should not have moved from where it was. We could have just added the new fields to it without moving it. 

Unnecessary movements like this produce bigger diffs than needed. AI is prone to making such gratuitous changes and it is your responsibility to manually edit and clean up AI generated code before submitting a PR.

crates/cargo-wdk/src/providers/exec.rs
```diff
@@ -37,10 +37,12 @@ impl CommandExec {
         &self,
         command: &'a str,
         args: &'a [&'a str],
+        hide_args: bool,
         env_vars: Option<&'a HashMap<&'a str, &'a str>>,
         working_dir: Option<&'a Path>,
     ) -> Result<Output, CommandError> {
-        debug!("Running: {} {:?}", command, args);
+        let log_args: &[&str] = if hide_args { &["<hidden>"] } else { args };
```
gurry: Instead of passing `hide_args` from outside we should selectively hide the value passed to the `/p` arg (because `/p` is the only sensitive arg) with a logic this is confined entirely within this method.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -475,26 +459,41 @@ impl<'a> PackageTask<'a> {
...
+    ///
+    /// When `signtool_args` is non-empty, the caller owns the full signtool
+    /// option set; cargo-wdk only wraps their (already tokenized) arguments
+    /// with the `sign` verb and the trailing file operand:
+    ///
+    /// `sign <signtool_args...> <file>`
+    ///
+    /// Test signing does not timestamp (matching the WDK `MSBuild` `TestSign`
+    /// target); timestamping is a production-signing concern handled by a
+    /// follow-up.
```
gurry: Remove this. We will add it later when we enable the `prod` mode. AI is prone to adding slop like this. Really important to review and edit the code thoroughly before you open the PR

crates/cargo-wdk/src/actions/build/package_task.rs
svasista-ms: I moved it because it was incorrectly placed (earlier) in between the imports and the constants: [diff](https://github.com/microsoft/windows-drivers-rs/pull/699/changes/BASE..3e87ce053785e28aa450ecb08ce8deae3995d481#diff-a6887c609082ccee97f37483a4395a3d87c8ccfd5d1b4a7ea0d79c3884536037L33-L45
). I can move it above `PackageTaskParams` struct definition instead of bottom of the file.

gurry: In that case it would have moved just below this statement `const STAMPINF_VERSION_ENV_VAR: &str = "STAMPINF_VERSION";`. Instead it has been moved well past a whole bunch of functions. Looks like something caused by AI not you.

crates/cargo-wdk/src/actions/build/package_task.rs
svasista-ms: Fixed ✅

crates/cargo-wdk/src/actions/build/package_task.rs
svasista-ms: Fixed ✅

crates/cargo-wdk/src/actions/build/package_task.rs
svasista-ms: Oh, no this was me 😸 I moved it after `impl PackageTask` assuming that was the pattern (asked AI to do it 😅). But I fixed it ([diff](https://github.com/microsoft/windows-drivers-rs/pull/699/changes#diff-a6887c609082ccee97f37483a4395a3d87c8ccfd5d1b4a7ea0d79c3884536037R40-R59)) in the latest commit.

crates/cargo-wdk/src/providers/exec.rs
svasista-ms: @gurry, thank you discussing the approach for redaction. As we discussed internally, instead of `hide_args` param, I added a `run_with_redaction` to `impl CommandExec` that accepts a slice of indices that must be redacted from the logs along with the unit tests. This is much cleaner than adding another param.

crates/cargo-wdk/src/actions/build/package_task.rs
svasista-ms: Done ✅

crates/cargo-wdk/README.md
```diff
@@ -131,3 +143,17 @@ If the `--verify-signature` flag is provided, the signatures are verified after
...
+    cargo wdk build --signtool-args "/s MyStore /n MyCert /fd SHA256"
+    ```
+
+- To test-sign with a PFX file whose password is supplied inline, run:
+
+    ```pwsh
+    cargo wdk build --signtool-args "/f C:\certs\my.pfx /p <password> /fd SHA256"
+    ```
+
+    > **Warning:** Passing a password inline via `/p` exposes it to other processes on the machine (e.g. through `Get-Process`/task managers) and risks it ending up in logs. cargo-wdk omits the signtool arguments from its own command logging, but the shell history and process arguments are outside its control. Prefer importing the PFX into a certificate store and selecting it with `/n`/`/sha1`, or provide the password through a more secure mechanism, instead of putting it on the command line.
```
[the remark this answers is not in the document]
gurry: This warning is not needed since we're redacting the password in logs. Please remove it.

crates/cargo-wdk/src/cli.rs
```diff
@@ -145,23 +167,97 @@ pub struct BuildArgs {
...
-            (SignModeArg::Off, false) => Ok(SignMode::Off),
-            (SignModeArg::Test, verify_signature) => Ok(SignMode::Test { verify_signature }),
+/// Resolves a typed, fully-validated [`SignMode`] from the parsed build
+/// arguments. Rules that clap cannot express declaratively are enforced here
+/// and surfaced as `clap::Error` for consistent CLI UX.
+impl TryFrom<&BuildArgs> for SignMode {
+    type Error = clap::Error;
+
+    fn try_from(args: &BuildArgs) -> Result<Self, clap::Error> {
+        match args.sign_mode {
```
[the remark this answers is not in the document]
svasista-ms: No change required. `anyhow::Result` has 2 type parameters.

crates/cargo-wdk/src/cli.rs
```diff
@@ -330,29 +426,129 @@ mod tests {
...
-            "`--verify-signature` cannot be used with `--sign-mode=off`."
-        );
+        fn parse_build_args(extra: &[&str]) -> Result<BuildArgs, clap::Error> {
+            let mut command_line = vec!["cargo-wdk", "wdk", "build"];
+            command_line.extend_from_slice(extra);
+            match Cli::try_parse_from(command_line)?.sub_cmd {
+                Subcmd::Build(build_args) => std::result::Result::Ok(build_args),
+                _ => unreachable!("build subcommand was requested"),
+            }
+        }
```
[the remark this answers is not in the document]
svasista-ms: No change required. `anyhow::Result` has 2 type parameters.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -525,22 +534,46 @@ impl<'a> PackageTask<'a> {
...
+        // File operand (must be last).
+        args.push(file_path);
+
+        let arg_refs: Vec<&str> = args.iter().map(String::as_str).collect();
+        // Redact the value following a `/p` (password) switch.
+        let redaction_indices: Vec<usize> = arg_refs
+            .iter()
+            .enumerate()
+            .filter_map(|(i, arg)| (*arg == "/p" && i + 1 < arg_refs.len()).then_some(i + 1))
+            .collect();
```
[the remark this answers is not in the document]
gurry: @svasista-ms We should fix this as suggeted. Otherwise we will log the password if the user uses upper case.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -503,21 +512,41 @@ impl<'a> PackageTask<'a> {
...
+
+        if signtool_args.is_empty() {
+            // Default WDR test-cert switches.
+            args.push("/v".to_string());
+            args.push("/s".to_string());
+            args.push(WDR_TEST_CERT_STORE.to_string());
+            args.push("/n".to_string());
+            args.push(WDR_LOCAL_TEST_CERT.to_string());
+            args.push("/fd".to_string());
+            args.push("SHA256".to_string());
```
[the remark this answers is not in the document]
gurry: @svasista-ms this looks serious. Please look into it

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -254,34 +263,26 @@ impl<'a> PackageTask<'a> {
...
-        )?;
-
-        if verify_signature {
+        if signtool_args.is_empty() {
+            self.generate_certificate()?;
+            self.copy(&self.src_cert_file_path, &self.dest_cert_file_path)?;
+        }
+        self.run_signtool_sign(&self.dest_driver_binary_path, signtool_args)?;
+        self.run_signtool_sign(&self.dest_cat_file_path, signtool_args)?;
+        if *verify_signature {
```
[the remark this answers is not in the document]
svasista-ms: Fixed ✅

crates/cargo-wdk/README.md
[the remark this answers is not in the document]
svasista-ms: Removed ✅

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -860,84 +898,202 @@ mod tests {
...
-    fn assert_infverif_mode_flag(
-        driver_model: DriverConfig,
-        target_platform: TargetPlatform,
-        expected_mode_flag: &'static str,
-    ) {
-        let package_name = "driver";
-        let working_dir = PathBuf::from("C:/abs/driver");
-        let target_dir = PathBuf::from("C:/abs/driver/target/debug");
-        let arch = CpuArchitecture::Amd64;
+    mod signtool {
```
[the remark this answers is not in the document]
svasista-ms: Test added back ✅

crates/cargo-wdk/src/actions/build/package_task.rs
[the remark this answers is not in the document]
svasista-ms: Already addressed: https://github.com/svasista-ms/windows-drivers-rs/blob/2d349fd9414ed2a3d08ba6a2b2b7f4592322308a/crates/cargo-wdk/src/actions/build/package_task.rs#L564

crates/cargo-wdk/src/actions/build/package_task.rs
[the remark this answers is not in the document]
svasista-ms: Added the `/t` switch back in https://github.com/microsoft/windows-drivers-rs/pull/699/commits/c8fcabecc1b8d7b401daec028f95cc7463e3e9aa

crates/cargo-wdk/src/providers/exec.rs
```diff
@@ -40,7 +40,42 @@ impl CommandExec {
...
+    /// If any index in `redaction_indices` is out of bounds for `args`.
+    #[allow(clippy::extra_unused_lifetimes)]
+    pub fn run_with_redaction<'a>(
+        &self,
+        command: &'a str,
+        args: &'a [&'a str],
+        redaction_indices: &'a [usize],
+        env_vars: Option<&'a HashMap<&'a str, &'a str>>,
+        working_dir: Option<&'a Path>,
+    ) -> Result<Output, CommandError> {
```
gurry, line 63: Add these comment lines to the function's comments to explain why we have all those lifetime annotations just like the `run` method above:

```
// The `'a` lifetime is required by mockall's `#[automock]` to generate the
// mock impl
```

crates/cargo-wdk/src/providers/exec.rs
```diff
@@ -40,7 +40,42 @@ impl CommandExec {
...
+        assert!(
+            redaction_indices.iter().all(|&i| i < args.len()),
+            "redaction index out of bounds for {} argument(s): {redaction_indices:?}",
+            args.len()
+        );
```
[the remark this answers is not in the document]
gurry: It is fine. There are ways to eliminate this problem through the type system, but we don't want to over-engineer at the moment.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -525,22 +535,49 @@ impl<'a> PackageTask<'a> {
...
+            // Default WDR test-cert switches.
+            args.push("/v".to_string());
+            args.push("/s".to_string());
+            args.push(WDR_TEST_CERT_STORE.to_string());
+            args.push("/n".to_string());
+            args.push(WDR_LOCAL_TEST_CERT.to_string());
+            args.push("/t".to_string());
+            args.push(DEFAULT_TIMESTAMP_URL.to_string());
+            args.push("/fd".to_string());
+            args.push("SHA256".to_string());
```
gurry: It will be cleaner to move all thes `args.push` statements to `sign_and_verify`. It will mean args always come from outside in the `signtool_args` parameter and `run_signtool_sign` will just have to do `args.extend(signtool_args.iter().cloned())` regardless of whether the user supplied signtool args or not. It will make `run_signtool_sign` more general.

crates/cargo-wdk/src/cli.rs
```diff
@@ -145,23 +167,97 @@ pub struct BuildArgs {
...
-        match (self.sign_mode, self.verify_signature) {
-            (SignModeArg::Off, true) => Err(anyhow::anyhow!(
-                "`--verify-signature` cannot be used with `--sign-mode=off`."
-            )),
-            (SignModeArg::Off, false) => Ok(SignMode::Off),
-            (SignModeArg::Test, verify_signature) => Ok(SignMode::Test { verify_signature }),
+/// Resolves a typed, fully-validated [`SignMode`] from the parsed build
+/// arguments. Rules that clap cannot express declaratively are enforced here
+/// and surfaced as `clap::Error` for consistent CLI UX.
+impl TryFrom<&BuildArgs> for SignMode {
```
gurry: `TryFrom` is not appropriate here. It is used for conversion from one type to the other and we're not converting `BuildArgs` to `SignMode`.

A better option is to turn this into a method on `BuildArg` called `get_sign_mode()`

crates/cargo-wdk/src/cli.rs
gurry: Move `build_error` into the function in which it is used and declare it as a nested function. That is a good pattern to follow  whenever you need a function that is used only inside one function.

crates/cargo-wdk/src/cli.rs
```diff
@@ -145,23 +167,97 @@ pub struct BuildArgs {
...
+}
+
+/// `value_parser` for `--signtool-args`: tokenizes the raw string into
+/// individual `signtool` arguments.
+///
+/// Rules:
+/// - Whitespace separates arguments
+/// - Quoted spans (single or double quotes) are preserved as a single argument
+/// - Unterminated quotes are rejected with an error
+fn parse_passthrough_args(raw: &str) -> std::result::Result<SigntoolArgs, String> {
```
```suggestion
fn parse_passthrough_args(raw: &str) -> Result<SigntoolArgs, String> {
```
Replace with an import of `std::result::Result`. 

AI tends to add fully qualified paths. Make sure you clean them up.

crates/cargo-wdk/src/cli.rs
```diff
@@ -145,23 +167,97 @@ pub struct BuildArgs {
...
+    if let Some(q) = quote {
+        return Err(format!(
+            "unterminated `{q}` quote in `--signtool-args`; make sure every quote is closed"
+        ));
+    }
+    if in_arg && !current.is_empty() {
+        args.push(current);
+    }
+
+    std::result::Result::Ok(SigntoolArgs(args))
```
```suggestion
  Ok(SigntoolArgs(args))
```

Remember to remove fully qualified paths AI scatters in your code.

crates/cargo-wdk/src/cli.rs
```diff
@@ -330,29 +426,129 @@ mod tests {
...
-        assert!(result.is_err());
-        assert_eq!(
-            result.err().unwrap().to_string(),
-            "`--verify-signature` cannot be used with `--sign-mode=off`."
-        );
+        fn parse_build_args(extra: &[&str]) -> Result<BuildArgs, clap::Error> {
+            let mut command_line = vec!["cargo-wdk", "wdk", "build"];
+            command_line.extend_from_slice(extra);
+            match Cli::try_parse_from(command_line)?.sub_cmd {
+                Subcmd::Build(build_args) => std::result::Result::Ok(build_args),
```
```suggestion
                Subcmd::Build(build_args) => Ok(build_args),
```

crates/cargo-wdk/README.md
```diff
@@ -108,6 +116,15 @@ If the `--verify-signature` flag is provided, the signatures are verified after
 
 `--verify-signature` cannot be combined with `--sign-mode=off` because if signing is off there is nothing to verify. Passing both will cause `build` to fail with an error.
 
+##### Customizing signtool arguments
+
+To sign with your own certificate or tweak any signing option, pass `--signtool-args` with a string of the arguments to forward to `signtool sign`.
+
+- When `--signtool-args` is **omitted**, cargo-wdk signs with the auto-generated WDR test certificate as described above.
+- When `--signtool-args` is **provided**, you own the full `signtool sign` option set (certificate selection, digest algorithm, etc.). cargo-wdk only wraps your arguments with the `sign` verb and the trailing file operand. If your arguments don't include a certificate selector (e.g. `/n`, `/s`, `/f`, `/sha1`), `signtool` auto-selects a code-signing certificate from your personal store and fails if none — or more than one — is found.
```
```suggestion
- When `--signtool-args` is **provided**, you own the full `signtool sign` option set (certificate selection, digest algorithm, etc.). `cargo-wdk` will prepend the `sign` verb to your arguments and append the trailing file operand so you should not provide them. If your arguments don't include a certificate selector (e.g. `/n`, `/s`, `/f`, `/sha1`), `signtool` auto-selects a code-signing certificate from your personal store and fails if none, or more than one, is found.
```

crates/cargo-wdk/README.md
```diff
@@ -133,3 +150,15 @@ If the `--verify-signature` flag is provided, the signatures are verified after
...
+
+    ```pwsh
+    cargo wdk build --signtool-args "/s MyStore /n MyCert /fd SHA256"
+    ```
+
+- To test-sign with a PFX file whose password is supplied inline, run:
+
+    ```pwsh
+    cargo wdk build --signtool-args "/f C:\certs\my.pfx /p <password> /fd SHA256"
+    ```
```
gurry: Don't add these examples. Like I said in the past, we don't have to add examples for every capability we have.

svasista-ms: Removed ✅

crates/cargo-wdk/README.md
svasista-ms: Applied the suggested changes ✅

crates/cargo-wdk/src/cli.rs
svasista-ms: Done ✅

crates/cargo-wdk/src/cli.rs
svasista-ms: anyhow's `Result` (with the explicit `String` error) can be used here. It need not be `std` `Result`.  Removed the fully qualified path ✅

crates/cargo-wdk/src/cli.rs
svasista-ms: Done ✅

crates/cargo-wdk/src/cli.rs
svasista-ms: Fixed ✅

crates/cargo-wdk/src/cli.rs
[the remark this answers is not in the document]
gurry: Right now we do not validate these args in order to keep the things simple. We could start doing that later. For now this is fine.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -503,20 +519,21 @@ impl<'a> PackageTask<'a> {
         Ok(())
     }
 
-    /// Signs the specified file using signtool command using certificate from
-    /// certificate store.
+    /// Signs the file with `signtool` by executing the following command:
+    /// `sign <signtool_args...> <file>`
```
krishnakumar4a4:
```suggestion
    /// `sign <signtool_args...> <file_path>`
```

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -525,22 +542,33 @@ impl<'a> PackageTask<'a> {
...
+        args.extend(signtool_args.iter().cloned());
+        // File operand (must be last).
+        args.push(file_path);
+
+        let arg_refs: Vec<&str> = args.iter().map(String::as_str).collect();
+        let redaction_indices: Vec<usize> = arg_refs
+            .iter()
+            .enumerate()
+            .filter_map(|(i, arg)| {
+                (arg.eq_ignore_ascii_case("/p") && i + 1 < arg_refs.len()).then_some(i + 1)
```
krishnakumar4a4: ~This check `i + 1 < arg_refs.len()` is unnecessary as we are already using an iterator `arg_refs`.~

At first glance, `i + 1 < arg_refs.len()` looked like an out of bounds check, but its not. Later understood that it is to validate if there is an argument following the `/p` switch. And also the `then_some(i+1)` looked like a conversion to `1-based` index until I went and looked into CommandExec.

I think you need to leave a comment here explaining why these additional checks and additions.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -276,34 +286,40 @@ impl<'a> PackageTask<'a> {
...
-            WDR_TEST_CERT_STORE,
-            WDR_LOCAL_TEST_CERT,
-        )?;
-
-        if verify_signature {
+        let sign_args = if signtool_args.is_empty() {
+            self.generate_certificate()?;
+            self.copy(&self.src_cert_file_path, &self.dest_cert_file_path)?;
+            // Default WDR test-cert switches.
+            vec![
```
krishnakumar4a4: Instead of calling to_string on each item. We could

```rust
["/v", "/s", WDR_TEST_CERT_STORE, "/n", WDR_LOCAL_TEST_CERT,
 "/t", DEFAULT_TIMESTAMP_URL, "/fd", "SHA256"]
    .map(ToString::to_string)
    .to_vec()
```

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -864,6 +897,218 @@ mod tests {
...
+                "C:/certs/my.pfx".to_string(),
+                "/p".to_string(),
+                "secret".to_string(),
+            ];
+            task.run_signtool_sign(Path::new("C:/pkg/driver.sys"), &signtool_args)
+                .expect("signing should succeed");
+        }
+
+        #[test]
+        fn sign_redacts_password_value() {
```
krishnakumar4a4:
```suggestion
        fn sign_redacts_password_value_arg_by_redaction_index() {
```

crates/cargo-wdk/src/actions/build/tests.rs
```diff
@@ -2057,7 +2060,7 @@ impl TestBuildAction {
         target_arch: CpuArchitecture,
     ) -> Self {
         let cwd = self.cwd.clone();
-        self.expect_final_package_dir_exists(driver_name, &cwd, true)
+        self.expect_final_package_dir_exists(driver_name, &cwd, false)
```
krishnakumar4a4: The name of the function and its responsibility is confusing now. Earlier `expect_final_package_dir_exists` with `true` indicates that the setup sets up only the `mock_fs_provider.expect_exists()` to return true.

Now with new changes, the function name and boolean value not communicating the intent as it is doing multiple things.
`.expect_final_package_dir_exists(...true)` -> expects dir, removal and creation
`.expect_final_package_dir_exists(...false)` -> expects no dir, no removal but there will be creation

----

Ideally, I would have liked to keep the assertions separate. But if we decide to keep it as single function to reduce verbosity, `.expect_final_package_dir_removal(driver_name, &cwd, true)` could be a better name which communicates what the function would do differently with the provided boolean value. Assuming the call to `exists` and `create_dir` will always be called.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -36,17 +36,25 @@ const MISSING_SAMPLE_FLAG_WDK_BUILD_NUMBER_RANGE: RangeInclusive<u32> = 25798..=
 const WDR_TEST_CERT_STORE: &str = "WDRTestCertStore";
 const WDR_LOCAL_TEST_CERT: &str = "WDRLocalTestCert";
 const STAMPINF_VERSION_ENV_VAR: &str = "STAMPINF_VERSION";
+const DEFAULT_TIMESTAMP_URL: &str = "http://timestamp.digicert.com";
 
 /// Signing mode.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum SignMode {
```
krishnakumar4a4, line 43: The `Profile` type is declared in `src/actions/mod.rs`, but it is only used by the `BuildTask` module.

In contrast, `SignMode` and `TargetPlatform` are declared directly within the `PackageTask` module, as they are only used there. This feels inconsistent.

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -356,61 +356,370 @@ mod kmdf_driver_with_target_override {
...
+            ensure_cert_in_store("WDRTestCertStore", "WDRLocalTestCert");
+            run_clean_cmd(&project_path);
+
+            let driver_name = driver.replace('-', "_");
+            let package_dir = format!("{project_path}/target/debug/{driver_name}_package");
+            let common_args = "/s WDRTestCertStore /n WDRLocalTestCert /fd SHA256";
+
+            // First build produces a signed driver binary we can reuse as an
+            // independent second file operand.
+            run_build_cmd(&project_path, Some(&["--signtool-args", common_args]), None);
```
krishnakumar4a4: Since the goal was to sign an additional file passed through --signtool-args, we can create a random file ourselves and let the build sign it. There is no need to find a DLL, copy it, strip its signature, and so on.

krishnakumar4a4 (changes requested):

crates/cargo-wdk/src/actions/build/package_task.rs
krishnakumar4a4: it seems this comment was outdated by a code change before I published this review. Resolving it.

crates/cargo-wdk/src/actions/build/tests.rs
gurry: I think `expect_final_package_dir_creation` is better as:

- creation is always done in both cases
- really creation is the main thing we want to verify (we care less about deletion etc.)

crates/cargo-wdk/tests/build_command_test.rs
gurry: This is what I felt too. That extra build seems wasteful. But `signtool` cannot sign arbitrary types of files. It is limited to things like DLLs, EXEs, CABs etc. None of them are easy to create by hand. We could carry one in the codebase but that also seems awkward. 

@svasista-ms can you please research and find out which file type amongst those supported by `signtool` can we easily create on the fly?

crates/cargo-wdk/src/actions/build/tests.rs
krishnakumar4a4: IMO, Removal is equally important, as it represents the new behaviour we introduced. 
If we go with `expect_final_package_dir_creation`, I would also like to see a separate test that explicitly verifies the deletion behaviour.

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -864,6 +900,237 @@ mod tests {
...
+            expected_redaction_indices: Vec<usize>,
+        ) -> CommandExec {
+            let mut command_exec = CommandExec::default();
+            command_exec
+                .expect_run_with_redaction()
+                .withf(move |command, args, redaction_indices, _env, _cwd| {
+                    command == "signtool"
+                        && args == expected
+                        && redaction_indices == expected_redaction_indices.as_slice()
+                })
```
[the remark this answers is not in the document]
svasista-ms: Not an issue. Resolving

crates/cargo-wdk/src/actions/build/tests.rs
gurry: I'm actually fine with the deletion check being just a part of the creation test. Deletion is only a secondary detail. The main goal is to get a package created which the user can deploy and use. Deletion is there just to avoid staleness in that process. So yeah, one test is perhaps okay. We can discuss over a call as well if you feel strongly about this.

krishnakumar4a4 (dismissed):

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -356,61 +356,344 @@ mod kmdf_driver_with_target_override {
...
+                None,
+                Some(&project_path),
+            );
+            let assertion = cmd.assert().failure();
+            let stderr = String::from_utf8_lossy(&assertion.get_output().stderr).to_string();
+            assert!(
+                stderr.contains("No certificates were found")
+                    || stderr.contains("signing driver binary"),
+                "expected a signtool certificate-selection failure, got: {stderr}"
+            );
```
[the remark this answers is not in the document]
gurry: Why is this test passing if the text we check is wrong?

crates/cargo-wdk/src/actions/build/mod.rs
```diff
@@ -31,10 +33,49 @@ use wdk_build::{
...
 };
 
-use crate::actions::Profile;
 #[double]
 use crate::providers::{exec::CommandExec, fs::Fs, metadata::Metadata, wdk_build::WdkBuild};
 
+/// `x86_64/Amd64` target triple name
+const X86_64_TARGET_TRIPLE_NAME: &str = "x86_64-pc-windows-msvc";
+/// `aarch64/Arm64` target triple name
+const AARCH64_TARGET_TRIPLE_NAME: &str = "aarch64-pc-windows-msvc";
```
gurry: The comments are redundant (most likely AI slop) because they just repeat what the const name says. We can remove them.

gurry (approved):

crates/cargo-wdk/src/actions/build/tests.rs
krishnakumar4a4: As discussed, I am good with the newer change `expect_final_package_dir_created(..., removes_existing: bool)` which also satisfies deletion behaviour.

crates/cargo-wdk/tests/build_command_test.rs
krishnakumar4a4: Since SignTool can only sign specific types of files, I am fine with leveraging an existing DLL file instead of creating a new one, as creating a new DLL is not trivial.

krishnakumar4a4 (approved):

PR #721, opened 2026-08-18
feat(wdk): add error-level debug output

### LINKED ISSUE #616: Implement an `eprintln!` macro and replace the `println!` implementation in `dbg!`
Replace the `println!` solution in the `dbg!` macro with a no_std implementation of `eprintln!` to target different debug message levels.

Early thoughts about what a 'no_std' implementation of `eprintln!` could be:

- Define and utilize the `[in] Level` parameter in the `DbgPrintEx` function (part of wdm.h) to differentiate `println!` and `eprintln!`, mimicking stderr.
- [DbgPrintEx function](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-dbgprintex)
- Potential levels:
    - `println!` -> DPFLTR_INFO_LEVEL
    - `eprintln!` -> DPFLTR_ERROR_LEVEL
    - [Reading and Filtering Debugging Messages](https://learn.microsoft.com/en-us/windows-hardware/drivers/debugger/reading-and-filtering-debugging-messages)

ITSMERNB:
## Summary
- add a no_std `eprintln!` macro for WDK debug output
- route WDM/KMDF `print!` and `println!` through `DbgPrintEx` at `DPFLTR_INFO_LEVEL`
- route `eprintln!` and `dbg!` through `DbgPrintEx` at `DPFLTR_ERROR_LEVEL`
- preserve the existing `DbgPrint` component/filter behavior with `DPFLTR_DEFAULT_ID`
- keep UMDF output on `OutputDebugStringA`, which has no message-level parameter

Fixes #616

## Testing
- `cargo test -p wdk --locked` (29 unit tests + doctests pass)
- `cargo clippy -p wdk --locked --all-targets -- -D warnings`
- `git diff --check`
- KMDF/UMDF fixture checks reach `wdk-sys` but the local WDK install is missing `Windows Kits\10\Include\10.0.26100.0\km\crt`, so bindgen stops before compiling the driver-model-specific crate path.

### DIFF
--- a/crates/wdk/src/lib.rs
+++ b/crates/wdk/src/lib.rs
@@ -15,7 +15,7 @@
     driver_model__driver_type = "KMDF",
     driver_model__driver_type = "UMDF",
 ))]
-pub use print::_print;
+pub use print::{_eprint, _print};
 #[cfg(any(
     driver_model__driver_type = "WDM",
     driver_model__driver_type = "KMDF",
--- a/crates/wdk/src/print.rs
+++ b/crates/wdk/src/print.rs
@@ -12,12 +12,12 @@ use std::ffi::CString;
 #[cfg_attr(
     any(driver_model__driver_type = "WDM", driver_model__driver_type = "KMDF"),
     doc = r"
-The output is routed to the debugger via [`wdk_sys::ntddk::DbgPrint`], so the `IRQL` 
+The output is routed to the debugger via [`wdk_sys::ntddk::DbgPrintEx`], so the `IRQL`
 requirements of that function apply. In particular, this should only be called at 
 `IRQL` <= `DIRQL`, and calling it at `IRQL` > `DIRQL` can cause deadlocks due to
 the debugger's use of IPIs (Inter-Process Interrupts).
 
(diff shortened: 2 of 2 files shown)


ITSMERNB:
Addressed the two review notes in 3ccc4eb: DbgPrintEx now keeps DPFLTR_DEFAULT_ID so existing print!/println! visibility is preserved while only the message level changes, and the unused enum discriminants are removed. `cargo test -p wdk --locked` passes 29 unit tests plus doctests, and clippy is clean with `-D warnings`.

ITSMERNB:

ITSMERNB:
Quick follow-up on this one. GitHub has now marked the external-contributor workflow runs from August as failed after they sat awaiting approval; those runs contain zero executed jobs, so this is not a test/build failure on the branch. The current head is still `3ccc4eb`, mergeable, and the local `wdk` tests/clippy validation from the last update was clean. If you would like any rebase or code changes before review, I am happy to do them; otherwise this looks to be waiting on the maintainer-side workflow/review step.

PR #734, opened 2026-09-02
feat(cargo-wdk): add `--infverif-args` passthrough to customize `inverif` options

### LINKED ISSUE #730: Support `infverif` passthrough in `cargo-wdk`
`cargo-wdk` hardcodes the `infverif` invocation, so there is no way to customize its arguments.

Current invocation:
```
infverif /v <mode> [/msft|/samples] <pkg>\<name>.inf
```

Add `--infverif-args "<raw args>"` to `cargo wdk build` so that custom arguments can be passed through to `infverif`.

svasista-ms:
`cargo-wdk` currently hardcodes its `infverif` invocation during packaging. Only the validation mode is reachable (through `--target-platform`), so options like `/rulever` (pin or preview the InfVerif rule version), `/info`, `/stampinf`, `/osver`, `/provider` cannot be used at all.

This PR adds `--infverif-args` to the `cargo wdk build` command similar to other tools.

The mode flags `| /h, /w, /u |` are derived by `cargo-wdk` from `--target-platform`, and the `.inf` file operand is the generated package INF. So, they are rejected at parse time.

Resolves #730 

### Screenshots

- Help Output:

after-01-help

- With custom `infverif-args`:

after-03-appended

- Reserved switches are rejected:

after-04-reserved


crates/cargo-wdk/src/cli.rs
```diff
@@ -228,6 +238,34 @@ impl BuildArgs {
...
+    fn infverif_args(&self) -> Result<Option<Vec<String>>, clap::Error> {
+        const MODE_FLAGS: [&str; 4] = ["h", "w", "u", "k"];
+
+        let Some(args) = self.infverif_args.clone().map(|parsed| parsed.0) else {
+            return Ok(None);
+        };
+        for arg in &args {
+            let lower = arg.to_ascii_lowercase();
+            let reason = if MODE_FLAGS.contains(&lower.trim_start_matches(['/', '-'])) {
+                "cargo-wdk derives the mode flag from `--target-platform`"
```
```suggestion
                format!("cargo-wdk derives the mode flag `/{lower}` from `--target-platform`")
```

Mention the mode flag name in the reason

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -1374,6 +1390,85 @@ mod tests {
...
+                    stderr: vec![],
+                })
+            });
+
+        let task = PackageTask::new(params, &wdk_build, &command_exec, &fs);
+        assert!(task.run_infverif().is_ok());
+    }
+
+    #[test]
+    fn run_infverif_with_custom_args_appends_them_before_the_inf_path() {
```
```suggestion
    fn run_infverif_with_custom_args_forwards_them_verbatim() {
```

We should maintain naming consistency with similar tests for other child tools

crates/cargo-wdk/src/actions/build/package_task.rs
```diff
@@ -1374,6 +1390,85 @@ mod tests {
...
+    }
+
+    #[test]
+    fn run_infverif_with_custom_args_appends_them_after_the_sample_flag() {
+        assert_infverif_args(
+            true,
+            Some(vec!["/stampinf".to_string()]),
+            vec!["/v", "/u", "/samples", "/stampinf"],
+        );
+    }
```
gurry: I don't think we need a separate test for `/samples`. Just merge it with the above test if possible

crates/cargo-wdk/src/cli.rs
```diff
@@ -249,13 +249,14 @@ impl BuildArgs {
         };
         for arg in &args {
             let lower = arg.to_ascii_lowercase();
-            let reason = if MODE_FLAGS.contains(&lower.trim_start_matches(['/', '-'])) {
-                "cargo-wdk derives the mode flag from `--target-platform`"
+            let mode_flag = lower.trim_start_matches(['/', '-']);
+            let reason = if MODE_FLAGS.contains(&mode_flag) {
+                format!("cargo-wdk derives the mode flag `/{mode_flag}` from `--target-platform`")
```
gurry: Nit: this needs to be the actual flag that the user typed. If the user typed `/h` then reason should contain `/h` and if they typed `-h` then it should be `-h`. Otherwise it will be confusing to the user.

The string I shared was just an example, but I'm sorry if I misled you.

svasista-ms: Thank you for calling this out! I have fixed it by printing the user supplied arg as is ([latest commit](https://github.com/microsoft/windows-drivers-rs/pull/734/changes/976b78c2e21918c3be7e71753d5e759b69a8a3f3#diff-4861850c7366ce9ab803db669bc522b1265ae68a766778ab50c3a9976fde66d9R252-R256))

crates/cargo-wdk/src/cli.rs
```diff
@@ -228,6 +238,37 @@ impl BuildArgs {
         }
         Ok(Some(args))
     }
+
+    /// Resolves the arguments to forward to `infverif`. Rejects the arguments
+    /// cargo-wdk supplies itself: the mode flags derived from the
+    /// `--target-platform` option and the INF file operand.
+    fn infverif_args(&self) -> Result<Option<Vec<String>>, clap::Error> {
+        const MODE_FLAGS: [&str; 3] = ["h", "w", "u"];
```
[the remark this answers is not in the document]
svasista-ms: Updated PR description. `/k` is not controlled by `cargo-wdk`, so it should not be in the rejection list.

crates/cargo-wdk/src/cli.rs
```diff
@@ -556,6 +599,61 @@ mod tests {
...
+                .expect_err("reserved argument should be rejected");
+            assert!(
+                err.to_string().contains(expected_reason),
+                "unexpected error for {value:?}: {err}"
+            );
+        }
+
+        #[test]
+        fn infverif_args_rejects_mode_flags() {
+            for (value, mode_flag) in [("/h", "/h"), ("/w", "/w"), ("/U", "/U"), ("/info -w", "-w")]
```
gurry: Also add a more complex case containing mutltiple reject-able args along with some acceptable args.

svasista-ms: Added a case with mixed input 👍

crates/cargo-wdk/README.md
```diff
@@ -84,6 +84,9 @@ Driver Signing:
 Inf2Cat Options:
       --inf2cat-args <ARGS>        Custom arguments to pass to `inf2cat` when generating the catalog file, e.g. `--inf2cat-args '/os:10_x64,10_GE_X64 /uselocaltime'`
 
+InfVerif Options:
+    --infverif-args <ARGS>       Custom arguments to pass to `infverif` when validating the INF, e.g. `--infverif-args '/rulever 10.0.22621 /info'`
```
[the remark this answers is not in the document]
svasista-ms: Fixed 👍

crates/cargo-wdk/README.md
gurry, line 116: Add a section like this for infverif.

Eventually we want to collect all such pass through sections under a single parent section, but we can do that later. For now just adding a top level section is enough

crates/cargo-wdk/src/cli.rs
```diff
@@ -164,6 +164,16 @@ pub struct BuildArgs {
     )]
     pub inf2cat_args: Option<PassthroughArgs>,
 
+    /// Custom arguments to pass to `infverif` when validating the INF,
+    /// e.g. `--infverif-args '/rulever 10.0.22621 /info'`.
+    #[arg(
+        long,
+        value_name = "ARGS",
+        value_parser = parse_passthrough_args,
+        help_heading = "InfVerif Options"
```
gurry, line 175: Don't we need `allow_hyphen_values = true` here similar to stampinf?

Also do we need it in the case of other tools?

crates/cargo-wdk/src/cli.rs
```diff
@@ -556,6 +599,66 @@ mod tests {
...
+            for value in ["extra.inf", "/info C:\\pkg\\other.INF"] {
+                assert_infverif_args_rejected(
+                    value,
+                    "cargo-wdk supplies the INF file operand itself",
+                );
+            }
+        }
+
+        #[test]
+        fn infverif_args_allows_other_switches() {
```
```suggestion
        fn infverif_args_allows_other_args() {
```

crates/cargo-wdk/src/cli.rs
```diff
@@ -556,6 +599,66 @@ mod tests {
...
+                Some(vec![
+                    "/rulever".to_string(),
+                    "10.0.22621".to_string(),
+                    "/info".to_string()
+                ])
+            );
+        }
+
+        #[test]
+        fn infverif_args_defaults_to_none() {
```
```suggestion
        fn infverif_args_allows_empty_list_of_args() {
```

crates/cargo-wdk/tests/build_command_test.rs
```diff
@@ -745,6 +745,28 @@ fn kmdf_driver_with_custom_inf2cat_args_builds_successfully() {
...
+        None,
+        None,
+        None,
+        Some(&["--infverif-args", "/rulever 10.0.22621 /stampinf", "-v"]),
+    );
+    assert!(
+        stderr.contains(
+            "Running: infverif [\"/v\", \"/u\", \"/rulever\", \"10.0.22621\", \"/stampinf\", "
+        ),
+        "expected `--infverif-args` to be appended after the mode flag and before the INF path; \
```
gurry:
```suggestion
        "expected `--infverif-args` to work \
```

All you we want here is for the infverif-args to work right?

crates/cargo-wdk/README.md
svasista-ms: Done 👍

crates/cargo-wdk/src/cli.rs
svasista-ms: Thank you for catching this. `infverif` accepts both `-` and `/`, I've fixed it for `infverif` in this PR. 

I checked the other tools:
1. `stampinf` accepts both | allowed in `cargo-wdk` ✅ 
2. `infverif` accepts both | allowed in `cargo-wdk` ✅ 
3. `signtool` also accepts both | **- not allowed** in `cargo-wdk`. Has to be fixed. ❌ 
4. `inf2cat` accepts only `/` | correct in `cargo-wdk` ✅

crates/cargo-wdk/src/cli.rs
svasista-ms: Fixed ✅

crates/cargo-wdk/src/cli.rs
svasista-ms: This test was redundant. Passing empty is allowed and results in default behavior. We don't have such a test for other tool args, so I have removed it.

crates/cargo-wdk/tests/build_command_test.rs
svasista-ms: Yes, changed to a simpler message

crates/cargo-wdk/src/cli.rs
```diff
@@ -164,6 +164,18 @@ pub struct BuildArgs {
...
     pub inf2cat_args: Option<PassthroughArgs>,
 
+    /// Custom arguments to pass to `infverif` when validating the INF,
+    /// e.g. `--infverif-args '/rulever 10.0.22621 /info'`.
+    #[arg(
+        long,
+        value_name = "ARGS",
+        // `infverif` args can be `-` prefixed.
+        allow_hyphen_values = true,
+        value_parser = parse_passthrough_args,
```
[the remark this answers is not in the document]
svasista-ms: `Infverif` accepts both `-` and `/`

crates/cargo-wdk/src/cli.rs
```diff
@@ -556,6 +601,63 @@ mod tests {
...
+            );
+        }
+
+        #[test]
+        fn infverif_args_rejects_mode_flags() {
+            for (value, mode_flag) in [
+                ("/h", "/h"),
+                ("/w", "/w"),
+                ("/U", "/U"),
+                ("-w", "-w"),
```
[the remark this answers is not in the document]
svasista-ms: 1. Infverif accepts both `-` and `/`. So if `-v` is passed as the first argument to `infverif-args`, `allow_hyphen_values` must be enabled. Otherwise, it will be rejected.

2. Enabling hyphen values is required for the first argument only. 

3. Swallowing of args would happen if the cli is deliberately wrong, e.g `--infverif-args --locked`, in this case `--locked` will be swallowed by `infverif-args` and not passed to `build` command.

No change is required.

crates/cargo-wdk/src/cli.rs
gurry: In the tests you have written for stampinf and infverif ensure that the input args use a mixture of `/` and `-`. We can then repeat this for other two tools in separate PRs later.

Once you update this for both stampinfo and infverif I'll resolve this comment.

crates/cargo-wdk/src/cli.rs
```diff
@@ -617,7 +619,7 @@ mod tests {
                 ("/h", "/h"),
                 ("/w", "/w"),
                 ("/U", "/U"),
-                ("/info -w", "-w"),
+                ("-w", "-w"),
```
gurry: why did you remove `/info` here?

crates/cargo-wdk/src/cli.rs
gurry, line 620: why did you remove `/info` from here?

crates/cargo-wdk/README.md
```diff
@@ -112,6 +115,15 @@ If you have a workspace with a mix of sample and non-sample driver projects, the
...
 To target a specific set of Windows versions or to customize the behaviour of `inf2cat` in any other way, pass `--inf2cat-args` with a string of the arguments to forward to `inf2cat`. `cargo-wdk` itself provides the `/driver` argument so do not include it or its alias `/drv`.
 
+#### Customizing `infverif` arguments
+
+To pin the InfVerif rule version, print rule details, or customize the behaviour of `infverif`, pass `--infverif-args` with a string of the arguments to forward to `infverif`.
+
+**Note:**
+
+- `cargo-wdk` derives the validation mode flag (`/h`, `/w` or `/u`) from `--target-platform`, so passing any of them is an error.
+- `cargo-wdk` passes the generated INF file itself, so passing a `.inf` operand is an error.
```
```suggestion
To customize the behaviour of `infverif`, pass `--infverif-args` with arguments to forward to `infverif`. Args `-h`, `-w`, `-u` and paths to `.INF` files are not allowed because they are always supplied by `cargo-wdk` itself. 

```

crates/cargo-wdk/src/cli.rs
```diff
@@ -228,6 +240,37 @@ impl BuildArgs {
...
+            let mode_flag = arg.trim_start_matches(['/', '-']).to_ascii_lowercase();
+            let reason = if MODE_FLAGS.contains(&mode_flag.as_str()) {
+                format!(
+                    "cargo-wdk derives the mode flag `{arg}` from the `--target-platform` option"
+                )
+            } else if Path::new(arg)
+                .extension()
+                .is_some_and(|extension| extension.eq_ignore_ascii_case("inf"))
+            {
+                "cargo-wdk supplies the INF file operand itself".to_string()
```
```suggestion
                "cargo-wdk supplies the INF file path itself".to_string()
```

crates/cargo-wdk/src/cli.rs
```diff
@@ -228,6 +240,37 @@ impl BuildArgs {
         }
         Ok(Some(args))
     }
+
+    /// Resolves the arguments to forward to `infverif`. Rejects the arguments
+    /// cargo-wdk supplies itself: the mode flags derived from the
+    /// `--target-platform` option and the INF file operand.
```
```suggestion
    /// `--target-platform` option and the INF file path.
```

svasista-ms: Updated ✅

crates/cargo-wdk/src/cli.rs
svasista-ms: Updated ✅

crates/cargo-wdk/README.md
svasista-ms: Suggestion applied ✅

crates/cargo-wdk/src/cli.rs
svasista-ms: Added it back. I think it got deleted when I edited it in the previous commit.

crates/cargo-wdk/src/cli.rs
svasista-ms: Added it back ✅

gurry (approved):

krishnakumar4a4 (approved):

